PR #57, opened 2014-12-15
Turn off the quoting class name with slashes, when use sqlite

NemoD503:
We can not to quote class name with slashes, when using sqlite, because we will see an error like

     Class 'Autor\\Plugin\\Models\\Model' not found

But, when we use mysql, we need use str_replace, or we will see another error

     Class 'AutorPluginModelsModel' not found

So, I think we can check the config of connection to prevent this error.


daftspunk:
Thanks!

PR #58, opened 2014-12-22
Fix error with numeric concatination in PostgreSQL

shtricul:
If column has number type, it needs cast to string


daftspunk:
Thanks!

PR #59, opened 2014-12-23
File->getThumb just halt without any error returned

fmulyono:
If the physical file doesn't exist in the disk, but the file record is still in the database, getThumb produced no error at all, but the whole PHP execution just stop.


daftspunk:
Thanks!

PR #61, opened 2015-01-09
Turning October\Rain\Support\Markdown into a Facade

mossadal:
Turning October\Rain\Support\Markdown into a Facade
for a new MarkdownServiceProvider, thus allowing users
to swap in another Markdown parser.

The existing Markdown class has been renamed to Parsedown,
and is used as the default parser.

Note: The built-in Parsedown interpreter often corrupts
MathJax input,
see https://octobercms.com/forum/post/extending-a-rainlabblog-overriding-beforesave

To work, this change also requires a change in the main
OctoberCMS repository.

In modules/system/providers.php, we need to load the
MarkdownServiceProvider: add

    'October\Rain\Support\MarkdownServiceProvider',

to line 42. I'll make another pull request for this.


src/Support/MarkdownData.php
```diff
@@ -0,0 +1,18 @@
...
+ * @package october\support
+ * @author Alexey Bobkov, Samuel Georges
+ */
+class MarkdownData
+{
+    public function __construct($text) {
+        $this->text = $text;
+    }
+
+    var $text;
```
daftspunk: I think "var" should be "public"?

mossadal:
Yes, you're right. Bad old habits die hard. I also added some documentation concerning the events.

daftspunk:
Good stuff, thanks

PR #63, opened 2015-01-23
Let plugins add hashable fields

Flynsarmy:
I need to add a hashable attribute to the backend users model but the hashable array is protected. Figured I'd add a method to allow plugins to add hashable fields.


daftspunk:
Looks good, we don't ever want to allow removing these attributes though! =)

PR #65, opened 2015-01-26
Add settable attributeNames array for use in validation trait

Flynsarmy:
This PR adds support for an `attributeNames` model property which the *Validation* trait uses to display better attribute names in error messages.

## Usage

In your model:
```php
public $attributeNames = ['my_attrib' => 'My Attrib Name'];
```

Based on [this Ardent PR](https://github.com/laravelbook/ardent/pull/213/files). Keeps support for localization that was already present. Order of priority is explicitly passed, model property, localization file.

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -15,6 +15,12 @@ trait Validation
      */
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
Beautiful, thanks

PR #68, opened 2015-02-03
Add addDateAttribute method

Flynsarmy:
Models currently have a `dates` property that converts date attributes to carbon objects but [it's protected](https://github.com/octobercms/library/blob/master/src/Database/Model.php#L54) so plugins can't add carbon date fields. This PR adds a simple adder method to make this possible.


daftspunk:
LGTM

PR #71, opened 2015-02-10
Fixes fatal error exception when saving BrokenImage

carlos-arroyo:
Fixes fatal error exception when saving BrokenImage file because the folder structure doesn’t exists. 

screen shot 2015-02-10 at 1 40 48 pm


daftspunk:
Thanks

PR #72, opened 2015-02-12
Add forcingRelative method

Flynsarmy:
`URL::full()` is currently erroring out because you're using the wrong property name [here](https://github.com/Flynsarmy/library/blob/master/src/Html/UrlGenerator.php#L36) so I fixed that and added a getter method which may prove helpful in the future.


Flynsarmy:
@daftspunk requested I close this PR. It would apparently not be helpful to be able to determine if the app is currently using relative URLs. Instead I've sent #73 to fix the issue.

PR #80, opened 2015-02-28
fixes issue with objects as model fields iteration

keiosweb:
Objects can be model field values when using transformative traits that create objects on fetch and dismantle them to data on save, ie.: Carbon, [Financial](https://github.com/keiosweb/oc-trait-financial) or [Serializable](https://github.com/keiosweb/oc-trait-serializable). Router helper is crashing if it encounters objects that cannot be forced into a string. If this is by design, there should be information about requirement for all model fields to be convertable to strings, but as arrays are simply omitted and October doesn't use automatic Carbon instantiation on timestamp fields, there is no need to force objects into strings here.

### DIFF
--- a/src/Router/Helper.php
+++ b/src/Router/Helper.php
@@ -77,7 +77,7 @@ class Helper
     {
         $defaultColumns = ['id'];
         foreach ($columns as $column) {
-            if (!isset($object->{$column}) || is_array($object->{$column}))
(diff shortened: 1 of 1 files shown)


daftspunk:
Two problems:

1. This is on the wrong branch, should go to the `develop` branch please =)

2. Some objects can be cast to strings so we need to check for it:

```
|| (is_object($object->{$column}) && !method_exists($object->{$column}, '__toString'))
```

Issue #78, opened 2015-02-24
::extend method is broken


keiosweb:
October\Rain\Database\Traits\Validation uses global scope registration via static trait boot method, this method is called **once on model registration** and ~~every single time object using this trait is constructed~~ every single time Illuminate\Database\Eloquent\Model::flushEventListeners is called and if this method uses static::extend call, which has no check if callback was already registered in static $extendableCallbacks array, it just keeps adding callbacks for ~~every constructed model~~ every flushEventListeners call and those callbacks are then executed, every single one of them. This applies to every other trait that uses static::extend() call in its boot method and ~~most~~ many of them do.

Let's imagine a collection of 50 models using validation trait **and 20 test cases, every one of them rebooting event handlers**, that gives us ~~50~~ 20 extend callbacks, each of them binding an event to model.saveInternal, those event handlers will be executed too, but on every single model, that gives us 50 models * 20 event handlers, ~~250~~ 1000 function calls and let's hope that no one will actually try to transform model fields with trait built on model instance like I tried, because this causes bugs that are is impossible to untangle without xdebug and profiler.

I'd suggest registering callbacks in a hashmap under key and dropping every other extend call to same key if it's already taken, removing things from global state would be nice too, it would actually allow people to unbork this.

So, about that global state, yeah, people hate it. For a reason.

Correction: I created a test plugin to help to understand where's the problem: OctoberPluginTestCase calls flushEventListeners for every model of tested plugin (because global state has to be killed between tests). Here's repo: https://github.com/keiosweb/oc-plugin-extendisborked, git clone it to /plugins/keios/, rename plugin directory to extendisborked, get into it, composer update and phpunit.

keiosweb:
This issue (as it is related to default Laravel behavior and model traits were meant to only provide global scopes) can be circumvented by adding following code to every model trait, that uses ::extend call:

``` php

protected static $traitNameAlreadyBooted = false;

public static function bootTraitName()
{
    if (static::$traitNameAlreadyBooted) {
        return;
    }

    static::$traitNameAlreadyBooted = true;

    static::extend(function($model) { ...

```

keiosweb:
[...]

PR #82, opened 2015-02-28
fix for multiple model boot and events binding during tests, related to ...

keiosweb:
...https://github.com/octobercms/library/issues/78 and https://github.com/laravel/framework/issues/1181

Moved event booting to model constructor, event boots are registered in static array on class just like model boots are registered on Illuminate\Database\Eloquent\Model, hence no multiple event bindings. Boot method is executed only once for every Model class always, event bindings are correctly bound from non-static context (so can be rebound on every application run) and flushed properly by flushEventListeners method. Related changes in test suite: https://github.com/keiosweb/oc-plugin-testsuite/commit/ce830745c2f98bd9f7424a204bf1b2e2876f30aa, tests are passing correctly, no issues with local october installation found. October\Rain\Database\Model::boot() removed, as is now obsolete - model bootNicerEvents call is done by constructor.

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -149,12 +149,15 @@ class Model extends EloquentModel
      */
     protected static $relationTypes = ['hasOne', 'hasMany', 'belongsTo', 'belongsToMany', 'morphTo', 'morphOne', 'morphMany', 'morphToMany', 'morphedByMany', 'attachOne', 'attachMany', 'hasManyThrough'];
 
+    protected static $eventsBooted = [];
+
     /**
      * Constructor
      */
     public function __construct(array $attributes = [])
     {
         parent::__construct();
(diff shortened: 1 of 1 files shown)


daftspunk:
OK - fingers crossed on this one, although I don't see any major differences to the logic. Good job!

src/Database/Model.php
```diff
@@ -149,12 +149,15 @@ class Model extends EloquentModel
...
 
+    protected static $eventsBooted = [];
+
     /**
      * Constructor
      */
     public function __construct(array $attributes = [])
     {
         parent::__construct();
+        static::bootNicerEvents();
```
daftspunk: No real reason for this to be static anymore! Yay! I have revised this to be a local method.

PR #84, opened 2015-03-02
Updated Mailer so that it returns the response

ariponce:
In Laravel 5, the Mailer's send method returns the response, which can be useful when using services like Mandrill.

https://github.com/laravel/framework/blob/5.0/src/Illuminate/Mail/Mailer.php#L166


daftspunk:
Thanks!

PR #86, opened 2015-03-04
Feature: adds translation string parsing for Validation trait

keiosweb:
handles translation strings for custom messages and attribute names


daftspunk:
LGTM

PR #91, opened 2015-03-26
Bugfix when Model's relation is not a Collection

shina:
I'm not sure if this is the best solution. However, it fixes the problem of pushing a Model with hasOne relation.


shina:
Now I'm afraid it is not a bug, when I installed the same project on another machine it worked.

PR #94, opened 2015-03-27
fix validation trait static event bindings 

keiosweb:
but dont remove native validation events

updated to agreed version.

references issue #78


scottbedard:
Thanks for your work on this Lucas!

PR #96, opened 2015-04-02
Set file visibility in cloud storage

ariponce:
When uploading a file to a cloud storage, set the visibility automatically based on the model config.


daftspunk:
Thanks!

PR #101, opened 2015-04-13
Wrap single model in array in push(), fixes #1067 in octobercms/october

pfrend:
Wrap single model in array instead of casting which creates array of all model fields, meaning push will fail.


daftspunk:
We need to check if it is a model before doing this, in case it is an array of models. I will amend.

PR #102, opened 2015-04-14
Fixed exceptions not being out-put in case of wrong PHP syntax

MatissJanis:
For example, place this faulty line anywhere in Your code:
    
    echo abc;

Expected output: exception page

Current output: blank white page with no information


daftspunk:
Thanks

PR #104, opened 2015-04-15
add October specific implementation of ->saveMany(), references #103

pfrend:
As per issue #103 , adds saveMany implementation to align with other October specific persistance methods.

Not sure if the deferred binding will work if a $sessionKey is passed, as the $models you pass in saveMany tend to not have been saved yet, meaning the call to bindDeferred wouldn't have an ID to write to the table, but I guess that's a separate issue. 

One possible solution to this could be in the save method of the HasOneOrMany trait, to save the $model before calling $this->add() maybe?

### DIFF
--- a/src/Database/Relations/HasOneOrMany.php
+++ b/src/Database/Relations/HasOneOrMany.php
@@ -22,6 +22,21 @@ trait HasOneOrMany
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
LGTM!

PR #107, opened 2015-04-23
fix bug in deferred bindings

tiancheng91:
fix bug in deferred bindings when use multi connections with diff table prefix


tiancheng91:
please close this, It's still in error

PR #113, opened 2015-05-29
Translate templates

alvaro-canepa:


daftspunk:
These stubs should be kept simple and clean, since they are a resource for learning. Introducing language keys will steepen the learning curve too much.

Thanks

PR #119, opened 2015-06-28
fix for validation of array values

shina:
This PR try to fix the validation of array values.

When defining a $rule for a model's jsonable attribute value, it was not unserializing before the validation occur.
For this I had to make Model::jsonable read public with getJsonable() method.


shina:
This is only for Jsonable attributes.
Perhaps another field types have this problem too.

daftspunk:
Thanks!

PR #120, opened 2015-07-15
Update getStorageDirectory method to allow custom folder names

ariponce:
Right now, the getStorageDirectory method returns a hard-coded path, but we can change the uploads folder on the CMS config file, causing this method to return the wrong path.


keiosweb:
Added comment in PR.

daftspunk:
@keiosweb is right

PR #121, opened 2015-07-16
fix mailgun and mandrill

davisvagris:
Don't need a october.driver or something else. This fix on new guzzle version. Small situation. i have laravel socialite they need new version on guzzle. but laravel 5.0 need old version.

### DIFF
--- a/src/Mail/MailServiceProvider.php
+++ b/src/Mail/MailServiceProvider.php
@@ -49,4 +49,14 @@ class MailServiceProvider extends MailServiceProviderBase
         });
 
     }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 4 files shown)


daftspunk:
Can you provide details, what happened? Has something changed?

davisvagris:
Fixed

daftspunk:
So this feature removes the requirement on Guzzle?

daftspunk:
You will find plugins can no longer override the core packages anymore. Instead you should fork Socialite to support October.Driver's guzzle version.

keiosweb:
>You will find plugins can no longer override the core packages anymore.

:+1: less debugging ftw

PR #122, opened 2015-07-25
Fix in_array argument order

Flynsarmy:
You have the `in_array` arguments in `addDateAttribute` method around the wrong way resulting in the error *in_array() expects parameter 2 to be array, string given* when called.


daftspunk:
Good save, good save

PR #123, opened 2015-07-26
Don't allow duplicate hashable values

Flynsarmy:
Similar to #122 we shouldn't allow adding of the same hashable value more than once.


daftspunk:
LGTM

PR #126, opened 2015-08-13
Fix cancelDeferredActions query

gabsource:
The where clauses where firing SQL Exceptions because of bad syntax.


daftspunk:
Good pick up, thanks @gabsource

PR #129, opened 2015-09-15
Change Exception to AuthException

alvaro-canepa:
The translation doesn't work because the use of `Exception` instead of `AuthException` in `Auth::authenticate`


daftspunk:
Thanks!

PR #130, opened 2015-09-26
Remove unwanted / suffix from single wildcard segments

Flynsarmy:
## Issue
* I have a CMS Page with slug `blog/:slug*` 
* I hit the URL `/blog/foo` in my browser
* I call `$this->param('slug')` in my component `onRun` method and get *foo/*

## Expected Result
I should be getting *foo*. No other param wildcard or otherwise has a */* suffix.

## Additional Info
This issue only occurs when you hit a wildcard segment with a single URL segment. For example if I hit the URL `/blog/foo/bar` then `$this->param('slug')` will correctly return *foo/bar*.

My PR includes new tests to cover this.

### DIFF
--- a/src/Router/Rule.php
+++ b/src/Router/Rule.php
@@ -195,7 +195,7 @@ class Rule
                 /*
                  * Determine if wildcard and add stored paramters as a suffix
                  */
-                if (Helper::segmentIsWildcard($patternSegment)) {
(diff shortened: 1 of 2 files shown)


daftspunk:
Good pick up! Thanks for the tests too

PR #131, opened 2015-09-30
Removes default on $args in __call

michaelquery:
# Issue

The __call method's second parameter, here called `$params`, is required and will always be present as an array (http://www.php.net/manual/en/language.oop5.overloading.php#object.call), adding a default value (like `$params = null`) is not needed and even causes problems, as described in padraic/mockery#263.

# Fix
Removing this default allows us to manipulate the object as expected (especially in Mockery tests).


daftspunk:
LGTM! Thanks

PR #133, opened 2015-10-08
Use AuthException instead of Exception

vojtasvoboda:
Use AuthException instead of Exception. Exception is parent of AuthException, so every Exception catches in other plugins would also work with new AuthException.

AuthException is more semantic and is used (for example) in RainLab.User Account component, ResetPassword component etc.


daftspunk:
Please use `develop` branch 

Merged in ce0a6f3639d81d19ede8d1a3350170ca24b09615

PR #135, opened 2015-10-09
Clean up

eezhal92:
Sorry, I attempted pull request to master branch, but it close it immediately.

Improve docblock and use `AuthException` instead of `Exception` at `October\Rain\Auth\Models\Preferences` for consistency cs.


daftspunk:
Thanks

PR #139, opened 2015-11-04
Modified validator to obey connection

tschallacka:
When a model uses a different database connection via $connection the validator
would not obey this, making the validator useless for those models.
By adding an optional connection string the validator will use the database connection
provided without interfering with any other code that might directly call the validator
instantiation code.
I made an issue https://github.com/octobercms/october/issues/1580 there, this is my proposed fix basically.

    class RichIndexPage extends Model
      {

      /**
       */
       public $table = 'sometable';
       public $connection = 'database_two';

       public $rules = [ 'name' => 'required|unique:sometable,name'];
       /**
        * @var array Guarded fields
        */
       protected $guarded = ['*'];
       /**
        * @var array Fillable fields
        */
       protected $fillable = [;
       /**
        * @var array Relations
        */
       public $hasOne = [];
       public $hasMany = [];
       public $belongsTo = [];
       public $belongsToMany = [];
       public $morphTo = [];
       public $morphOne = [];
       public $morphMany = [];
       public $attachOne = [];
       public $attachMany = [];
      }

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -2,6 +2,7 @@
 
 use Lang;
 use Input;
+use App;
 use October\Rain\Database\ModelException;
 use Illuminate\Support\MessageBag;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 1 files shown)


tschallacka:
crap, wrong branch, sorry.

PR #140, opened 2015-11-04
Modified validator to obey connection

tschallacka:
When a model uses a different database connection via $connection the validator
would not obey this, making the validator useless for those models.
By adding an optional connection string the validator will use the database connection
provided without interfering with any other code that might directly call the validator
instantiation code.
I made an issue https://github.com/octobercms/october/issues/1580 there, this is my proposed fix basically.

```
class RichIndexPage extends Model
  {

  /**
   */
   public $table = 'sometable';
   public $connection = 'database_two';

   public $rules = [ 'name' => 'required|unique:sometable,name'];
   /**
    * @var array Guarded fields
    */
   protected $guarded = ['*'];
   /**
    * @var array Fillable fields
    */
   protected $fillable = [;
   /**
    * @var array Relations
    */
   public $hasOne = [];
   public $hasMany = [];
   public $belongsTo = [];
   public $belongsToMany = [];
   public $morphTo = [];
   public $morphOne = [];
   public $morphMany = [];
   public $attachOne = [];
   public $attachMany = [];
  }
```

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -2,6 +2,7 @@
 
 use Lang;
 use Input;
+use App;
 use October\Rain\Database\ModelException;
 use Illuminate\Support\MessageBag;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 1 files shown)


src/Database/Traits/Validation.php
```diff
@@ -75,12 +76,19 @@ public static function bootValidation()
...
      */
-    protected static function makeValidator($data, $rules, $customMessages, $attributeNames)
+    protected static function makeValidator($data, $rules, $customMessages, $attributeNames,$connection = null)
     {
-        return Validator::make($data, $rules, $customMessages, $attributeNames);
+        $validator = Validator::make($data, $rules, $customMessages, $attributeNames);
+        if ($connection !== null) {
+           $verifier = App::make('validation.presence');
+           $verifier->setConnection($connection);
+           $validator->setPresenceVerifier($verifier);
```
daftspunk: `$verifier` is a singleton and should already be set?

tschallacka:
It might not be set on models that use a different connection? Im not sure why it works this way like it does. I just followed the example codes and reimplemented it and it worked for me? So im not sure whats going on behind the scenes in the app management.

LukeTowers:
@daftspunk What needs to be done to get this merged? @mbaxter91288 is this the issue that you were running into a while ago?

mbaxter91288:
@LukeTowers yeah it was

daftspunk:
@LukeTowers I wasn't able to understand how it works, now that this has been peer reviewed we can accept it with confidence.

Cheers guys

LukeTowers:
@mbaxter91288 Does your use case work now without the hack you ended up using? I can't remember what you landed on for a solution to that ATM, but let me know if this works for you now

PR #143, opened 2015-11-05
Validator obey changed primary key name

tschallacka:
If the validator doesn't get supplied the primary key name it won't
obey the unique filter on update if the primary key name is different
from `id`.
This way if for some reason primary key is changed via `$primaryKey`
the unique filter will still work by doing `AND WHERE "otherkey" <> 2`
instead of `AND WHERE "id" <> 2`


daftspunk:
LGTM!

tschallacka:
\o/ wooo

PR #145, opened 2015-11-07
Fix "file exists" error on concurrent uploads

jwilson8767:
This should resolve issues where concurrent uploads return the error `mkdir(): file exists.` by taking into consideration the possibility of the race condition.


jwilson8767:
Fix based on solution posted [here](http://stackoverflow.com/questions/19964287/mkdir-function-throw-exception-file-exists-even-after-checking-that-directory) by Ka.

daftspunk:
Thanks!

PR #146, opened 2015-11-07
Add parameter type enforcement to ValidationException. closes octobercms/october#1253

jwilson8767:
As noted in octobercms/october#1253, bad parameters given to ValidationException are ignored and should not be.


daftspunk:
Good job, thanks!

PR #148, opened 2015-11-16
Custom filename for thumb images

fuunnx:
Add the possibility to give a custom filename to the generated thumbs which can be usefull for seo purpose


daftspunk:
Looks alright, just need you to write some documentation on how to use it. Feel free to just write it here and we can find a good spot for it in the docs.

PR #149, opened 2015-11-24
Get for Height and Width and Resize images on upload to default maxWidht and Heigth 

estudiogenius:
Now ... we can specify some settings for sending images to Storage in config / cms.php as follows:
    'storage' => [
        'uploads' => [
            'disk'   => 'local',
            'folder' => 'uploads',
            'path'   => '/storage/app/uploads',
            'image'  => [
                'maxWidth' => 1900,
                'maxWHeight' => 1200,
                'quality' => 85,
            ],
        ],
...
PS: this update will not change the default config file...

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -1,5 +1,6 @@
 <?php namespace October\Rain\Database\Attach;
 
+use Config;
 use Storage;
 use File as FileHelper;
 use October\Rain\Database\Model;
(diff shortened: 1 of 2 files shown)


pedromazala:
+1

wilcorrea:
+1

Belluzzo:
+1

daftspunk:
This manipulates every image upload? I'm not sure why this is necessary, this functionality should be introduced as a plugin with a simple event used here instead.

PR #150, opened 2015-11-24
Make getContentType() public

vojtasvoboda:
When I want to send file as email attachment, it would be nice to have `->getContentType()` method available:

```php
$attachment = $record->image; // System\Models\File
Mail::send($template, $templateParameters, function($message) use ($email, $attachment) {
    $params = ['as' => $attachment->getFilename(), 'mime' => $attachment->getContentType()];
    $message->attach($attachment->getLocalPath(), $params);
    $message->to($email);
}
```

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -207,7 +207,7 @@ class File extends Model
     /**
      * Returns the file content type.
      */
-    protected function getContentType()
+    public function getContentType()
     {
         if ($this->content_type !== null) {
             return $this->content_type;


daftspunk:
Should be available as:
```
$params = ['as' => $attachment->file_name, 'mime' => $attachment->content_type];
```

vojtasvoboda:
Yes it works! Thank you. But even so, I think it could be public (as other getters).

snimek obrazovky 2016-01-11 v 15 08 55

PR #151, opened 2015-12-03
Fixed sorting of replacements in Translator / Created tests

tobias-kuendig:
The replacements in Translation/Translator were ordered by value instead of by key. This caused some errors when replacing variables like `to` and `total` that begin with the same letters.

https://github.com/octobercms/library/blob/master/src/Translation/Translator.php#L187
```
// Needs to be sorted by key, not value

dump($replace);

#items: array:3 [
    "from" => 1
    "to" => 0
    "total" => 0
]
```

See also https://github.com/octobercms/october/issues/793

### DIFF
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -184,8 +184,8 @@ class Translator implements TranslatorInterface
      */
     protected function sortReplacements(array $replace)
     {
-        return (new Collection($replace))->sortBy(function($r) {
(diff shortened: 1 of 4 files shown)


keiosweb:
:+1:

daftspunk:
Merged in 262b46f1bf467d1ff8c7c933f4a4bfa19ac4859f

Thanks!

PR #152, opened 2015-12-04
Don't drop additional attributes in unique validation rule

tobias-kuendig:
Unique validation rules that have additional attributes (`idColumn` or `where`) were only working as expected when creating entries. If a entry existed, these additional attributes got dumped during processing.

```php
# Example Rule 
unique:table,column,key,id,whereCol,5

# Got truncated to
unique:table,column,key
```


daftspunk:
Merged in 2e9d938ffcd99c98ff6dc8e7648f3d9c7c326244
Then overhauled in 077f3e92f5e8b852e1e184fb6074dd9f403a7260

Thanks!

PR #154, opened 2015-12-18
Remove numeric index in array for jsonable if all the index for array  

brenodouglas:
it's array:
```php
$array = ['1' => 'foo', '2' => 'bar'];
```

In database result in this json: 
```json
{"1": {"hour":"12:00"}, "2": {"hour": "13:00"}}
```
But now the result will be:
```json
[{"hour": "12:00"}, {"hour":"13:00"}]
```
in the case of associative array the number will come as the first case.

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1266,7 +1266,7 @@ class Model extends EloquentModel
 
         // Handle jsonable
         if (in_array($key, $this->jsonable) && (!empty($value) || is_array($value))) {
-            $value = json_encode($value);
+            $value = json_encode(array_values($value));
         }
 
         // Handle direct relation setting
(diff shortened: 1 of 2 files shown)


jwilson8767:
resolves octobercms/october#1662

daftspunk:
Not really sure this is such a good idea, see #1662 for a solution.

brenodouglas:
@daftspunk Just tried your solution and my result is:

```
"array_values() expects parameter 1 to be an array or collection" on line 49 of /home/vagrant/Code/CMS/plugins/afipe/cineteatro/models/Exibicao.php
```

This occurs when my model Exibicao is one relation and It is within a relationship popup

my method following your solutions with error above:
```php
public function setHorariosAttribute($value)
{
        $this->attributes['horarios'] = json_encode(array_values($value));
}
```

daftspunk:
$this->attributes['horarios'] = $this-> fixKeysNumber($value);

PR #160, opened 2016-01-03
Laravel 5.1 php 7

jwilson8767:


daftspunk:
Confirmed all of these fixes have been introduced in `stable`

Thanks!

PR #174, opened 2016-01-24
Fix issue with count relations not being given prefixes

jwilson8767:
dshoreman found the problem, here is the fix.


jwilson8767:
Closed as dshoreman will include with other prefix fixes in a different pr.

PR #178, opened 2016-01-28
unicode character in slugs

sajjad-ser:
please support Unicode characters in slugs and urls

### DIFF
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -104,4 +104,24 @@ class Str extends StrHelper
         return Html::clean($string);
     }
 
+
+    /**
+     * Generate a URL friendly "slug" from a given string.
+     *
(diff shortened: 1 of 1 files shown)


daftspunk:
Why does Laravel not do this? Also this needs unit tests.

sajjad-ser:
I sent a pull request in laravel. they told me if i want to do this i can extend the class for my self.
I will do this modification in persian plugin and use it in my projects with no problem.

LukeTowers:
@sajjad-ser Are you willing to put together a unit test for this so we could merge it? If not, let me know

daftspunk:
Laravel appear to have fixed this in v5.5, which we are upgrading to soon. 

```
Str::slug('foo bar', 'en');
```

PR #186, opened 2016-02-12
Add a new event - backend.auth.afterLogin

filipac:
I wanted to listen to an event when a users logs-in and I realised that October does not yet have that. It would be great to support this out of the box to avoid users make changes to core files or listen to hack-ish events to do something after login.

### DIFF
--- a/src/Auth/Models/User.php
+++ b/src/Auth/Models/User.php
@@ -7,6 +7,7 @@ use InvalidArgumentException;
 use RuntimeException;
 use Exception;
 use DateTime;
+use Event;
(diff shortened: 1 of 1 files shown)


tschallacka:
+1

daftspunk:
I think you want this to go in the Backend User model? See RainLab.User for example.

filipac:
I opened a new PR on backend module, i moved the event (and renamed it) to apply only on the backend User model. https://github.com/octobercms/october/compare/master...filipac:patch-1

PR #187, opened 2016-02-13
User model implements the Authenticatable interface

gpasztor87:


daftspunk:
Why?

gpasztor87:
I wanted to implement the `tymon/jwt-auth` package for my Api plugin, and it's need to the User model implements the Authenticatable interface [link](https://github.com/tymondesigns/jwt-auth/blob/a24c66ab7c18e65cc7a6fee58afe2e704eb2ebef/src/JWTGuard.php#L65).

But it's now not important.

PR #193, opened 2016-03-14
Added GeneratorCommand for template generators.

gpasztor87:
The template generator commands now extends the GeneratorCommand class.
Result:
- No need extra class for registering the stubs.
- Easy to make custom template generators from plugins.


daftspunk:
Looks great

gpasztor87:
Thank you!

PR #199, opened 2016-03-24
getPath() already return full url

nicolas-lescop:
We need this fix for making the campaign manager plugin to work.

### DIFF
--- a/src/Parse/Syntax/SyntaxModelTrait.php
+++ b/src/Parse/Syntax/SyntaxModelTrait.php
@@ -83,7 +83,7 @@ trait SyntaxModelTrait
             $path = $image->getPath();
         }
 
-        return Request::getSchemeAndHttpHost()  . $path;
+        return $path;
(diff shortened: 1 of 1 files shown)


daftspunk:
Please explain the full details of the issue. Why is this necessary? What does it fix? How does someone else reproduce the bug?

ariponce:
The problem occurs when using S3 for the uploads, as the config returns the url for the bucket, and the `getThumbForImage` method adds the current url to the path, causing a malformed url.
This pull request doesn't really solve the problem, as a local upload configuration would fail.

I think the correct fix would be to only add the scheme and HTTP host on the `getThumbForImage` method if one was not returned from the `getPath` method. That way both an S3 and local config would work.
Something like this:

`if (preg_match("/(\/\/|http|https)/", $path)) {
    return $path;
}`

daftspunk:
Thanks @ariponce

PR #201, opened 2016-03-30
Added activate_gzip helper

mahony0:


daftspunk:
This is too specific, gzip should be controlled by the web server configuration. See https://github.com/octobercms/october/pull/1885 for more notes.

PR #208, opened 2016-04-18
Fix getting zero value

vojtasvoboda:
Fix getting zero value to form input. Zero is also value. When I have zero value in my database, form looks like that:

Before fix: `<input name="test" value="">`

After fix: `<input name="test" value="0">`.


daftspunk:
Thanks.

Please check out 71799e9920bcedba77dde3c9561842c6d12ae2af to make sure it still functions as you expect.

PR #211, opened 2016-04-25
Minor improvements to the scaffolding stubs

adrenth:
Minor improvements to the scaffolding stubs.

- Added newline to end of each file
- Added translation message to controller view file stubs (create / update)
- Added code hinting to migration stubs (Blueprint $table)

It would be nice that everything is translatable in the stubs. I did not change everything.

Any ideas on how to get the stubs fully multilingual?


daftspunk:
This has come up before. We'd rather not include translations in these by default, they should be as simple as possible.

adrenth:
Ok, I reverted those changes. I can see why you want to keep things simple. That's the whole philosophy of OctoberCMS.

I create plugins on a daily basis, and need to change the generated templates a lot. Because a lot of our user base's main language is Dutch. But let that be a thing for the future :-)

I really like to see the "Blueprint" typehint in hintsrc/Scaffold/Console/model/create_table.stub, that does a bunch of good things to code completion in PhpStorm.

Let me know if this is OK, i'm happy to revert or change things you don't like.

src/Scaffold/Console/model/create_table.stub
```diff
@@ -1,15 +1,14 @@
...
 use October\Rain\Database\Updates\Migration;
 
 class Create{{studly_plural_name}}Table extends Migration
 {
-
     public function up()
     {
-        Schema::create('{{lower_author}}_{{lower_plugin}}_{{snake_plural_name}}', function($table)
-        {
+        Schema::create('{{lower_author}}_{{lower_plugin}}_{{snake_plural_name}}', function (Blueprint $table) {
```
dshoreman: Blueprints have been added to the migration stub as part of #212 which switches over to `October\Rain\Schema\Blueprint`

daftspunk:
I'm not sure what this PR actually does now...

As a general rule of thumb, non-essential "clean up" should be performed as part of functional changes. To prevent commit spam.

PR #212, opened 2016-04-27
Make timestamps() create nullable timestamps

dshoreman:
Overrides the default Illuminate Blueprint class to implement a fix
that was added in 5.2 but never backported to the LTS version we use.

More info: https://dev.mysql.com/worklog/task/?id=7467
In particular, see F-8 under the Requirements tab.

Also adds a Schema facade with blueprint injector

*Source: http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes*

### DIFF
--- a/src/Support/Facades/Schema.php
+++ b/src/Support/Facades/Schema.php
@@ -1,0 +1,39 @@
+<?php namespace October\Rain\Support\Facades;
+
+use October\Rain\Support\Facade;
+use October\Rain\Database\Schema\Blueprint;
+
+class Schema extends Facade
+{
+    /**
+     * Get a schema builder instance for a connection.
+     *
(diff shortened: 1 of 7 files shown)


src/Support/Facades/Schema.php
```diff
@@ -0,0 +1,39 @@
...
+     * Get a schema builder instance for the default connection.
+     *
+     * @return \October\Rain\Database\Schema\Builder
+     */
+    protected static function getFacadeAccessor()
+    {
+        return static::getBuilder();
+    }
+
+    protected static function getBuilder($connection = null)
```
daftspunk: Dirty

daftspunk:
Excellent work

PR #214, opened 2016-05-23
Fix `Fatal error: Call to a member function lists() on array` when using HasMany relationship

tobias-kuendig:
Using Build 335, when editing a model in the backend that has a `hasMany` relation, a fatal error is thrown.

This is beacause the `$this->parent->$relationName->lists()` call below returns an array, not a collection/relation.

auswahl_134

### DIFF
--- a/src/Database/Relations/HasMany.php
+++ b/src/Database/Relations/HasMany.php
@@ -81,7 +81,8 @@ class HasMany extends HasManyBase
         $relationName = $this->relationName;
 
         if ($this->parent->$relationName) {
-            $value = $this->parent->$relationName[$this->localKey];
+            $relation = $this->parent->$relationName;
+            $value = $relation[$this->localKey];
         }
 
         return $value;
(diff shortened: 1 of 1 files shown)


daftspunk:
As per the contribution guidelines, please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

tobias-kuendig:
I have created a test plugin to demonstrate the issue: https://github.com/tobias-kuendig/oc-hasmanytest

Install it, go to `Products` and click on `Click me to trigger error`.

The product model uses a `hasMany` relation:

```php
namespace OFFLINE\Hasmanytest\Models;

use Model;

class Product extends Model
{
    public $hasMany = [
        'attributes' => 'OFFLINE\HasManyTest\Models\Attribute'
    ];
}
```

On the `update product` page there is a partial embeded, that renders the attributes relation.

daftspunk:
Hey @tobias-kuendig , I think this might be caused by the relationship name `attributes`. It also exists as a property on every model (containing the model attributes) and this is why an array is returned, throwing the error. Install the [test plugin](https://github.com/daftspunk/oc-test-plugin) and navigate to Posts and see that a Post hasMany Comments, there is no issue in this example. Try using a different relationship name instead of "attributes".

PR #220, opened 2016-07-11
Change asset compiling

christianWilling:
hi this is the scss filter https://github.com/octobercms/october/pull/2180 i hope this work for you

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,9 @@
         "symfony/yaml": "2.8.*|3.0.*",
         "twig/twig": "~1.23|~2.0",
         "league/csv": "~8.0",
-        "jenssegers/date": "~3.0"
+        "jenssegers/date": "~3.0",
+        "leafo/lessphp": "^0.3.7",
+        "leafo/scssphp": "~0.1"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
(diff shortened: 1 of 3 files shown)


daftspunk:
We had some issues with the leafo/less package because it was no longer being maintained, the `oyejorge/less.php` package appears to work better. Refs: https://github.com/octobercms/october/issues/179

Is it possible to adapt this code to the oyejorge package instead?

daftspunk:
Please make sure your PRs are submitted to the `develop` branch.

christianWilling:
yes i have created an new pr https://github.com/octobercms/library/pull/225

amdad:
Thanks for this!

What about including optionally _compass_ to use in theme? What will be the easiest way to do this in new compiler?

christianWilling:
hi compass isn't installed on the most server thats the reason why it should work in php

amdad:
Yes, I just ask about this PHP implementation. What will be the easiest way to have this in October since SCSS finally be in core. Will you write some plugin if I ask nicely? :) Or some other way, to import compass elements. Modularity is also important. Don't need to have all at once.

amdad:
@christianWilling ping you to check this topic.

daftspunk:
Compass can be included manually in SCSS as far as I know. Have a look at this project here for an example:

https://github.com/daftspunk/utility/tree/1.0-stable/scss/frameworks

Note the `_compass.scss` file, it is probably an older version but it works.

amdad:
Thank U!

PR #222, opened 2016-07-14
 Use Eloquent save() method rather than newQuery() method in Sortable

manogi:
…Trait to trigger save and update event.

I would like the save and update events to fire when sorting with the Sortable trait - this does not seem to be the case with the newQuery() update() method, so I propose to change that to the Eloquent save() model. Hope I make sense.

### DIFF
--- a/src/Database/Traits/Sortable.php
+++ b/src/Database/Traits/Sortable.php
@@ -54,9 +54,10 @@ trait Sortable
             throw new Exception('Invalid setSortableOrder call - count of itemIds do not match count of itemOrders');
         }
 
+        $order_column = $this->getSortOrderColumn();
+
         foreach ($itemIds as $index => $id) {
             $order = $itemOrders[$index];
-            $order_column = $this->getSortOrderColumn();
             $instance = $this->find($id);
             $instance->$order_column = $order;
             $instance->save();
--- a/src/Database/Traits/Sortable.php
+++ b/src/Database/Traits/Sortable.php
@@ -56,7 +56,10 @@ trait Sortable
 
         foreach ($itemIds as $index => $id) {
             $order = $itemOrders[$index];
-            $this->newQuery()->where('id', $id)->update([$this->getSortOrderColumn() => $order]);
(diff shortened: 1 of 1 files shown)


daftspunk:
Hmm, yes this is for performance reasons. The proposed change is 2x more expensive. We can look to the `NestedTree` trait which also performs modifications using `update` / QueryBuilder, this proposed change would then make it inconsistent.

Then we can refer to `SoftDelete` trait that also uses `update`, however it brings a custom event with it (beforeDelete/afterDelete). It is a tough decision, can you provide more information about your event and what it does, why it is needed. This will help me understand where it should sit in the logic hierarchy. 

Reference notes: There are several places where the logic may be injected.

1. Sort AJAX handler in the controller
1. Model save/update events (PR wants this)
1. Custom "Tree" model event (non-existent today)

manogi:
I like the idea of just firing the afterSave event at the end of the setSortableOrder method instead of my original PR. Should I make a new PR with this added at the end of that method (I tested it, it does what I need at least):
```
    $this->fireEvent('model.afterSave');
    if ($this->methodExists('afterSave')) {
        $this->afterSave();
    }
```

manogi:
OR I could add something to Backend\Behaviors\ReorderController like this, which also works and would work for Sortabel and NestedTree both:

```
$model->fireEvent('model.afterSave');
if ($model->methodExists('afterSave')) {
    $model->afterSave();
}
```

This seems the better solution.

manogi:
Hmm, the NestedTree trait seems to bind to the afterSave event, so there would probably a problem if it also fires it... you probably have a better understanding of how to fire the afterSave event in the right place, I'm not sure I have a deep enough understanding of how these things should be implemented.

manogi:
To answer your question: I listen to these events to clear some query caches when certain changes are made.

manogi:
I close this PR as I now realize it is not really what I was about (which is firing an event when sorting). I will open another one that makes more sense in that direction.

PR #226, opened 2016-07-27
Support for handling proxies

JoelESvensson:
See: https://github.com/octobercms/october/pull/2250

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "symfony/yaml": "2.8.*|3.0.*",
         "twig/twig": "~1.23|~2.0",
         "league/csv": "~8.0",
-        "jenssegers/date": "~3.0"
+        "jenssegers/date": "~3.0",
+        "fideloper/proxy": "^3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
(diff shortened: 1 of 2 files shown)


daftspunk:
Looks like a nice elegant solution. This should really be handled as a plugin to minimise the bloat on the platform. What do you think?

JoelESvensson:
I definitely agree. The problem is that I don't know how to programatically push the middleware to the global stack with a plugin.

daftspunk:
Take a look here for details on how to do it: https://github.com/octobercms/docs/pull/167

JoelESvensson:
Ah, thanks! Is `CmsController` where you would put global middleware?

daftspunk:
Prepending or appending to the Http contract (see the PR changes) would make it truly global. Attaching to `CmsController` or `BackendController` would affect that instance only.

JoelESvensson:
Closing this in favor of the plugin
https://github.com/JoelESvensson/october-trusted-proxy

JoelESvensson:
Unfortunately the plugin solution doesn't work on paths where plugins are disabled. More specifically the ones here: https://github.com/octobercms/october/blob/master/modules/system/ServiceProvider.php#L115

How do you think I should solve this @daftspunk?

daftspunk:
@JoelESvensson, this is an artificial wall to stop plugins from breaking critical functionality. Set this property to `true` in your plugin information class to overcome it.

    /**
     * @var boolean Determine if this plugin should have elevated privileges.
     */
    public $elevated = true;

PR #227, opened 2016-07-29
Network: fix auth

rounce:
strpos never returns null


daftspunk:
Please resubmit to `develop` branch

PR #229, opened 2016-08-02
Ensure that fileContent is an array

acasar:
This fixes https://github.com/octobercms/october/issues/2274

In some cases the ``file()`` method fails and returns ``false`` which causes array_slice to fail (line 179) and only Error 500 is returned to the user.

```
array_slice() expects parameter 1 to be array, boolean given in ExceptionBase.php:179
```

With this change we ensure that ``fileContent`` property is always an array.


daftspunk:
Great find!

PR #230, opened 2016-08-04
fix performMove update sql when using Uuid as the primary key.

kuloed:
related to octobercms/october#2268

### DIFF
--- a/src/Database/Traits/NestedTree.php
+++ b/src/Database/Traits/NestedTree.php
@@ -883,7 +883,7 @@ trait NestedTree
             ELSE $wrappedRight END";
 
         $parentSql = "CASE
-            WHEN $wrappedId = $currentId THEN $parentId
+            WHEN $wrappedId = '$currentId' THEN '$parentId'
             ELSE $wrappedParent END";
 
         $result = $node->newQuery()


daftspunk:
Incorrect integer value: 'NULL' for column 'parent_id' at row 9

SQL:

    update `october_test_channels` set `nest_left` = CASE
        WHEN `nest_left` BETWEEN 1 AND 15 THEN `nest_left` + 17 - 15
        WHEN `nest_left` BETWEEN 16 AND 17 THEN `nest_left` + 1 - 16
        ELSE `nest_left` END, `nest_right` = CASE
        WHEN `nest_right` BETWEEN 1 AND 15 THEN `nest_right` + 17 - 15
        WHEN `nest_right` BETWEEN 16 AND 17 THEN `nest_right` + 1 - 16
        ELSE `nest_right` END, `parent_id` = CASE
        WHEN `id` = '9' THEN 'NULL'
        ELSE `parent_id` END, `updated_at` = 2016-08-06 00:31:16 where (`nest_left` between 1 and 17 or `nest_right` between 1 and 17) order by `nest_left` asc

kuloed:
oops sorry for that.
i make another pull request because i accidentally deleting in my remote branch.

see #232

PR #233, opened 2016-08-06
fix performMove update sql when using Uuid as the primary key.

kuloed:


gabsource:
I came to this issue too, patched locally with te code below for the moment.

Some database engines may not like matching integer to strings so it might be better to only escape the values if they are not an integer.

The espace code comes from the framework `Grammar::wrapValue` (protected so can't call it directly).

```
        if ($parentId === null) {
            $parentId = 'NULL';
        } elseif ( ! is_int($parentId)) {
            $parentId = '"'.str_replace('"', '""', $parentId).'"';
        }

        $currentId = $node->getKey();

        if ( ! is_int($currentId)) {
            $currentId = '"'.str_replace('"', '""', $currentId).'"';
        }
```

daftspunk:
@gabsource In other areas we use `DbDongle::cast(..., "TEXT");` to ensure string:string comparison.

kuloed:
Sorry for the very late response, i already change the code like @daftspunk suggest.

src/Database/Traits/NestedTree.php
```diff
@@ -855,22 +856,26 @@ protected function performMove($node, $target, $position)
...
 
-        if ($parentId === null)
+        if ($parentId === null) {
             $parentId = 'NULL';
+        } else {
+            $parentId = $pdo->quote($parentId);
+        }
 
-        $currentId = $node->getKey();
+        $currentId = $pdo->quote($node->getKey());
```
daftspunk: This needs to be cast to TEXT to ensure a "like for like" comparison.

kuloed: i already quote the value, so it will make the value as string right?

daftspunk: Oh I see, touche!

LukeTowers:
@daftspunk What's required to merge this?

LukeTowers:
@daftspunk bump

PR #234, opened 2016-08-06
Order Sortable records by default

acasar:
Not sure if this is a bug or not, but when I added a ``Sortable`` trait to the model I expected the results to be ordered by default, which they were not.

So I created this PR which adds a global scope to fix that. There is also an option to easily override the default order by just calling the ``orderBy`` as per usual, which also helps to maintain backwards compatibility. I added some tests to show how this all works.

If ``Sortable`` wasn't intended to work this way feel free to close this PR.

### DIFF
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -14,6 +14,12 @@ use Illuminate\Database\Eloquent\Builder as BuilderModel;
  */
 class Builder extends BuilderModel
 {
+    /**
+     * An array of flags.
+     *
+     * @var array
+     */
+    protected $flags = [];
 
     /**
      * Get an array with the values of a given column.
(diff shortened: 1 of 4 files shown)


acasar:
@daftspunk I just saw https://github.com/rainlab/translate-plugin/blob/master/behaviors/TranslatableModel.php#L50

Having a similar problem in this PR as well. I need to check if the scope was already removed before attempting to remove it again. Right now I'm doing it with a property on the scope, which is essentaly a global state and can break in some cases.

Maybe we could add some mechanism to the Eloquent ``Builder`` class that would allow to set some flags? Like this:

```php
// apply scope
// ...
$builder->setFlag('sortable_scope_applied');

// ...

if(! $builder->hasFlag('sortable_scope_applied')) {
    // remove scope
    // ...
    $builder->unsetFlag('sortable_scope_applied')
} 
```

This could also solve the issue in the translate plugin.

daftspunk:
Thanks @acasar!

acasar:
@daftspunk I went ahead and added this "flagging" mechanism. If you have any remarks let me know.

daftspunk:
Sorry, I've just had time to look at this. The flag functionality is probably too uncommon to be added to the Builder. The first commit with `protected $scopeApplied;` is best, instead of over engineering it. Also I think there is some improvements coming in Laravel 5.3 that might affect this, so to be on the safe side let's go with the KISS approach.

daftspunk:
Merged in dc739761e81b10df1c997e2adef84cb75e9854e8

PR #235, opened 2016-08-18
Add support for nested repeaters in syntax parser

ternavsky:
It would be great if syntax parser could support nested repeaters. Especially given the fact that Form Widget supports them.

### DIFF
--- a/src/Parse/Syntax/FieldParser.php
+++ b/src/Parse/Syntax/FieldParser.php
@@ -73,25 +73,11 @@ class FieldParser
      */
     protected function processTemplate($template)
     {
-        // Process repeaters
-        list($template, $repeatTags, $repeatfields) = $this->processRepeaterTags($template);
+        // Process all tags
+        list($tags, $fields) = $this->processAllTags($template);
 
-        // Process registered tags
-        list($tags, $fields) = $this->processTags($template);
         $this->tags = $this->tags + $tags;
         $this->fields = $this->fields + $fields;
-
-        /*
-         * Layer the repeater tags over the standard ones to retain 
-         * the original sort order
-         */
-        foreach ($repeatfields as $field => $params) {
-            $this->fields[$field] = $params;
-        }
-
-        foreach ($repeatTags as $field => $params) {
-            $this->tags[$field] = $params;
-        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
Support is tentative. As per the contribution guidelines, please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

LukeTowers:
@ternavsky Please provide code examples (ideally in the form of a PR to the [test plugin](https://github.com/daftspunk/oc-test-plugin/)) that demonstrate the current issue and support your provided fix.

daftspunk:
Looked nice. Shame we couldn't see it in action. This needs replication instructions or a test case we can use as a point of reference to confirm something has actually been fixed.

LukeTowers:
@daftspunk this was for https://github.com/rainlab/pages-plugin/issues/255. @bennothommo do you think you could take a stab at testing this out and perhaps adding some unit tests for it?

Here's an example of how someone would use it in the wild:
```twig
<div class="sidebar-content">
        {repeater name="sidebar_content" prompt="Add a section" tab="Sidebar Content"}
            <div class="sidebar-section">
                <h2 class="sidebar-section-title">{text name="section_title" label="Title" placeholder="Annual Reports"}{/text}</h2>
                {repeater name="links" prompt="Add a link" label="Sidebar Links"}
                    <a class="sidebar-link" target="_blank" href="{text name="link_url" label="URL" placeholder="https://example.com/" span="left"}{/text}">
                        {text name="link_label" label="Label" placeholder="2018-2019" span="right"}{/text}
                    </a>
                {/repeater}
            </div>
        {/repeater}
    </div>
```

The above works if you use nested repeaters to store to twig variables, and then loop over the variable data to render yourself but the actual auto rendering that should happen with that structure fails because the inner repeater tag doesn't get rendered.

LukeTowers:
> The above works if you use nested repeaters to store to twig variables, and then loop over the variable data to render yourself but the actual auto rendering that should happen with that structure fails because the inner repeater tag doesn't get rendered.

Ignore that, the above example doesn't work at all in the backend or frontend, it ignores the nested repeater and just treats the nested repeater's fields as its own. But I would like it to support that syntax.

bennothommo:
@LukeTowers I've taken a bit of a look at this - basically at this point, it can't be done, as we're using regular expression matching for the tags. While regular expressions can be recursive, it is extremely difficult for it to work out correct nesting for matches.

In your example above, I've found it matches the outer `{repeater}` to the closing tag of the inner `{repeater}`, which is obviously incorrect. Everything I've read on the net about recursive matches strongly suggest we shouldn't use them for parsers.

If we wanted to do this properly, we'd likely have to go down the route of either parsing the code line by line and detecting nesting of repeaters, or even looking to implementing a node-based parser. Both are likely substantial amounts of work.

LukeTowers:
@bennothommo what did this PR actually do then?

bennothommo:
@LukeTowers looks like it went down the recursive regular expression route. I'll give it a test sometime soon and see if it worked as advertised.

bennothommo:
Well colour me impressed, it actually does work, in that it extracts the nested repeaters and nests them correctly in the fields config. I would say the reason it did not work for https://github.com/rainlab/pages-plugin/issues/255 is because the Pages plugin probably doesn't look into nested fields beyond the one level for repeaters.

bennothommo:
I'm going to roll this out to a new PR as I cannot merge `develop` due to diverged histories, and there's a little bit of cleanup required, as well as adding the unit tests. @ternavsky, I will ensure you get credit for your original fix.

PR #236, opened 2016-08-28
Override runningInConsole() method

alvaro-canepa:
In some cases `php_sapi_name()` returns `cgi-fcgi` in both CLI and Web.
Source: [Drupal 7 - drupal_is_cli](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_is_cli/7.x)

### DIFF
--- a/src/Foundation/Application.php
+++ b/src/Foundation/Application.php
@@ -173,6 +173,15 @@ class Application extends ApplicationBase
     {
         return $this->executionContext == 'back-end';
     }
+    
+    /**
+     * Determine if we are running in the console.
+     *
+     * @return bool
+     */
+    public function runningInConsole() {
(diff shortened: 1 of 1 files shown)


daftspunk:
I would be curious to see what Laravel has to say about this, [considering they are the owners of this logic](https://github.com/laravel/framework/blob/29923153f049bd1006953a74388511eda729318d/src/Illuminate/Foundation/Application.php#L510-L518).

LukeTowers:
@alvaro-canepa Please detail under what conditions `php_sapi_name()` returns `cgi-fcgi` in both web and console. If this is a bug that October will encounter, then we can look at merging this fix - although if it really is an issue, then please bring it up on the [Laravel repository](https://github.com/laravel/framework)

PR #237, opened 2016-09-06
The optimization of small

happyelevenday:
remove the repeat custom extension
remove curl_init test replace by composer require field


daftspunk:
This part might be accepted. 

> remove the repeat custom extension

The rest not, since composer is a second class citizen in October, not everybody uses it. Can you describe when a repeat custom extension would occur?

As per the contribution guidelines, please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

PR #241, opened 2016-09-29
Add checkbox support to parser

acasar:
This PR adds support for ``checkbox`` field to the October syntax parser. This can be very useful when using "Static Pages" plugin where we can allow users to toggle display of certain dynamic layout sections.

Example:
```
{checkbox name="showTitle" label="Show page title"}
    <h1>{{ page.title }}</h1>
{/checkbox}
```

Parsing to ``Twig`` is supported, but I don't think that ``Brackets`` parser currently supports conditional statements, so I didn't find a way to support it there.

### DIFF
--- a/src/Parse/Syntax/FieldParser.php
+++ b/src/Parse/Syntax/FieldParser.php
@@ -49,6 +49,7 @@ class FieldParser
         'mediafinder',
         'dropdown',
         'radio',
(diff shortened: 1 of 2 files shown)


daftspunk:
Very cool. Have wanted this for a while! :+1:

PR #242, opened 2016-10-16
Replacing text in the language constants

PopcornPHP:
When create a plugin - I tired of constantly change the words on the language constants, which are set out in the core

### DIFF
--- a/src/Scaffold/Console/controller/create.stub
+++ b/src/Scaffold/Console/controller/create.stub
@@ -6,9 +6,7 @@
 <?php Block::endPut() ?>
 
 <?php if (!$this->fatalError): ?>
-
     <?= Form::open(['class' => 'layout']) ?>
-
         <div class="layout-row">
             <?= $this->formRender() ?>
         </div>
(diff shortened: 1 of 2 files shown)


src/Scaffold/Console/controller/create.stub
```diff
@@ -20,29 +18,23 @@
...
                 </span>
             </div>
         </div>
-
     <?= Form::close() ?>
-
 <?php else: ?>
-
     <p class="flash-message static error"><?= e($this->fatalError) ?></p>
     <p><a href="<?= Backend::url('{{lower_author}}/{{lower_plugin}}/{{lower_name}}') ?>" class="btn btn-default">Return to {{lower_title_name}} list</a></p>
```
LukeTowers: While you're at it, want to convert the rest of the hard coded strings to translation strings?

daftspunk:
This has come up before, we prefer to keep the stubs in English for the sake of simplicity. Try the [Scaffold Translation](https://octobercms.com/plugin/bnb-scaffoldtranslation) plugin.

PR #244, opened 2016-11-09
Try to make a valid object before json encode

H6:
Check if provided `$value` is a string and try to convert it into a object, so that the `json_encode` will have a valid object to encode.

PS: This problem was related to a [form-builder post in the forum](https://octobercms.com/plugin/support/renatio-formbuilder/problem-with-validationinvalid-argument-applied-to-foreach). Anyway after this "first try to decode" and then "encode" flow, it worked.

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1303,6 +1303,11 @@ class Model extends EloquentModel
         if (($_value = $this->fireEvent('model.beforeSetAttribute', [$key, $value], true)) !== null)
             $value = $_value;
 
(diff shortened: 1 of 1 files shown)


H6:
I have added an issue for this... https://github.com/octobercms/october/issues/2473

daftspunk:
Waiting on result of #2473

LukeTowers:
@daftspunk octobercms/october#2473 has been resolved, this could be closed now.

PR #246, opened 2016-11-25
Pass parent model to relation scopes

LukeTowers:
This improves the extensibility of relationships by passing the parent relation model to the query scope that will be applied to relation. It allows the use of attributes of the parent relation model in the query scope applied to the relation.

This is a mirror of octobercms/october#2419 and octobercms/october#2496, except that it effects scopes applied in the relationship configuration itself instead of the recordfinder or relationcontroller widgets. If necessary, I can create a case in the test plugin, but if this is simple enough with the added reference to the prior PR to not require a case in the test plugin, that would be simpler for me :)


daftspunk:
LGTM

PR #250, opened 2016-12-27
Ensured firing of model.setAttribute event when mutator exists

mnishihan:
If we return the returned value of the mutator call straight away, that will disable the firing of 
**model.setAttribute** event, and then no third party plugins will be able to implement any kind of post processing logic when an attribute is set with explicit mutators. 

Returning from the mutator call will also break method chaining as intended by returning `$this` at the end of this method.

### DIFF
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -750,11 +750,15 @@ class Model extends Extendable implements ArrayAccess, Arrayable, Jsonable, Json
         // the model, such as "json_encoding" an listing of data for storage.
         if ($this->hasSetMutator($key)) {
             $method = 'set'.Str::studly($key).'Attribute';
-
-            return $this->{$method}($value);
+            // If we return the returned value of the mutator call straight away, that will disable the firing of 
+            // 'model.setAttribute' event, and then no third party plugins will be able to implement any kind of 
+            // post processing logic when an attribute is set with explicit mutators. Returning from the mutator 
+            // call will also break method chaining as intended by returning `$this` at the end of this method.
+            $this->{$method}($value);
         }
-
-        $this->attributes[$key] = $value;
+        else {
+            $this->attributes[$key] = $value;
+        }
 
         // After Event
         $this->fireEvent('model.setAttribute', [$key, $value]);


LukeTowers:
Is it common practice to return a value from custom mutators? Just wondering if this will be a breaking change for anyone, and if it will be whether that's a nonstandard use case that we shouldn't worry about.

mnishihan:
I've never seen a setter method to return anything unless `$this` is returned for method chaining. Even though this method is public & can be called from external sources, it is mainly invoked from the `__get` [magic method](https://github.com/octobercms/library/blob/master/src/Halcyon/Model.php#L1551). A quick search for this method invocation, also revealed that only `Backend\Behaviors\RelationController` is calling [this method](https://github.com/octobercms/october/blob/master/modules/backend/behaviors/RelationController.php#L1002) directly in the core and I'm not sure whether the `$newModel` is a instance of `Halcyon/Model` in this case.

Anyway, if backward incompatibility is required, I can update the affected code as follows:

```
        $result = $this;
        if ($this->hasSetMutator($key)) {
            $method = 'set'.Str::studly($key).'Attribute';
            $result = $this->{$method}($value);
        }
        else {
            $this->attributes[$key] = $value;
        }

        // After Event
        $this->fireEvent('model.setAttribute', [$key, $value]);

        return $result;
```

daftspunk:
LGTM, can you switch the branch to `develop`

LukeTowers:
@mnishihan As soon as you switch the branch to develop we can look at getting this merged.

mnishihan:
Switched to develop branch.

BTW, when I created this PR, it seems develop branch was non-existent. But I appreciate the introduction of develop branch, as it creates a clean workflow for verifying & merging PRs. (y)

Lastly, sorry for the later action. I'm not sure how I missed previous notifications for comments of @daftspunk. May be those were lost in piles of other github notification emails I receive from other repositories.

LukeTowers:
@mnishihan No worries, that's why I usually @mention people to remind them about the open issue / pr

@daftspunk Is this good to merge now?

LukeTowers:
@mnishihan The docblock comment for this method mentions that `@return $this`. If you make the changes you commented above for backwards compatibility, that would be great. Unless @daftspunk thinks that it would be better to remove the docblock comment and not return anything.

mnishihan:
@LukeTowers Thanks man :-) 

BTW, w.r.t. the docblock comment of updated method, current code is more accurate, as in my PR `$this` is returned in all cases. But, if backward compatibility is absolutely required (though strongly believe, returning the return value of mutator method is a bug or design flaw and should be corrected ASAP), I'll wait for @daftspunk's confirmation before updating my PR with the backward compatible changes as mentioned in my other comment.

LukeTowers:
Sounds good, just wanted to make sure those two aspects were in alignment. Sounds like they are so as soon as @daftspunk approves this should be good to go

LukeTowers:
@daftspunk I summon thee :)

daftspunk:
LGTM

PR #251, opened 2017-01-05
Small fix to prevent 'non existing fields' throwing erros on save

tschallacka:
This fix is to prevent errors like field xyz doesn't exist in table abc if a "gost"  field is left in the nullable array.

It first checks if the given variable is a string, and then tries to check if the given variable is empty.

This error occurs in databases where fields may be removed but nullable fields not updated and something lingers. 
Another solution could be to explicitly throw a warning about an invalid field in nullable traits, but those checks would be pretty expensive for production code, even if it is developer friendlier to fix legacy code.

I'm not entirely sure if it should be included because errors like these could be cought by unit tests of course, but finding out the root cause of the error might be a hassle.

### DIFF
--- a/src/Database/Traits/Nullable.php
+++ b/src/Database/Traits/Nullable.php
@@ -51,7 +51,7 @@ trait Nullable
     public function nullableBeforeSave()
     {
         foreach ($this->nullable as $field) {
-            if (empty($this->{$field})) {
+            if (is_string($this->{$field}) && empty($this->{$field})) {
                 if ($this->exists) {
                     $this->attributes[$field] = null;
                 }


tschallacka:
Say we have a table with the following specification:

```
public function up()
    {
        Schema::create('tschallacka_geographic_cities', function(Blueprint $table) {
            $table->engine = 'InnoDB';
            $table->increments('id');
            
            $table->timestamps();
            
            $table->string('name',255);
            $table->integer('province_id')->unsigned()->nullable();
            $table->integer('state_id')->unsigned()->nullable();
            $table->integer('country_id')->unsigned()->nullable();
            
            $table->integer('coordinates_id')->unsigned()->nullable();
        });
        
    }
```

now there is a model uses the trait nullable, and it has a remnant from a previous plugin version where there was a field gost_id. 
But the developer decided to remove/rename it for whatever reasons, he updated all forms, relations, etc... all the logical things, but he forgot that `$nullable` array.
```
class City extends Model
{

    /**
     */
    public $table = 'tschallacka_geographic_cities';
    use \October\Rain\Database\Traits\Nullable;
    
    protected $nullable = ['gost_id','state_id','country_id','province_id'];
}
```

what happens is, if a form is saved now there will be an error

> SQL error unknonw field ghost_id in query insert into ....

Because the nullable trait assumes the $nullable array to be the true defining standard. So if the property doesn't exist in the model, it actually makes it, which causes it to be saved, which throws the error.

Now this isn't a real problem if a developer is working on it and he has full control of the plugin.
But what if his plugin is extended by a third party, and that third party code extends the model to add that value to the nullable array and then the database structure changes, but that third party plugin isn't updated to reflect those changes.

So this fix kinda prevents breaks in rare scenario's, but as I said, i'm not sure if this should be implemented because this forces plugin devs to to update their software and that is good. But on the other hand, rare scenario break downs and the duchtape fixes that result from them are bad.

LukeTowers:
Thank you for the explanation, that's very helpful 👍 

How expensive would it be to throw a notice or something like that to at least provide a notice that the code is invalid?

I assume that any code that currently uses that will be currently throwing SQL exceptions, so perhaps we could just make this better at detecting that situation and then throwing a more developer friendly exception than the default SQL exception? 

Personally, I'd like to avoid code that magically vacuums up bugs instead of providing warnings about buggy code.

tschallacka:
Well, this runs in the beforeSave.

To add checking code, you need to always verify with Schema if field exists. This means extra queries. = slower saving.

For production code on high volume websites that might be undesireable.

Maybe add a check if october is in production mode, use this.

If in development mode, use schema verification and throw exceptions if field does't exist in table.

```
if(!!Schema::hasColumn($this->getTable(), $field)) {
    $class = get_class($this);
    $table = $this->getTable():
    throw new ValidationException([$field=>"Column '$field' is present in \$nullable in $class but does not exist in database table $table"]);
} 
```

LukeTowers:
What about catching the specific SQL error and rethrowing it with a more helpful exception instead?

tschallacka:
Well, thing is, sql errors can be caused by legion of things. Trying to sort an unsortable column in relation managerrender, a form field updating, shennenigans of a plugin that extended the model.

Pointing it to the nullable trait is hard because the exception occurs so much later in the chain.

daftspunk:
I'm not sure I like the is_string check, seems too dirty. Shouldn't we use this instead?

```
if array_key_exists($field, $this->attributes)
```

tschallacka:
The is_string test is fast since it does only a type check. 
The attributes array will differ when someone has done a `$foo->relation()->selectRaw('count(foo) as doesntexist')`

The is_string is a bit dirty, but on the other hand you're using empty() which is a string method. Thats why i opted for the is_string method

daftspunk:
The best way to handle this would be from the other side -- to omit `ghost_id` from nullable entirely. Then *add* it dynamically if the logic demands it, using the `addNullable` method.

tschallacka:
So only if its set in the attributes array,
if it is and is empty set it to null.
If not present obey "schema" rules, e.g ->nullable() in the migration file.

Is that a good flow?

Outlook voor Android<https://aka.ms/ghei36> downloaden


________________________________

PR #252, opened 2017-01-28
Update GeneratorCommand.php

gpasztor87:
Do not throws an exception, if the file is exists.


gpasztor87:
This is also prevent overwriting existing files, but allows generating other (missing) files in the command.

LukeTowers:
But do we want the generator to continue generating files when it will be generating an incomplete set? @daftspunk any thoughts?

daftspunk:
I think it generates all the files up until that point. I think there is a `--force` flag you can use to overcome this.

LukeTowers:
@daftspunk Sounds like a good idea. @gpasztor87 Would that solve your problem?

gpasztor87:
Yes, thank you.

PR #253, opened 2017-02-09
Database memory cache

acasar:
Following discussion in https://github.com/octobercms/october/issues/2640 I went ahead and implemented a simple memory caching system to prevent duplicate queries.

The results are pretty good as you can see below:

Before:
image

After:
image

The implementation is quite simple - the results of every database query are saved in an array. And when performing insert/delete/update/truncate operations, the cache for the target table is cleared.

If needed, the cache can also be cleared manually by using:

```php
October\Rain\Database\MemoryCache::instance()->flush();
October\Rain\Database\MemoryCache::instance()->forget('backend_users');
```

Or during query construction:

```php
User::flushMemoryCache()->get(); // flush the cache
User::clearMemoryCache('backend_users')->get(); // forget the cache for the given table 
```

EDIT: updated the api

### DIFF
--- a/src/Database/MemoryCache.php
+++ b/src/Database/MemoryCache.php
@@ -1,0 +1,129 @@
+<?php namespace October\Rain\Database;
+
+use October\Rain\Support\Traits\Singleton;
+
+/**
+ * Query memory cache class.
+ *
+ * Stores query results in memory to avoid running duplicate queries
+ *
+ * @package october\database
+ * @author Alexey Bobkov, Samuel Georges
+ */
+class MemoryCache
+{
+    use Singleton;
+
+    /**
+     * Cached results.
+     *
+     * @var array
+     */
+    protected $cache = [];
+
+    /**
+     * The mapper between hashed keys and table names.
+     *
+     * @var array
+     */
+    protected $tableMap = [];
+
+    /**
+     * Check if the memory cache is enabled.
+     *
+     * @return bool
+     */
+    public function enabled()
+    {
+        return true; // TODO: Add a config option
+    }
+
+    /**
+     * Check if the given query is cached.
+     *
+     * @param  QueryBuilder  $query
+     * @return bool
+     */
+    public function has(QueryBuilder $query)
+    {
+        return isset($this->cache[$this->hash($query)]);
+    }
+
+    /**
+     * Get the cached results for the given query.
+     *
+     * @param  QueryBuilder  $query
+     * @return array|null
+     */
+    public function get(QueryBuilder $query)
+    {
+        if($this->has($query)) {
+            return $this->cache[$this->hash($query)];
+        }
+
+        return null;
+    }
+
+    /**
+     * Store the results for the given query.
+     *
+     * @param  QueryBuilder  $query
+     * @param  array  $results
+     * @return array
+     */
+    public function put(QueryBuilder $query, array $results)
+    {
+        $hash = $this->hash($query);
+
+        $this->cache[$hash] = $results;
+        $this->tableMap[$query->from][] = $hash;
+
+        return $results;
+    }
+
+    /**
+     * Delete the cache for the given table.
+     *
+     * @param $table
+     */
+    public function forget($table)
+    {
+        if(!isset($this->tableMap[$table])) {
+            return;
+        }
+
+        foreach ($this->tableMap[$table] as $hash) {
+            unset($this->cache[$hash]);
+        }
+
+        unset($this->tableMap[$table]);
+    }
+
+    /**
+     * Clear the memory cache.
+     */
+    public function flush()
+    {
+        $this->cache = [];
+        $this->tableMap = [];
+    }
+
+    /**
+     * Calculate a hash key for the given query.
+     *
+     * @param  QueryBuilder  $query
+     * @return string
+     */
+    protected function hash(QueryBuilder $query)
+    {
+        // First we will cast all bindings to string, so we can ensure the same
+        // hash format regardless of the binding type provided by the user.
+        $bindings = array_map(function($binding) {
+            return (string) $binding;
+        }, $query->getBindings());
+
+        $name = $query->getConnection()->getName();
+
+        return md5($name . $query->toSql() . serialize($bindings));
+    }
+}
--- a/src/Database/QueryBuilder.php
+++ b/src/Database/QueryBuilder.php
@@ -217,7 +217,7 @@ class QueryBuilder extends QueryBuilderBase
      */
     public function update(array $values)
     {
-        $this->clearMemoryCache($this->from);
+        $this->clearMemoryCache();
 
         return parent::update($values);
     }
@@ -230,7 +230,7 @@ class QueryBuilder extends QueryBuilderBase
      */
     public function delete($id = null)
     {
-        $this->clearMemoryCache($this->from);
+        $this->clearMemoryCache();
 
         return parent::delete($id);
     }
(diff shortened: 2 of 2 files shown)


src/Database/MemoryCache.php
```diff
@@ -0,0 +1,129 @@
...
+    protected $tableMap = [];
+
+    /**
+     * Check if the memory cache is enabled.
+     *
+     * @return bool
+     */
+    public function enabled()
+    {
+        return true; // TODO: Add a config option
```
acasar: I left this config option open, since I'm not sure where is the best place to put it.

LukeTowers:
@acasar This is fantastic, thank you very much for your work on it!

@daftspunk Will we retain the default caching behaviour as well? I feel like if we do, we should at least try to hit the memoryCache first before looking to the default cache. https://github.com/octobercms/library/pull/253/files#diff-c2d61720b8ad88e24d003784b704f697R72

src/Database/MemoryCache.php
LukeTowers: This is fine for now, @daftspunk will review it later and evaluate if we even need an option to disable it.

src/Database/QueryBuilder.php
```diff
@@ -65,14 +65,18 @@ public function cacheTags($cacheTags)
...
-     {
+     * {@inheritDoc}
+     */
+    public function get($columns = ['*'])
+    {
         if (!is_null($this->cacheMinutes)) {
             return $this->getCached($columns);
         }
 
+        if (MemoryCache::instance()->enabled()) {
```
LukeTowers: Please move the MemoryCache check before the cacheMinutes check so that queries hit the memory cache before hitting the default cache. Once that's done we can merge this. Thanks for your work!

acasar:
@LukeTowers I moved the MemoryCache check before the cacheMinutes check and I also updated the API to make the method names consistent with the Laravel's cache.

src/Database/QueryBuilder.php
```diff
@@ -181,4 +208,72 @@ public function count($columns = '*')
...
+     *
+     * @param  string|null  $table
+     * @return \Illuminate\Database\Query\Builder|static
+     */
+    public function clearMemoryCache($table = null)
+    {
+        $cache = MemoryCache::instance();
+
+        if ($cache->enabled()) {
+            $table ? $cache->forget($table) : $cache->flush();
```
LukeTowers: Do you mind explaining the rational behind this? I'm not sure that clearing the entire cache from a call to the query builder is a good idea, I preferred the other method of defaulting to the current table. If you want to have a method to clear the entire cache, then perhaps it should be separate from this one or evaluated as a parameter passed here.

LukeTowers:
@acasar Made another comment. As a side note, please don't squash your commits within a PR, it makes it difficult to parse the history of the commit. If the commits get out of hand and need to be squashed, we'll take care of it before merging.

src/Database/QueryBuilder.php
acasar: I was thinking about use cases where you'd need to call this method. Clearing the current table's cache is imo never needed since the affecting statements always clear it after themselves. So I think there are only two use cases:

1. When you're using a join or a subquery on a table that's been recently updated:

```php
User::join('group', 'foo', '=', 'bar')->clearMemoryCache('group')->get();
```

2. When the query is so complex (multiple joins, subqueries etc...) that it makes more sense to delete the entire cache altogether.

But maybe having a separate method for that is really a better idea?

LukeTowers: I think a separate method for a full cache clear would make more sense, if only for the reason that as a developer I wouldn't expect a call to `clearMemoryCache()` on my localized query to affect the entire memory cache

acasar:
@LukeTowers Fixed the builder helpers and didn't squash this time :)

LukeTowers:
@acasar Perfect, thanks very much! I'll merge it now, @daftspunk is going to review it later on but it should be good to go now :)

daftspunk:
Nice job lads

daftspunk:
All done! Please check the work here d08d39e5f6b75e779a2a396ce967bbf76e69c766 and here https://github.com/octobercms/october/commit/d19f0a1229f131f6b989df721176b7e4054644a8

LukeTowers:
Looks great, thanks!

daftspunk:
Good thing we made a config item for it... those test failures: https://travis-ci.org/octobercms/october/jobs/201630731

Fixed in https://github.com/octobercms/october/commit/d0d45f839cb1f7072c99fbe49b6c0d83790b4c76

daftspunk:
Going to rethink this.

The negative impact of memory cache is a 2x footprint for large datasets, loosely if we need to load 1GB of data in to RAM, it would require 2GB of RAM using a Db query directly. In contrast, if we were using a Model query, returning model objects, we would require a significant amount of RAM (much more than the cache demands) to store Model objects for 1GB of data. *5GB perhaps? This of course is all arm chair speculation.*

The configuration option `cms.enableDuplicateQueryCache` doesn't really make sense because it creates a logic fork. The programmer should manage the caching as their responsibility, as opposed to the application configuration. It is important that an application behaves consistently for support and reliability.

Thus:

- This config option should be removed.
- The caching should only apply to queries derived from Models, by default.
- Db queries can still support caching, not by default.
- Db queries should still bust the cache of Model queries.
- Models should declare if they don't want to disable caching.

Prototype code for model disabling caching:

```
MyModel extends Model {
    public $duplicateQueryCache = false;
}
```

As a side note: this should make the platform less magical and easier to troubleshoot, should there be any issues that arise from the memory cache. For example, if I am having trouble with `Model::where('foo', 'bar')->get();` being cached, it will be more obvious something is wrong when `Db::table(...)->where('foo', 'bar')->get();` produces different results. Essentially it acts as a hallmark to know the issue relates to the memory cache.

acasar:
@daftspunk After spending some time on it I also agree that caching should be responsibility of the programmer, otherwise we can create issues similar to "magic_quotes".

To be on the safe side we should maybe even consider an opt-in approach for models, rather than opt-out?

alxy:
May I ask why this is considered to be a public property? In my eyes, this is an internal behavior of that specific model (weather it caches or not) and should not be modified from outside of the Model definition. I recommend to make it protected.

LukeTowers:
@acasar I think I would prefer that it is opt-out. I'm personally not convinced that it is as potentially bad for performance as is proposed, as the results need to be loaded anyways and the assumption is that garbage collection is doing it's job properly. 

I feel like it's necessary to actually run performance tests before we can come up with anything concrete for this.

daftspunk:
The reason we need this is to overcome a flaw in October's component implementation. If I have two components that share the same model, they can't share it across the same request easily. This should improve performance in these instances.

```
class Component1
{
    public function user()
    {
        return User::first();
    }
}

class Component2
{
    public function user()
    {
        return User::first();
    }
}
```

It is pretty safe as the cache is busted in almost every "write" scenario. In other cases, such as `Db::statement`, there is the `$model->reload()` function that will bust the cache and reload the model contents. Our ORM is already pretty magical in a way that makes common sense, likewise it makes sense for it to operate here by default, but not at the lower query level where the expectations are more explicit.

The property is public so it can be disabled by extension. For presidence see that `$timestamps` is public.

So far tests are good, since relation queries don't seem to be affected.

daftspunk:
To clarify, the caching exists at the lower query (Db) level but is dormant by default, however it stills busts cache for write functions always. The cache is woken up by a Model preparing a [high level] query, if the `$duplicateCache` property is true (defaults: true). It can also be woken up by `Db::table()->enableDuplicateCache()` if a low level query wants to use it.

PR #254, opened 2017-02-23
Set show checkboxes to true on create:controller

neilcarpenter1:


neilcarpenter1:
I think that because we have an onDelete method available to us, it would be ideal to display a delete button and show checkboxes on backend lists by default when using php artisan create:controller

LukeTowers:
Why'd you close it?

neilcarpenter1:
Because i merged in the develop branch - would that still be ok?

LukeTowers:
Ah, gotcha. Yeah, don't add merge commits if you can help it, but if not I'll just squash the PR when I merge it into core. So not a big problem

neilcarpenter1:
@LukeTowers would I be best off creating a new pull request without the merged commits?

LukeTowers:
Yep

PR #255, opened 2017-02-23
Set checkboxes true create controller and add delete button onto backend list

neilcarpenter1:
Recreated pull request to because I messed up first time round


LukeTowers:
Reference: #254

LukeTowers:
I'm not opposed to this change, @daftspunk any thoughts?

edhardie:
+1 for this

PR #256, opened 2017-02-24
Updated thumb filename to avoid adblockers

alexdevmotion:
Using a filename like "thumb_101_180x300_0_0_crop.jpg" will cause some ad blockers to block the image download (see here also http://stackoverflow.com/questions/17255493/how-to-stop-adblock-plus-blocking-images-in-html-page).
Simply removing the "x" solved the problem in my case (maintains uniqueness and prevents blocking).

blocked_image_downloads

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -400,7 +400,7 @@ class File extends Model
      */
     protected function getThumbFilename($width, $height, $options)
     {
-        return 'thumb_' . $this->id . '_' . $width . 'x' . $height . '_' . $options['offset'][0] . '_' . $options['offset'][1] . '_' . $options['mode'] . '.' . $options['extension'];
+        return 'thumb_' . $this->id . '_' . $width . $height . '_' . $options['offset'][0] . '_' . $options['offset'][1] . '_' . $options['mode'] . '.' . $options['extension'];
     }
 
     /**


LukeTowers:
@daftspunk Thoughts?

daftspunk:
We could probably replace the `x` with an underscore. However I think this is caused by the resolution itself 180x300 (being a standard banner size), if changed to 181x301 it will probably fix it too.

alexdevmotion:
Both using underscore and changing the size solved the issue as well, however, I'm not entirely sure it will for all adblockers.

LukeTowers:
I'm not entirely sure we want to try to support all adblockers, that seems like an unnecessary headache. Personally, I'd like to retain the character separator in the file name for dimension sizes, otherwise having those numbers there at all in the first place is useless. You could have a 100x220 thumb that would have the same filename as a 1002x20 thumb.

alexdevmotion:
That's right, concatenating isn't the best solution, but changing it to underscore would be harmless and also prevent adblockers detecting it as an ad.

LukeTowers:
If the separator was retained (regardless of what it is), I don't foresee any issues, unless the thumb generating function has an assumption on the x separator characters. Any further thoughts @daftspunk?

LukeTowers:
@daftspunk bump

PR #258, opened 2017-03-02
Made filename appear in thumb files.

tschallacka:
Files that are generated using the getThumb command will now have the original uploaded filename present in it's filename.
This makes it possible to have seo friendly images in ones pages.
I only added the custom name feature to the getThumb option since a webdeveloper should already make images to the right size instead of one size fits all because of the SEO punishments google offers for too large filesizes.

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -394,13 +394,27 @@ class File extends Model
         return $thumbPublic;
     }
 
+    /*
+     * Generates a slug based upon the uploaded filename
+     * @return string
+     */
+    protected function getFilenameSlug() 
+    {
+        $extension = '.' . $this->getExtension();
+        $baseName = basename($this->file_name, $extension);
+        
+        return str_slug($baseName) . '-';    
+    }
+    
     /**
      * Generates a thumbnail filename.
      * @return string
      */
     protected function getThumbFilename($width, $height, $options)
     {
-        return 'thumb_' . $this->id . '_' . $width . 'x' . $height . '_' . $options['offset'][0] . '_' . $options['offset'][1] . '_' . $options['mode'] . '.' . $options['extension'];
+        $name = $this->getFilenameSlug();
+        
+        return 'thumb_' . $name . $this->id . '_' . $width . 'x' . $height . '_' . $options['offset'][0] . '_' . $options['offset'][1] . '_' . $options['mode'] . '.' . $options['extension'];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
This creates a mixture of both, producing a pretty long file name. I was thinking more along the lines of one or the other, enabled via config. Either hashed filenames, or nice filenames with a method for collision detection. Then perhaps allowing the user to modify the filename in the File Upload widget.

Consider this on the roadmap.

tschallacka:
How about storing the thumb files in a folder that is named by the hash?

So it would be /542/542/4ad/5425424adeaeaea421/some_filename_420x300.jpg?

That way cleaning up after an image is easier too when its deleted since only the hash folder has to be deleted. Without the "pattern" checking that currently occurs.

I could modify it to that.

If you really wish to prevent collisions I'd go with an UUIDv4 instead of a hash. that way syncing between systems isn't an issue either when people use a mixture of servers and systems. (i use it because my website runs on two servers geographically seperated that need to sync stuff.)

I even have a plugin to add UUID support + database driver extensions for october that I'm planning to ship soon, i just need to add more documentation and psr/octoberspec proof it and write unit tests.

If you're interested i could put a rush on it so you can look at it and see if it's a thing to add to October.


Outlook voor Android<https://aka.ms/ghei36> downloaden


________________________________

tschallacka:
Hmmmh, a better solution would be to use htaccess or web.config. I right now implemented this like this on my IIS server.

    <rewrite>
                <rules>
				<rule name="RewriteToBasePng" stopProcessing="true">
					<match url="(.*)/uploads/public/(.*)/(.*)/(.*)/(.*)\/(.*)_(.*)w_(.*)q.png(.*)" />
					<conditions logicalGrouping="MatchAll" trackAllCaptures="false" />
					<action type="Rewrite" url="{R:1}/uploads/public/{R:2}/{R:3}/{R:4}/{R:5}/{R:7}w_{R:8}q.png{R:9}"/>
				</rule>
				<rule name="RewriteToBaseJPG" stopProcessing="true">
					<match url="(.*)/uploads/public/(.*)/(.*)/(.*)/(.*)\/(.*)_(.*)w_(.*)q.jpg(.*)" />
					<conditions logicalGrouping="MatchAll" trackAllCaptures="false" />
					<action type="Rewrite" url="{R:1}/uploads/public/{R:2}/{R:3}/{R:4}/{R:5}/{R:7}w_{R:8}q.jpg{R:9}"/>
				</rule>

PR #263, opened 2017-05-16
Add fromData() method

alxy:
This is the easiest implementation possible, however it requires a temporary file to be created. It can be discussed, if there is a need to delete the file afterwards explicitly.

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -138,8 +138,8 @@ class File extends Model
         FileHelper::put($tempPath, $data);
 
         $file = $this->fromFile($tempPath);
-        unset($tempPath);
-        
+        FileHelper::delete($tempPath);
+
         return $file;
     }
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
If we go the temp file route, then we'll need to unlink it afterwards to remove it immediately afterwards. Something like

```php
$path = //create temp file;
$file = $this->fromPath($path);
unlink($path);
return $file;
```

Have you looked into using the PHP IO streams instead for this? http://php.net/manual/en/wrappers.php.php

alxy:
I added a commit to delete the temporary file.

LukeTowers:
That doesn't delete it, that just removes the reference to it in PHP.

alxy:
And this is why I should never commit untested code.  Yeah, I will test this first and then provide the correcting PR.

LukeTowers:
Also, could you please look into using the PHP streams feature I linked? If we could use that instead we wouldn't even need to create a temporary file

PR #267, opened 2017-06-14
Add Interlacing

Teranode:
This adds interlacing to png/gif/jpeg (webp excluded cause idk if it uses it)

### DIFF
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -412,10 +412,8 @@ class Resizer
 
         $imageQuality = $this->getOption('quality');
         
-        $interlace = $this->getOption('interlace');
-        
-        if($interlace) {
-            imageinterlace($image, $interlace);
+        if($this->getOption('interlace')) {
+            imageinterlace($image, true);
         }
 
         // Determine the image type from the destination file
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -125,6 +125,7 @@ class Resizer
      *  - mode: Either exact, portrait, landscape, auto or crop.
      *  - offset: The offset of the crop = [ left, top ]
      *  - sharpen: Sharpen image, from 0 - 100 (default: 0)
+     *  - interlace: Interlace image, 0 (false - default) 1 (true)
      *  - quality: Image quality, from 0 - 100 (default: 95)
      * @return self
      */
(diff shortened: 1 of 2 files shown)


LukeTowers:
Could you open a PR to the docs repo documenting this new option please?

Teranode:
Here you go:
https://github.com/octobercms/docs/pull/244

daftspunk:
Related issues:

https://github.com/octobercms/library/pull/267
https://github.com/octobercms/docs/pull/244
https://github.com/octobercms/docs/pull/245
https://github.com/octobercms/october/issues/2795

daftspunk:
Any reason why this shouldn't be enabled by default? Seems like all upside with no downside, right?

src/Database/Attach/Resizer.php
```diff
@@ -125,16 +125,18 @@ public function reset()
...
     {
         $this->options = array_merge([
-            'mode'    => 'auto',
-            'offset'  => [0, 0],
-            'sharpen' => 0,
-            'quality' => 90
+            'mode'      => 'auto',
+            'offset'    => [0, 0],
+            'sharpen'   => 0,
+            'interlace' => 0,
```
daftspunk: If this is a boolean it should probably be defined as such.

LukeTowers:
@daftspunk @Teranode  It only really makes sense to enable it by default on JPEGs, and only if `imageinterlace()` actually creates a progressive JPEG instead of an interlaced one. See https://stackoverflow.com/questions/13449314/when-to-interlace-an-image for more information

daftspunk:
Let's factor this in to the code then, interlace default true for JPG, false for all others.

Teranode:
I usually have it disabled by default because the images are slightly larger that non-progressives

daftspunk:
According to the answer provided it makes JPG files smaller...

Teranode:
Im not sure if that answer is valid, the output differs case-by-case

daftspunk:
Fair enough. Why are we calling `imageinterlace()` at all if the value is set to false?

LukeTowers:
Also we shouldn't be treating image types differently if there is only one option for all three

LukeTowers:
@Teranode don't forget to update the docs please :)

Teranode:
@LukeTowers i was factoring in the fact that I had submit a PR for WEBP at the time, i wasn't sure if it used interlacing so I had the options separated.

Teranode:
@LukeTowers @daftspunk would that work?

LukeTowers:
WEBP doesn't use interlacing AFAIK. If you aren't using `$interface` more than once, then you should just make the call to `getOption()` within the conditional.

daftspunk:
Sorry to be a pain, but please add a space in between if and the bracket.

```
if(
```

Should be

```
if (
```

PR #268, opened 2017-08-11
Force schema while using 'force' link policy

adsa95:
This is a fix for octobercms/october#3030.

The `forceRootUrl()` method of `Illuminate\Routing\UrlGenerator` sounds like it should include schema as well, but apparently don't (see my comment in the above issue). Not sure if this is the correct way to solve this or if changes should actually be made in `Illuminate\Routing\UrlGenerator`. However I think this is a suitable fix.


daftspunk:
Thanks @adsa95!

PR #269, opened 2017-08-14
L5.5 MorphToMany Compatibility

witul:
MorphPivot constructor extending Eloquent\Model now, so there is following error:
Type error: Argument 1 passed to Illuminate\Database\Eloquent\Model::__construct() must be of the type array, object given, called in .../vendor/october/rain/src/Database/Relations/MorphToMany.php on line 164

So i just looked for a new Laravel implementation and made the code exactly the same.


daftspunk:
Thanks @witul!

PR #272, opened 2017-08-30
Fix sending mail to Collection

adsa95:
As per the documentation the first argument in `Mail::sendTo()` should be able to be of type `String`, `Array`, `Object` or `Collection`. However as a `Collection` is a type of `Object` it would get caught in the `is_object()` statement when processing recipients, which would lead to "no recipient" error. This is the current structure in the `processRecipients($recipients)` function:
````php
if (is_string($recipients)) {
    ...
}
elseif (is_object($recipients)) {
    ...
}
elseif (is_array($recipients) || $recipients instanceof \ArrayAccess) {
    ...
}
````
This PR simply adds an additional check to make sure that Collections are passed on to the next check and each item in the Collection is added as an recipient as expected.
````php
elseif (is_object($recipients) && !($recipients instanceof Collection) {
    ...
}
````
I used `Illuminate\Support\Collection` for the check, because it seemed to be the "root" Collection class. Simply changing the order of the if-statements did not work as it seems the `Model` class is an instance of `ArrayAccess`, and I think it's important to be able to pass along a Model-object with an email instead of simply a raw object.

### CODE REVIEW

**Comment by adsa95** [comment]
I also added a convenience `$bcc` parameter to `sendTo()` and `rawTo()` that adds all recipients as Bcc instead of To addresses, useful for easily sending newsletters etc when you don't want all e-mails to show for every recipient.

**Comment by LukeTowers** [comment]
Please provide an example of the code you are using right now that is broken before your fix and works after your fix. Also, when adding new features, please make a related PR to the octobercms/docs repository documenting them. Thanks for your work on this!

**Comment by adsa95** [comment]
Example code:
````php
$recipients = Subscriber::all();// where Subscriber has 'email' property, and one or more exist
Mail::rawTo($recipients, 'This is a test message');
````
Error:
````
dev.ERROR: Swift_TransportException: Cannot send message without a recipient in (Project path)/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/MailTransport.php:133
````
Since `Subscriber::all()` returns a `Collection`, which is an object, it gets caught in the second if-statement if you look at the structure in my initial post. However that if-statement is meant to process single recipients where the object has an email property, which the `Collection` don't and as such it is ignored and no recipients are added, resulting in the error above.

After my PR `Collection`s will not get caught in the if-statement meant for single recipients, but instead continue down to the if-statement meant to catch arrays etc, leading to all recipients in the `Collection` getting added as expected, eliminating the error.

**Comment by LukeTowers** [comment]
Awesome, thanks a bunch for your work on this @adsa95!

**Comment by daftspunk** [comment]
Some further changes made to this here: https://github.com/octobercms/library/commit/64887a784290cf58f1f5c3020e4b9240efdfe650


Changed the expression order to this instead:
```
if (is_string($recipients)) {
    ...
}
elseif (is_array($recipients) || $recipients instanceof \ArrayAccess) {
    ...
}
elseif (is_object($recipients)) {
    ...
}
```

adsa95:
@daftspunk I think this rearrangement has unwanted consequences. As stated in the bottom of my first post, doing it this way makes single `Model`s getting catched in the array statement, because the model class implements `ArrayAccess`:

````
Illuminate/Database/Eloquent/Model.php

abstract class Model implements ArrayAccess, ...
````
This means that the following would not work, which would be very inconvenient:
````php
$subscriber = Subscriber::first(); // where subscriber has 'email' property
Mail::sendTo($subscriber, ...); // 
````
Instead it gives the following error:
````
dev.ERROR: Swift_RfcComplianceException: Address in mailbox given [...] does not comply with RFC 2822, 3.6.2. in (Project Path)/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php:345
````
Although this might seem like it's because of an ill-formatted email, it's actually because the process function handles it as an array and reads the wrong value(s) as emails.

daftspunk:
Thanks for this @adsa95, I will revert it back today.

daftspunk:
Check the latest commit (ab2f08a91db5bf71c90650bed5b476a7ab1a8547), does it work for you? 

I think the root issue is that `ArrayAccess` is too generic for this purpose, the original intention was to target a `Collection` so it has been changed to reflect this.

adsa95:
That seems to work great, and sounds like the best approach!

PR #273, opened 2017-09-05
Add default config_list.stub `showPageNumbers` config

jofrysutanto:


LukeTowers:
Refs: https://github.com/octobercms/october/pull/3084

PR #274, opened 2017-09-17
Method forceSchema does not exist

lasat:
Fixed Typo

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -499,7 +499,7 @@ trait HasRelationships
 
         $localKey = $localKey ?: $this->getKeyName();
 
-        $secondLocalKey = $secondLocalKey ?: $through->getKeyName();
+        $secondLocalKey = $secondLocalKey ?: $throughInstance->getKeyName();
 
         $instance = $this->newRelatedInstance($related);
 


NemoD503:
+1 to this PR.I have trouble with tests - all test fails with error `BadMethodCallException: Method force Schema does not exist.` when providers register.

NemoD503:
I think, it`s not good, that provider registration depends on config and tests use only **one** variant  of config value (UrlGeneratorPolicy in this case) so test pass only with default config set, but really project fails on start, if user choose "force" UrlGeneratorPolicy.

LukeTowers:

LukeTowers:
@NemoD503 does this PR fix the issue for you? @lasat could you please provide a more detailed description of the issue that this PR fixes?

NemoD503:
@LukeTowers yes, this pr fixed my issue, for temporary fix i had have to use @lasat repo, and now all test pass.
It`s really a typo, because you can see correct name of the same method in this file below, [here](https://github.com/octobercms/library/blob/master/src/Html/UrlServiceProvider.php#L43) and [here](https://github.com/octobercms/library/blob/master/src/Html/UrlServiceProvider.php#L47). So it`s just a typo.

NemoD503:
@LukeTowers I've made pr with [demonstration](https://github.com/NemoD503/october/pull/1). Just replace policy to force and all test will fail.

LukeTowers:
@NemoD503 thanks! @lasat great job on spotting this!

PR #276, opened 2017-10-02
Fix pagination total count

alxy:
This ensures that all scopes are applied to the query and is in compliance with the L5.5 paginate() method.
Fixes https://github.com/octobercms/october/issues/3138 .

### DIFF
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -123,7 +123,7 @@ class Builder extends BuilderModel
             $perPage = $this->model->getPerPage();
         }
 
-        $total = $this->query->getCountForPagination();
+        $total = $this->toBase()->getCountForPagination();
         $this->query->forPage((int) $currentPage, (int) $perPage);
 
         return new LengthAwarePaginator($this->get($columns), $total, $perPage, $currentPage, [


LukeTowers:
Could you explain in a bit more detail how exactly this works? Wouldn't `toBase()` just return a blank copy of the query object? And why is the regular query still run on the `->query` property afterwards? Does `toBase()` modify the underlying query instance?

alxy:
First of all, please note that this mimics the default Laravel behavior: https://github.com/laravel/framework/blob/5.5/src/Illuminate/Database/Eloquent/Builder.php#L708 (I'm not really sure why it is necessary to overwrite this method anyway).

The [toBase()](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Database/Eloquent/Builder.php#L1167-L1175) just returns a base ``Builder`` with scopes like ``SoftDelete`` applied to it. This will ensure all constraints given by the scopes are actually applied to the query. Please note, that ``getQuery()`` also actually returns ``$this->query``.

daftspunk:
LGTM. Looks like a missed migration from L5.1 -> L5.5

PR #277, opened 2017-10-04
Fixes https://github.com/octobercms/october/issues/2845

alxy:
This is an additional fix to https://github.com/octobercms/october/issues/2845 for ``BelongsTo`` proxy fields.
For more details, refer to the issue comments.

### DIFF
--- a/src/Database/Relations/BelongsTo.php
+++ b/src/Database/Relations/BelongsTo.php
@@ -68,6 +68,7 @@ class BelongsTo extends BelongsToBase
             if (!$value->exists) {
                 $value->bindEventOnce('model.afterSave', function() use ($value){
                     $this->associate($value);
+                    $this->child->save(['timestamps' => false]);
                 });
             }
 


daftspunk:
Discussion continued at https://github.com/octobercms/october/issues/2845

daftspunk:
This is fixed using alternative method. See related commits (above).

The reason the proposed code wasn't implemented like this is because of the potential for circular logic. So as a rule the parent to child relationship is a one way function (it cannot flow in the opposite direction).

#### Post belongsTo Status (Parent -> Child)

```
$post = new Post;
$status = new Status;

$post->status = $status;

$status->save(); // <-- Child is never allowed to save parent, only associate
$post->save();
```

#### Post hasOne Gallery (Parent -> Child)

```
$post = new Post;
$gallery = new Gallery;

$post->gallery = $gallery;

$gallery->save();
$post->save(); // <-- Parent is permitted to re-save child for association
```

PR #281, opened 2017-10-20
Added support for datepicker field

vanmil:
This change adds support for the Datepicker field type in the syntax parser. The available datepicker options (mode, format, minDate, maxDate, firstDay) discussed [here](https://octobercms.com/docs/backend/forms#widget-datepicker) are working correctly.

### DIFF
--- a/src/Parse/Syntax/Parser.php
+++ b/src/Parse/Syntax/Parser.php
@@ -219,10 +219,17 @@ class Parser
                 $result = '{% if ' . $field . ' %}' . $params['_content'] . '{% endif %}';
                 break;
             case 'datepicker':
-                if ($params['mode'] === 'date'){
-                    $result = '{{ ' . $field . '|date("Y-m-d") }}';
-                } elseif ($params['mode'] === 'time') {
-                    $result = '{{ ' . $field . '|date("H:m:s") }}';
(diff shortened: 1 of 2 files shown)


src/Parse/Syntax/Parser.php
```diff
@@ -218,6 +218,13 @@ protected function evalTwigViewField($field, $params, $prefix = null)
             case 'checkbox':
                 $result = '{% if ' . $field . ' %}' . $params['_content'] . '{% endif %}';
                 break;
+            case 'datepicker':
+                if ($params['mode'] === 'date'){
```
LukeTowers: This doesn't actually properly support the `mode` option. The `mode` option supports three values, `date`, `time`, and `datetime`. Up to you if that makes more sense with a bunch of elseifs or a select case setup.

LukeTowers:

vanmil:
I changed it from an if construction to switch case construction and also added the datetime case.

For the `ignoreTimezone` option, it doesn't not work yet, because I found a more generic bug. The parser does not parses boolean strings to real booleans. I made a separate pull request for this, because it is a more generic thing which also has implications for other field types besides the datepicker, see #283  for that request. When that request is added, the `ignoreTimezone` option does work.

PR #282, opened 2017-10-21
Add media_path() helper

datune:
I'll update the documentation if this is accepted.

### DIFF
--- a/src/Support/helpers.php
+++ b/src/Support/helpers.php
@@ -192,6 +192,20 @@ if (!function_exists('uploads_path'))
     }
 }
 
+if (!function_exists('media_path'))
+{
+    /**
+     * Get the relative path to the media folder.
+     *
+     * @param  string  $path
+     * @return string
+     */
+    function media_path($path = '')
+    {
+        return Config::get('cms.storage.media.path').($path ? '/'.$path : $path);
+    }
+}
+
 if (!function_exists('themes_path'))
 {
     /**
--- a/src/Support/helpers.php
+++ b/src/Support/helpers.php
@@ -202,7 +202,7 @@ if (!function_exists('media_path'))
      */
     function media_path($path = '')
     {
-        return Config::get('cms.storage.media.path').($path ? '/'.$path : $path);
+        return app('config')->get('cms.storage.media.path').($path ? '/'.$path : $path);
     }
 }
 
--- a/src/Support/helpers.php
+++ b/src/Support/helpers.php
@@ -195,14 +195,22 @@ if (!function_exists('uploads_path'))
 if (!function_exists('media_path'))
 {
     /**
-     * Get the relative path to the media folder.
+     * Get the relative path to the media folder if running within OctoberCMS.
      *
      * @param  string  $path
      * @return string
      */
     function media_path($path = '')
     {
-        return app('config')->get('cms.storage.media.path').($path ? '/'.$path : $path);
+        /**
+         * In theory, someone can use the October Library without using OctoberCMS fully by
+         * itself, which is why we can't directly reference things like cms.media.path
+         * */
+        if (app('config')->hasGroup('cms')) {
+            return app('config')->get('cms.storage.media.path').($path ? '/'.$path : $path);
+        }
+
+        return $path;
     }
 }
 


LukeTowers:
The concept is accepted, however I can't accept this implementation. See how the other path helpers are implemented? You'll need to implement that: https://github.com/octobercms/library/blob/master/src/Support/helpers.php#L205

In theory, someone can use the October Library without using OctoberCMS fully by itself, which is why we can't directly reference things like `cms.media.path` which would only be present in the full OctoberCMS instance.

Also, 

datune:
1. Sorry for the wrong branch, it gets confusing which repo accepts PR's to the master branch and which not...(docs, rainlab/builder, etc.. does accept them, octobercms does not, apparently library also not...)

2. I changed the implementation as per your request.

LukeTowers:
I almost feel like this would be better suited to being contained within the system module itself instead of the core library.

datune:
Well, I guess we could add it as a static method to [Cms.php](https://github.com/octobercms/october/blob/master/modules/cms/helpers/Cms.php)?

LukeTowers:
If all you want is a PHP method to accomplish the same affect as the `|media` Twig filter, that already exists. `\System\Classes\MediaLibrary::url($relativePath)` will get you the full URL to the image. If you would like to get the full path to the relativePath provided, then maybe we should look at adding a `::path($relativePath)` helper.

datune:
@LukeTowers If I wanted to accomplish that, I might as well just call Config::get('cms.storage.media.path'). The reason is to be consistent, there are a myriad of XXX_path() helpers available, but not a media_path(). That was my intention. to have a simple, GLOBAL helper. It's not that complicated now is it?

LukeTowers:
No need to break out the sarcasm @datune ;)
I suggested that because you suggested adding it as a static method to Cms.php. It's not quite as simple as the other path helpers because the Media Library is not quite as low level as plugins, themes, or even just the storage directory is. The media library is a piece of the System module, with the user interface in the Backend module.

datune:
@LukeTowers Aaaaw come on man, let me have a little sarcasm this early in the day (it's 07:45 over here) :-P Also, if you'll get what you give (telling me about the |media filter AND adding that is already exists, now who's being sarcastic ;-)

Well, I suggested Cms.php since it is already a Helper class, and I could not find any global helpers file for the system module (neither for the backend or cms module).  So where could we put it?

Luke, you're my only hope?

LukeTowers:
How is Cms.php a global helper? If we're going the route of static methods, then it would belong in the System\Classes\MediaLibrary class.

I'll allow the sarcasm, but only because it's my default form of communicating 😂

datune:
@LukeTowers I never said Cms.php is a global helper, I think you need some sleep man ;-) I said it is a Helper class, and because I could not find a global helper file, I suggested to use it ;-)

The point of it all is to have a short global helper function, just like all the others. If there's no appropiate place according to you, we might as well just forget about it.

TLDR:
Can use Config::get('cms.storage.media.path'), want to add media_path() to basically do the same.

LukeTowers:
The more I think about it the less it makes sense to have it available as a global helper function. If you want a better way to access the correct path other than the Config::get method I suggest adding a `::path()` method to the `System\Classes\MediaLibrary` instead. Thoughts?

datune:
I guess it depends on how you look at it. I can see where you are coming from in theory. But the MediaLibrary is not going anywhere, it seems like a pretty vital part of OctoberCMS, and sometimes you just gotta be (a) pragmatic (programmer). I can't personally see the harm in having a helper available to fetch the media path, after all, what's the difference of doing 

```php
public function decorateElement($imgName)
{
    return $this->element . '<img src="' . Config::get('cms.storage.media.path') . '/' . $imgName . '">';
}

```
all over your code versus
```php
public function decorateElement($imgName)
{
    return $this->element . '<img src="' . media_path($imgName) . '">';
}
```
Sure, we could go this route, and syntax wise I even prefer it, but this is splitting hairs. 

```php
public function decorateElement($imgName)
{
    return $this->element . '<img src="' . MediaLibrary::path($imgName) . '">';
}
```
The one thing I do agree with though, is there seems to be no appropiate place to add "global helper functions" outside of october core. So, in this case, it makes sense to keep it within the "scope" of the MediaLibrary class and add it as a static (helper) method. 

My original intention, for which, any time saving I might have ever achieved in the future you have now ruined times a hundred anyway, was to save time typing so much. I rue the day I came up with that silly idea, -)

If you're sure, I can add that method to MediaLibrary, but perhaps you'd like to give it another thought or five hundred? 

PS: I'm trying to be (in good fun) sarcastic, just in case I'm totally failing at it 🥇

PR #284, opened 2017-10-28
Allow singleton instances to be reset

jimcottrell:
Pretty straightforward. Discussion is in octobercms/october#3208.


LukeTowers:
I'll get @daftspunk to weigh in on these changes before merging, but thanks for your work so far!

daftspunk:
I like this idea a lot. Thanks!

PR #285, opened 2017-11-16
Throw exception when migration script is missing

harmenjanssen:
When you make a typo in the filename of a migration script, the script will be skipped
silently. It's not clear to the developer a script is skipped, and since
the plugin version is updated anyway, there also is no easy way to go
back a single version to re-execute the script.

This update makes sure the migration stops in its tracks and the error
won't slip through.


LukeTowers:
I believe this has been fixed by logging a warning instead.

PR #286, opened 2017-11-20
Black border on some cropped thumbnails

Fl0Cri:
## Environment
October build: 428
Php version: 7.0 64bits (tested on Windows and Linux)

## Reproduce steps
- Get an image with a non-black background (I used an uniform gray background of 1200x1600 px)
- Add this image using a `System\Models\File` relation (I used Rainlab Blog plugin)
- Create a cropped thumbnail to a smaller size (I did from twig `myImage.thumb(204, 272, 'crop')`)
- Note that in this particular case, the ratio of the original image and the thumbnail are same, but I suspect that other sizes could be problematic too.

## Actual result
The generated thumbnail get a black border of 1px width on the right side, like this:
thumb_1_204_272_0_0_crop

## Expected result
The thumbnail should be an uniform gray background without the black border, just like the source image.

## Proposed fix
By looking at the `October\Rain\Database\Attach\Resizer` class with a debugger, I found this funny situation at the end of the `resize()` method (I wrote the values coming from the debugger in red):
resizer php - october - visual studio code 2017-11-20 09 10 37

As you can see, `$cropStartX` should be 0 but get an unexpected value. Then, the `crop()` method will call the `imagecopyresampled()` function, which takes integers as parameters, so the value of `$cropStartX` will be cast to `-1` instead of 0.

The problem comes from the `$optimalWidth` variable which is not exactly `204`: in `getOptimalCrop()`, it's set to `1200 / (1200 / 204)`, so it gets a precision loss when computing the intermediate value.

My suggestion is to make a `round()` on the values returned by `getOptimalCrop()`, so width and height will both be clean integers. As all functions using these values takes integers as parameters, it should not be a big problem.
But one case that could theorically be problematic is when the other returned variable (`$optimalHeight` in my case) is between xx.5 and xx.999... : previously, it was floored, now it will be rounded to the upper integer.
In practice, I was not able to find image sizes where this has an influence.

### DIFF
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -612,8 +612,8 @@ class Resizer
             $optimalRatio = $widthRatio;
         }
 
-        $optimalHeight = $this->height / $optimalRatio;
-        $optimalWidth  = $this->width  / $optimalRatio;
+        $optimalHeight = round($this->height / $optimalRatio);
+        $optimalWidth  = round($this->width  / $optimalRatio);
 
         return [$optimalWidth, $optimalHeight];
     }


LukeTowers:

LukeTowers:
@Fl0Cri This looks good, but you said that previously it would be floored, what's the rationale behind not continuing to do that, except more explicitly?

Fl0Cri:
Oops, sory for the wrong branch.

The reason why it needs to be rounded instead of floored is because when php computes `$optimalWidth`, for example with `1200 / (1200 / 204)`, it gets something like `203.9999...` instead of `204.0`.
If we try to floor this value (or cast it to an integer like the actual code is doing), we get `203` instead of `204`.
Demo: 
invite de commandes 2017-11-21 08 43 04

### Note
Here follows a long brain dump of the ways I explored and the tests I made, sory if it is not realy well explained, feel free to ask if you need complements or better explainations on a point.

### A ugly (imo) but easy solution
A solution that I taught about is that instead of simply rounding, we could do `floor(round($optimalWidth, 7));`. So that very small errors are rounded, then the result is floored: for example 203.999... is rounded to 204 then floored to 204 and other values like 271.85 are rounded to 271.85 then floored to 271 like in the previous code.
But I don't like this approach because the precision of the round we would make (I've put 7 arbitrary for the example) should be aware of some low level details to be accurate (we are dealing with the precision of the floating point representation made by php, and this precision is not the same between 32 or 64bits platforms).

### Simplification of what the code actually does and what goes wrong
As a complement of what I reported in the first post, here's a more detailed demonstration of how the resize() method performs (without the fix) with a source image of `1200x1600`, and a target image of `204x272`:
1. Get the optimal dimensions: `203.999x272` without the fix
1. Resize source image to optimal dimensions. As `imagecopyresampled()` takes integers as values, the resized image is `203x272`
1. Cropping: a new (black) "container" image is created using the requested dimensions (not the resized ones), so its size is `204x272`
1. The resized image is copied inside the cropped container, cutting overflowing parts (nothing overflows with these specific values).
1. But because the container is `204x272`, and the image that is copied inside is `203x272`, we can see the black background of the container at the right side of the image.

### Comparison between the current PR and actual code with chosen values
Now based on the previous demonstration, we can play with it and inject values that whould cause the dimensions of the image to behave differently than with the actual method:
- Example with src: `1200x1600`, target: `204x272` (the initial problem)
    - Optimal dimensions: `203.999x272.0` (without fix), `204x272` (with fix)
    - With the fix, everything will work fine
    - Without the fix, there will be a black border, see Demo above
- Example with src: `1200x1599`, target: `204x272`
    - Optimal dimensions: `204.127x272.0` (without fix), `204x272` (with fix)
    - Same result, both correct.
- Example with src: `1200x1594`, target: `204x272`
    - Optimal dimensions: `204.768x272.0` (without fix), `205x272` (with fix)
    - ~~If crop is disabled, resized image will match the requested size using the actual method, but be 1px wider than requested using the method of this PR~~
    - ~~But note that the images ratios are - initial: `0.75`, resized-actual: `0.7576`, resized-PR: `0.7537`. So the  image with the PR is a little bit less stretched.~~ _(if crop is disabled, optimal dimensions are not provided by the same method, I've let this here because it is useful to understand the next point)_
    - If crop is enabled, the dimensions of the resulting image will match the requested ones in both case, but with the new PR the image will have 1px cropped at the right, versus a few more stretching in the actual implementation.
- Example with src: `1200x1607`, target: `204x272`
    - Optimal dimensions: `203.999x273.19` (without fix), `204x273` (with fix)
    - The returned image will have a black border at the right side and be 1px cropped at the bottom (without fix), or idealy sized in width and cropped by 1px at the bottom (with fix).

### A more robust way, but that needs a bit more rewriting
An other way I am thinking just now would be to refactor the `getOptimalCrop()` to something like this (untested):
```PHP
protected function getOptimalCrop($newWidth, $newHeight)
    {
        $heightRatio = $this->height / $newHeight;
        $widthRatio  = $this->width /  $newWidth;
        $optimalHeight = $newHeight;
        $optimalWidth = $newWidth;

        if ($heightRatio < $widthRatio) {
            $optimalWidth = $this->width / $heightRatio;
        }
        else if ($heightRatio > $widthRatio) {
            $optimalHeight = $this->height/ $widthRatio;
        }
        
        return [$optimalWidth, $optimalHeight];
    }
```
The point here is to avoid the computation of an intermediate value and only touch the necessary values.
Result of this method should be exactly the same as with actual code, except that instead of `203.999...`, it will return directly `$newWidth` = `204`.

What do you think. Do you want me to update the PR with this code or leave it as is for now?

### Finally
I am not sure about what way is the best so I let you choose. If you need more info or better explainations, feel free to ask

LukeTowers:
@Fl0Cri thanks for the detailed write up. I'm thinking that for now we can merge your changes and if anyone in the future has something to add we might end up moving to the proposed separate method.

PR #289, opened 2017-11-24
Fixes an issue with static pages plugin

SaifurRahmanMohsin:
When a static page layout is created without any variable fields it throws an error `Call to a member function getFields() on null". This fixes that issue.


bennothommo:
This does not seem to occur with the latest versions of the Static Pages plugin. Please feel free to re-open the PR if it does still occur.

PR #294, opened 2018-01-10
Generalize Database\Attach\File for storage disk

pikanji:
This prepares for the fix for octobercms/october#3346.
This fix does not change any behavior, but provides flexibility to use user-defined storage configuration other than default in config/filesystems.php.

Though the fix for octobercms/october#3346 could be done only in `System\Models\File`, I suggest to fix from `October\Rain\Database\Attach\File` because of the following reasons.

* Implicit use of default value is confusing.
* Generalizing `October\Rain\Database\Attach\File` makes the code cleaner.

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -1,9 +1,9 @@
 <?php namespace October\Rain\Database\Attach;
 
+use Config;
 use Storage;
 use File as FileHelper;
 use October\Rain\Database\Model;
-use October\Rain\Database\Attach\FileException;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\File\File as FileObj;
 use Exception;
@@ -57,6 +57,17 @@ class File extends Model
     public $data = null;
 
     /**
+     * @var string The name of the disk where uploaded files are stored.
+     * This is set based on the key returned by getStorageConfigKey().
+     */
+    protected $storageDiskKey = null;
+
+    /**
+     * @var \Illuminate\Filesystem\FilesystemAdapter An instance for storage disk obtained with $storageDiskKey.
+     */
+    protected $storageDisk = null;
+
+    /**
      * @var array Mime types
      */
     protected $autoMimeTypes = [
@@ -73,6 +84,18 @@ class File extends Model
     //
     // Constructors
     //
+
+    /**
+     * Constructor
+     */
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct();
+
+        $this->storageDiskKey = $this->getStorageConfigKey() ?
+            Config::get($this->getStorageConfigKey() . '.disk') : Storage::getDefaultDriver();
+        $this->storageDisk = Storage::disk($this->storageDiskKey);
+    }
 
     /**
      * Creates a file object from a file an uploaded file.
@@ -120,7 +143,7 @@ class File extends Model
 
         return $this;
     }
-    
+
     /**
      * Creates a file object from raw data.
      *
@@ -563,7 +586,7 @@ class File extends Model
                 FileHelper::delete($collection);
             }
             else {
-                Storage::delete($collection);
+                $this->storageDisk->delete($collection);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/Database/Attach/File.php
```diff
@@ -74,6 +85,18 @@ class File extends Model
...
     //
 
+    /**
+     * Constructor
+     */
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct();
+
+        $this->storageDiskKey = $this->getStorageConfigKey() ?
```
LukeTowers: Why does this need to be a method call instead of just a pre-initialized property? Also, why do we use `Config` here? It may make sense for your specific use case, but it makes the purpose of the key change between being a reference to a config value and being a straight forward reference to an actual storage disk. I think a better option might be to use the method `getStorageDiskKey()` instead of the property altogether and then override this method in the system\models\file to use the config value

pikanji:
@LukeTowers Thanks, that's much better! (Regarding `getStorageConfigKey()`, I chose to make it method because of its flexibility, nothing more, but it doesn't matter now.)

LukeTowers:
@pikanji Could you add some tests to compliment this improvement?

pikanji:
Of course!

pikanji:
@LukeTowers Since src/Database/Attach/File.php uses facades, I need #296 to be merged to execute the tests.

LukeTowers:
@pikanji why are those changes required to test a class that uses facades?

pikanji:
@LukeTowers It is simply because tests encounter errors regarding facades if the tests themselves or any class being used are using facades.

Currently any test is not written for Database/Attach/File.php. When I run my phpunit test for existing code, I encountered the errors below.
```
Error: Class 'Storage' not found
```
then of course I retried it with replacing `use Storage;` to `use Illuminate\Support\Facades\Storage;` then I got;
```
RuntimeException: A facade root has not been set.
```

Based on my search and debugging, I concluded that I needed a test case that booted up a driver Laravel app to initialize facades. I looked for tests for classes using facades for my reference but I couldn't find any class or better way. If you have any better solution, please let me know.

LukeTowers:
Is https://laravel.com/docs/5.5/mocking#storage-fake helpful at all or is it in regards to something completely different?

pikanji:
@LukeTowers That would be quite handy in writing tests! I'll use it. Thank you!
However, the problem here is that Storage facade is not available, and thus the invocation of fake method fails as well.
In order to use fake method, I still need to initialize facade by booting up a Laravel app with `LaravelTestCase` in #296.

LukeTowers:
Are you able to reference the service provider behind the Storage facade directly? https://github.com/laravel/framework/blob/5.5/src/Illuminate/Support/Facades/Storage.php I'm uncomfortable with merging the proposed testing changes because of how sweeping they are; I'd rather have a more specific test than basically copy an entirely new Laravel only application into the testing environment

pikanji:
I understand your concern, I don't want to add so many files just for testing. So, I spent quite some time to find other ways to minimize the change.
I'll try replacing all facade usages of the test subject class with the references to the service providers. But, this issue stands for all classes using facades, and thus we need to remove all the facade usage from octobercms/library to write tests for them. If we can do it, I think it the better way. So, I'll try.

LukeTowers:
I'd agree with you on that. I want to extend the test coverage but this is an issue we will need to properly deal with at some point. Perhaps we could get @daftspunk to weigh in on the issue.

pikanji:
@LukeTowers As you suggested, I tried with underlying service providers to avoid using facades. I found that these service providers also require an instance of Laravel application, meaning that I still need to boot up a Laravel application for testing.
But, there is a good news. I also found that laravel/framework uses orchestra/testbench for testing classes using Laravel features like facades, and we can use it for octobercms/library. It provides a TestCase that boots up a Laravel application with facades available.
I opened  another PR (#299) for it.

LukeTowers:
This should theoretically be fixed by https://github.com/octobercms/library/commit/48fed9b7e4c4c22ccc6b3c2e5b1273da3e2c38dd & https://github.com/octobercms/october/commit/aea4857ebaaa132499033014309bd5a6032dd66d

pikanji:
Confirmed.

PR #298, opened 2018-01-17
update ConfigWriter support config file which contains env function

buuug7:
update ConfigWriter for support write `key => env('KEY','value')` and `key => env('KEY')`.


buuug7 (commented):
all tests passed. (-_-

src/Config/ConfigWriter.php
```diff
@@ -84,16 +84,27 @@ protected function parseContentValue($contents, $path, $value)
 
         $count = 0;
         $patterns = array();
+
+        // this should be first
```
LukeTowers: Either explain this comment more thoroughly or remove it.

src/Config/ConfigWriter.php
```diff
@@ -84,16 +84,27 @@ protected function parseContentValue($contents, $path, $value)
...
         $patterns[] = $this->buildStringExpression($key, $items, '"');
         $patterns[] = $this->buildConstantExpression($key, $items);
         $patterns[] = $this->buildArrayExpression($key, $items);
 
         foreach ($patterns as $pattern) {
-            $result = preg_replace($pattern, '${1}${2}'.$replaceValue, $result, 1, $count);
-
-            if ($count > 0) {
-                break;
+            if ($pattern === $patterns[0]) {
```
LukeTowers: Why has this been added? What does it do?

tests/Config/ConfigWriterTest.php
```diff
@@ -129,5 +135,57 @@ public function testToContent()
         $this->assertArrayHasKey('database', $result['redis']['default']);
         $this->assertEquals('127.0.0.1', $result['redis']['default']['host']);
         $this->assertEquals(null, $result['redis']['default']['password']);
+
+
+        /*
+         * Test env like 'key'=>env('KEY'')
```
LukeTowers: Is your test example supposed to have a third quote in the env()?

tests/fixtures/config/sample-config.php
```diff
@@ -145,4 +145,27 @@
         ],
 
     ],
-];
+
+    // 'key'=>env('KEY'')
```
LukeTowers: Can you make some of these tests test for double quotes instead of just all single quotes?

composer.json
```diff
@@ -41,8 +41,8 @@
         "illuminate/routing": "5.5.*",
         "illuminate/validation": "5.5.*",
         "illuminate/view": "5.5.*",
-        "swiftmailer/swiftmailer": "~5.1",
-        "symfony/config": "2.7.*",
+        "swiftmailer/swiftmailer": "~6.0",
```
LukeTowers: Please don't include this change in this branch, add another commit rolling it back to the previous versions; that is a separate concern.

LukeTowers (changes requested):

src/Config/ConfigWriter.php
buuug7: for match  `'KEY' = > env('KEY')`,because the `$result = preg_replace($pattern, '${1}${2}' . $replaceValue, $result, 1, $count);` will lead the replace result to `KEY' = > env('KEY'`. the `)` will missing. so here added another rule of `$result = preg_replace($pattern, '${1}${2}' . ',' . $replaceValue . ')', $result, 1, $count);`
```

```

tests/fixtures/config/sample-config.php
buuug7: yes, i add later

src/Config/ConfigWriter.php
buuug7: already update

buuug7:
@LukeTowers already updated. please review it.

tests/Config/ConfigWriterTest.php
```diff
@@ -129,5 +138,85 @@ public function testToContent()
...
```
LukeTowers, line 144:
```suggestion
         * Test env like 'key'=>env('KEY')
```

bennothommo:
@buuug7 If you are still interested in seeing this implemented, would you mind sorting out the code style issues reported starting from here: https://github.com/octobercms/library/pull/298/checks?check_run_id=710735517#step:5:6

PR #300, opened 2018-01-26
Activate users on successful password reset

jimcottrell:
I found https://github.com/rainlab/user-plugin/issues/141 that describes the root issue this is attempting to help smooth over. As described there, I think this is a fairly rare situation, but in our case:

1. On a client project, we imported accounts from another system and sent everyone an activation email to verify the email address and set a password on the newly-created October user accounts.
2. A lot of people ignored the activation emails, as you do.
3. Those users eventually end up at the site, try to sign in with the account they know they have, fail, reset their password, and fail again, as now the password works, but they're still inactive and they have no way to resend the activation email.
4. Those users complain to the client.

Of course the easy thing to do was to change the setting to allow inactive accounts to log in, which I've done. However, it's not really ideal that now users will reset their password but still be prompted to verify their email after they log in. Since the password reset requires the receipt of a code via email just like activation does, it makes sense to me that a successful password reset should also activate the account, as the email address has now been verified. So that's what this PR does.

### DIFF
--- a/src/Auth/Models/User.php
+++ b/src/Auth/Models/User.php
@@ -223,14 +223,8 @@ class User extends Model
      */
     public function attemptActivation($activationCode)
     {
-        if ($this->is_activated) {
-            throw new Exception('User is already active!');
-        }
-
         if ($activationCode == $this->activation_code) {
-            $this->activation_code = null;
-            $this->is_activated = true;
-            $this->activated_at = $this->freshTimestamp();
+            $this->setActive();
             $this->forceSave();
             return true;
         }
@@ -288,6 +282,9 @@ class User extends Model
         if ($this->checkResetPasswordCode($resetCode)) {
             $this->password = $newPassword;
             $this->reset_password_code = null;
+            if (!$this->is_activated) { // Receiving the password reset code also verifies email address
+                $this->setActive();
+            }
             return $this->forceSave();
         }
 
@@ -664,5 +661,19 @@ class User extends Model
         $pool = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
 
         return substr(str_shuffle(str_repeat($pool, 5)), 0, $length);
+    }
+
+    /**
+     * Activate user
+     */
+    public function setActive()
+    {
+        if ($this->is_activated) {
+            throw new Exception('User is already active!');
+        }
+
+        $this->activation_code = null;
+        $this->is_activated = true;
+        $this->activated_at = $this->freshTimestamp();
     }
 }


JoakimBo:
Wouldn't it make more sense to be able to re-send the activation email than activate on password reset?

jimcottrell:
@JoakimBo I agree that there should also be a way to resend the activation email, but why should you have to reset your password and then re-send the activation email? You just received an email to reset your password. Didn't that confirm that your email address is valid? And it could be changed just in RainLab.User, but if it doesn't make sense as it is for that plugin, why would it make sense in others?

JoakimBo:
Hmm I think I understand you, If I were you I would probably try and make a PR for Rainlab.User that adds the option for a backend admin to enable a function that does the following (in sudo): "resend activation email on login for non-activated users".

Its not the biggest deal to me if this "fix" gets its place in the library but if I was a developer having used "is_activated" as a way to deactivate or ban users through my plugin and these suddenly get reactivated on login I would probably not understand why. I'm simply trying to think how already developed plugins would be affected by this.

jimcottrell:
I'm not sure I completely follow you. This "fix" (why the scare quotes?) doesn't reactivate anyone on login. If someone used "is_activated" to ban people, I suppose they would have a problem if they didn't also prevent those banned people from resetting their password, but I don't know to what extent that's even an issue. Honestly, it seems like it would be kind of poor development practice anyway, as it should probably be normalized to a separate `is_banned` field instead of trying to make "activation" take on multiple meanings.

Do you have any examples of other auth plugins that act like how you're suggesting? I'm happy to check and think more about whether this is the right layer to do what I'm proposing. Although I think re-sending RainLab.User activation emails is useful, I don't know if it's everything.

It's probably all academic because there's a good chance this PR will never be officially reviewed anyway, but hey, whatever.

JoakimBo:
So I'm obviously tired and shouldn't be looking at code right now :) , you can probably ignore most of what I've been on about, I was a bit ahead of myself, sorry about that.

jimcottrell:
No worries. I don't think it's bad to get pushback on whether it's the right scope for a change. My thinking has just been that the core library uses activation codes at its level of abstraction, so it seems that using `is_activated` to represent email verification is kind of implied at this level too (though the activation email itself is of course left to plugins), and that banning, deactivating, etc. would tend to use a different mechanism. I could be thinking about it differently than everyone else though, so I don't mind reconsidering. If you want to look at it again later from a better-rested perspective and see what issues you see, I'm open to any thoughts.

JoakimBo:
No you're not wrong, the banning is handled differently.

Atleast for now I guess one could atleast think about the scope, something like "would this be a good application for every instance where the auth/model/user is being used". With that framing in mind I don't really have an issue with this. Unless it's been accepted by then, perhaps I'll give another go at poking holes at this tomorrow ;)

LukeTowers:
> It's probably all academic because there's a good chance this PR will never be officially reviewed anyway, but hey, whatever.

Somebody feeling a bit ignored over there @jimcottrell? 😉 

While I agree with your premise that a successful password reset should in theory indicate a valid email; I would like to point out that (even though it might currently be used in that manner) `is_activated` != `is_valid_email`. I would like to see @daftspunk's thoughts on this before deciding if it should be merged or not.

jimcottrell:
I agree to an extent, but if the intent is for `is_activated` to be turned on and off or to be solely open to plugin interpretation, it would make more sense to me if the field were `is_active` or if the core model didn't offer activation codes and that were just something RainLab.User added on to the activation process. True, you can't equate an activated account with a verified email address, but I guess to me it seems like if you're supporting tying activation to verification in core (or else why have activation codes?), it's not completely unreasonable to think that derivative plugins would carry that through. Plugins don't even have to rely on email, but it's hard for me to imagine a case where a successful password reset wouldn't also imply the same contact verification that an activation code would handle.

Anyway, I'm not against the idea that maybe this is something only RainLab.User should do for what it considers `is_activated` to mean, and I get where you guys are coming from. When I wrote this change, I decided to submit it to core as it already seems to manifest some opinions about these matters, but I understand if this isn't seen to be worth it in core.

> Somebody feeling a bit ignored over there @jimcottrell? 😉

Ha, well, to be fair, something sitting for months with no feedback is kind of the definition of ignored, so I'm not sure that's an unreasonable feeling. I do understand the situation, of course, and everything can't be dealt with immediately even in the best circumstances, but it's not super encouraging for a contributor. :)

LukeTowers:
It comes down to the fact that the squeaky wheel gets the grease @jimcottrell. Your contributions are appreciated, but if it isn't a change that I can logically process in less than five minutes, or it doesn't come up in my usage of the project, or it's never brought up as a reminder - then any of the multitude of other issues that meet one or all of that criteria will get handled first. 

Keep in mind that it's been close to a one man show around here for the past few months as @daftspunk has been busy with life and is only just now coming back starting around the beginning of April.

jimcottrell:
@LukeTowers Yeah, I get that, and you've done a good job. There's some frustration to the contribution process though. I'm hopeful that some of that will clear up.

LukeTowers:
@jimcottrell did this ever get implemented at the RainLab.User level? I feel like that would be a better place for this.

jimcottrell:
@LukeTowers I haven't been following the RainLab.User development, but I never submitted anything for this there, so if it's in there it wasn't my doing. I still think it's a weird flow to reset the password then have to activate, and I don't think the what-ifs ever convinced me this isn't appropriate for core if password reset and activation codes are core functionality. Still, we've lived with it for our transferred users, and the lack of comment from others implies that it's not a state most people end up in.

LukeTowers:
@bennothommo do you have any input on this?

bennothommo:
@LukeTowers this would be resolved by having some way to re-send the activation email, as [metioned here](https://github.com/rainlab/user-plugin/issues/433#issuecomment-618817700). I'll put this up as an issue to work on with the User plugin (I don't think this needs to extend to the administration users).

LukeTowers:
@bennothommo I'm not sure that users being able to resend their activation email after logging in is a solution to the issue reported by @jimcottrell originally

bennothommo:
@LukeTowers why not? The person resets their password, then when they try to log in again, they get a warning saying the account is not activated. At the point, we could give them the option to re-send the activation email.

LukeTowers:
I'm going to close this for now in favour of leaving this decision up to individual projects / implementations of the Auth system rather than enforcing it at this level.

PR #302, opened 2018-02-09
Fix escaping of multiline values in INI

Fl0Cri:
Fixes https://github.com/rainlab/pages-plugin/issues/327

I've added a test case to demonstrate the problem but I could not install composer dependencies on php 7.0.6 (need 7.0.8) so I didn't run it to check.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,8 +41,8 @@
         "illuminate/routing": "5.5.*",
         "illuminate/validation": "5.5.*",
         "illuminate/view": "5.5.*",
-        "swiftmailer/swiftmailer": "~5.1",
-        "symfony/config": "2.7.*",
+        "swiftmailer/swiftmailer": "~6.0",
+        "symfony/config": "3.3.*",
         "symfony/dom-crawler": "~3.1",
         "symfony/css-selector": "~3.1"
     },
--- a/src/Parse/Ini.php
+++ b/src/Parse/Ini.php
@@ -251,7 +251,10 @@ class Ini
         }
 
         // String (default)
-        return '"'.str_replace('"', '\"', $value).'"';
+        $value = str_replace('"', '\"', $value);
+        $value = preg_replace('~\\\"([\r\n])~', '\\\"""$1', $value);
+
+        return '"'.$value.'"';
     }
 
     /**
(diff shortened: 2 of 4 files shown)


LukeTowers:
Could you update your PHP to run the tests?

Fl0Cri:
Okay, I'm on an other pc and I could run the tests.
- It's ok on Windows with php 7.1.7
- But on linux (using wsl) with php 7.1.11, I get an error telling me that the test fail, despite that both arrays are equals!? Have you already seen this?
```
1) IniTest::testMultilinesValues
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'var' => '\Test\Path\'
     'editorContent' => '<p>Some
         <br>"Multi-line"
         <br>text
     </p>'
 )
```

Fl0Cri:
Just installed and tested with php 7.2.2 on Windows, the test passes so it's maybe related to wsl or to a php bug in 7.1.11

LukeTowers:
@Fl0Cri have you tried testing it on anything else?

Fl0Cri:
I just found what's going wrong with my tests: Because the repo was cloned on Windows, git wrote everything with CRLF line endings to the filesystem.

The parse method run on Windows was also outputing line endings as CRLF so the test passes, but when run from linux it is outputing LF. The diff viewer of phpunit doesn't see that so it tells that both array are equals despite that the test is failing.

So if you clone the from a Linux Git on a Linux filesystem an run with a Linux environment everything should be ok.

LukeTowers:
Would there be an issue with forcing both to use the same EOL character when testing it so that this doesn't happen to anyone else? It's a tricky thing that might be difficult to catch by someone unaware of its existence; so I think it might be better if we can change the test so that this can't happen.

Fl0Cri:
Ok, the tests are passing now on my environment.

LukeTowers:
So the PR is now complete? Does it still fix the original issue? @Fl0Cri

Fl0Cri:
Yes, I've put the fix on the prod website where I noticed the bug and it's running fine with it.
My users did a lot of content editing with static pages and had no problems.

LukeTowers:
Awesome, thanks for your work on this!

PR #306, opened 2018-02-20
Implement adder for model attribute casting

scottbedard:
Closes https://github.com/octobercms/october/issues/3407

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -672,6 +672,17 @@ class Model extends EloquentModel
     //
     // Adders
     //
+    
+    /**
+     * Add attribute casts for the model.
+     * 
+     * @param  array              $attributes
+     * @return void
+     */
+    public function addCasts($attributes)
+    {
+        $this->casts = array_merge($this->casts, $attributes);
+    }
 
     /**
      * Adds a datetime attribute to convert to an instance of Carbon/DateTime object.
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -676,7 +676,7 @@ class Model extends EloquentModel
     /**
      * Add attribute casts for the model.
      * 
-     * @param  array              $attributes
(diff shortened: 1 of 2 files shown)


src/Database/Model.php
```diff
@@ -672,6 +672,17 @@ protected function performDeleteOnRelations()
     //
     // Adders
     //
+    
+    /**
+     * Add attribute casts for the model.
+     * 
+     * @param  array              $attributes
```
LukeTowers: What's with all the space between array and attributes?

scottbedard: Heh, couldn't tell you... I guess that's just what my editor decided to do. I'll go ahead and trim that.

LukeTowers:
Could you add some information to the docs as well for this and the other protected setters if there isn't any already?

scottbedard:
@LukeTowers I couldn't find any, so I've added a blurb to the end of our model docs.

src/Database/Model.php
```diff
@@ -672,6 +672,17 @@ protected function performDeleteOnRelations()
...
     // Adders
     //
+    
+    /**
+     * Add attribute casts for the model.
+     * 
+     * @param  array $attributes
+     * @return void
+     */
+    public function addCasts($attributes)
```
LukeTowers: Could you make this match the other setters? I.e. pass multiple strings or an array of strings?

scottbedard: No, that API doesn't make sense in this context. Unlike the other adders, a model's `casts` value is an associative array.

LukeTowers: Fair enough

scottbedard:
Woo! Thanks @LukeTowers for getting this merged so quickly!

LukeTowers:
Thanks for the contribution @scottbedard!

PR #308, opened 2018-02-22
Simplify retrieval of CMS objects to one cache request instead of two

Josniii:
One issue I've noticed with OctoberCMS is that it requires two cache requests to retrieve a CMS object like a page or a partial from the cache. First, the system calls `$cache->has($key)` to check if the key exists. This interally calls `get()`, resulting in a request being sent. Then, the system calls either `remember()` or `rememberForever()` to actually retrieve and/or store the result, resulting in another `get()` internally.

Since the first request is only used for a simple null check, I think it would be better to simply call `get()` and do the null check ourselves. Instead of calling `remember()` or `rememberForever()`, we call `put()`and `forever()`.

This way, we save a request when retrieving objects from the cache.


LukeTowers:
Doesn't retrieving an item from the cache leave it in memory?

Josniii:
I'd think that the cache server would keep it in memory, yes. However, the latency cost of sending a cache request can in some complex setups be pretty high.For example, if you're running an external cache server. In my testing with Elasticache, this resulted in a significant speed up in runtime.

LukeTowers:
Could this improvement be implemented on a lower level? I.e. implementing an in-memory cache for `Cache::get()` similar to the in-memory DB cache that we integrated a while ago? https://github.com/octobercms/library/pull/253

Josniii:
@LukeTowers - I looked into this a bit further and it looks like we already have [something like that.](https://github.com/octobercms/library/blob/master/src/Halcyon/MemoryCache.php)
At the end of `getCached()` the result is indeed stored in memory: 
```return MemoryCache::$cache[$key] = $result;```
And we do use any stored results as well:
```
if (array_key_exists($key, MemoryCache::$cache)) {
    return MemoryCache::$cache[$key];
}
```
So to me, it seems like we already have such a system implemented.

However, if we don't have the result in memory already, the cache retrieval is inefficient (because `has()` makes a request for the sole purpose of a null check), that's what this PR optimizes. To most people this is probably a minor issue, but when you have an external cache server, each request can take quite a while.

**Comment by LukeTowers** [comment]
So the issue is that the cache driver being used doesn't implement an in-memory cache when `has()` is called?

**Comment by Josniii** [comment]
I guess so. Ideally I'd say we cache the result that `has()` retrieves and we wouldn't have to worry about this. Any ideas on how to do that best?

**Comment by LukeTowers** [comment]
That could be tricky, since it seems that the `has()` method is implemented by each cache driver separately; what's the repo for the cache driver that you're using? https://github.com/octobercms/library/blob/master/src/Halcyon/Builder.php#L703

**Comment by Josniii** [comment]
`get_class($cache)` gives me `Illuminate\Cache\Repository` - so that would be the [Laravel repository.](https://github.com/laravel/framework). I'm using memcached.

**Comment by LukeTowers** [comment]
Well, we could either try to make a PR to laravel adding in memory cache for get(), or we could put our own layer on top of it and use dependency injection to replace it with our layer that extends theirs; implementing the in memory cache in our layer. Basically the point of this entire repo which is to put that layer between OctoberCMS projects and Laravel.

@Josniii Thoughts?

**Comment by Josniii** [comment]
@LukeTowers Not sure, but I think that in-memory cache might be a bit too general for Laravel since it is used for so many different types of applications? It makes sense in a CMS where we like to reuse small fragments a lot - maybe not so much for a lot of other types of applications. So I'd say the best place to put such a layer would be in October.

**Comment by Josniii** [comment]
Here's how I think we could do this. If you could see a more elegant way of implementing this I'd like some input:

- Replace the default `CacheManager` with a `MemoryCacheManager` through dependency injection, that extends the default one. This could be done in the `HalcyonServiceProvider`. Override `repository()` so it instances a custom MemoryRepository class. We could put a config check here so people can decide if they want this enabled or not.
- Implement a `MemoryRepository` that extends the Laravel `Repository`. Here we can override `get()` so that it stores the result in memory before calling `parent::get()` to retrieve things the first time. We'll probably also need to override any functions used to write to the cache, that way we can update the cache as needed as well.

How does this sound?

**Comment by Josniii** [comment]
Hi @LukeTowers 
I finally had some time to take a crack at this.
I've changed this around to try and implement what I suggested above. I might have missed some neccessary overrides for the `MemoryRepository`, but it seems to work fine. My testing sees a large decrease in external cache requests as hoped.

File: src/Halcyon/Builder.php [hunk 1]
```diff
@@ -244,8 +244,8 @@ public function first()
      * @param  array  $columns
      * @return \October\Rain\Halcyon\Collection|static[]
      */
-     public function get($columns = ['*'])
-     {
+    public function get($columns = ['*'])
```
LukeTowers: Indentation 4 spaces please

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,136 @@
...
+
+
+    /**
+     * Remove all items from the cache.
+     *
+     * @return bool
+     */
+    public function flush()
+    {
+        $this->cache = [];
```
LukeTowers: Should this be a call to `$this->flushInternalCache()` instead?

LukeTowers:
@Josniii awesome work, looking forwards to getting this merged! What are you ideas around making this behaviour configurable? Also, would it be possible to add some unit testing for this functionality?

Josniii:
@LukeTowers I think we should add a flag to ```config/cache.php``` like `'request_cache' => true,` to let people opt-out of the feature if they have something else that conflicts with this. Other than that, I don't really have any other suggestions. I've added ```MemoryRepositoryTest.php``` with tests for the feature, let me know if you think I'm missing anything 👍

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Retrieve an item from the internal memory cache without trying the external driver.
+     * Used in testing
+     *
+     * @param $key
+     * @param null $default
+     * @return mixed
+     */
+    public function getFromMemoryCache($key)
```
LukeTowers: If these methods are used in testing, then they need to be what the other (`get` & `put`) methods call to interact with the memory cache

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->flushInternalCache();
+        parent::flush();
+    }
+
+    /**
+     * Retrieve an item from the internal memory cache without trying the external driver.
+     * Used in testing
+     *
+     * @param $key
+     * @param null $default
```
LukeTowers: Remove this please

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Remove an item from the cache.
+     *
+     * @param  string  $key
+     * @return bool
+     */
+    public function forget($key)
+    {
+        unset($this->cache[$key]);
+        return parent::forget($key); // TODO: Change the autogenerated stub
```
LukeTowers: Deal with TODO

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,166 @@
...
+     * Store an item in the cache indefinitely.
+     *
+     * @param  string  $key
+     * @param  mixed   $value
+     * @return void
+     */
+    public function forever($key, $value)
+    {
+        $this->cache[$key] = $value;
+        parent::forever($key, $value); // TODO: Change the autogenerated stub
```
LukeTowers: Deal with TODO

tests/Halcyon/MemoryRepositoryTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->repository->decrement('baz', 10);
+        $this->assertEquals(-10, $this->repository->getStore()->get('baz'));
+        $this->assertEquals(-10, $this->repository->getFromMemoryCache('baz'));
+    }
+
+    public function testBothCachesForgetKeys()
+    {
+        $this->repository->put('foo', 'bar', 10);
+        $this->repository->forget('foo');
+        $this->assertNull($this->repository->get('foo'));
```
LukeTowers: Should this be getFromMemoryCache instead?

tests/Halcyon/MemoryRepositoryTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->assertNull($this->repository->getStore()->get('foo'));
+    }
+
+    public function testBothCachesAreFlushedCorrectly()
+    {
+        $this->repository->put('foo', 'bar', 10);
+        $this->repository->put('baz', 'qux', 10);
+        $this->repository->put('quux', 'corge', 10);
+        $this->assertEquals('bar', $this->repository->get('foo'));
```
LukeTowers: Should these `get()` calls be changed to `getFromMemoryCache()` for testing the memory cache specifically?

LukeTowers:
Re the config, I think having `'disableRequestCache' => false` added to `config/cache.php` with an appropriate comment block would be ideal. This will enable the cache by default for updates (since updating October does not modify config files) while still providing the option to disable it if required. If you could submit a PR to octobercms/october; that would be awesome!

Thanks again for your work on this so far @Josniii, I appreciate it!

Josniii:
@LukeTowers I've gone through your reviews and made changes based on your suggestions, please take a look and let me know if you have further comments. I'll try and get the PR for the CMS repository done tomorrow, I'll tag you there once it's ready. Thanks for being thorough!

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,170 @@
...
+     * @param  string|array $key
+     * @return mixed
+     */
+    public function get($key, $default = null)
+    {
+        if(is_array($key)) {
+            return $this->many($key);
+        }
+
+        if($value = $this->getFromMemoryCache($key)) {
```
LukeTowers: Does this support a falsey value being stored in the cache?

src/Halcyon/MemoryRepository.php
```diff
@@ -29,7 +29,9 @@ public function get($key, $default = null)
             return $this->many($key);
         }
 
-        if($value = $this->getFromMemoryCache($key)) {
+        $value = $this->getFromMemoryCache($key);
+
+        if(isset($value)) {
```
LukeTowers: I'd prefer that you use `!is_null($value)` instead for this, since that is what you are actually testing for, the fact that `isset()` returns `false` for `null` is not immediately obvious unless you're already aware of that fact so for readability I would prefer the `!is_null()` check instead. Thoughts?

Josniii: I agree, I was actually checking out whether or not `isset()` returns true for `null` myself. Better to just use `!is_null()` in the first case.

Josniii:
I've added a PR to OctoberCMS [here.](https://github.com/octobercms/october/pull/3477) I've also implemented `isEnabled` in `MemoryCacheManager` so it hooks to this config flag.

src/Halcyon/MemoryCacheManager.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    public function repository(Store $store)
+    {
+        $repository = new MemoryRepository($store);
+        return $repository;
+    }
+
+    public static function isEnabled()
+    {
+        return !Config::get('cache.disableRequestCache', false);
```
LukeTowers: Are you able to use the `$this->app->config` methodology for accessing the config value instead of pulling in the Config facade? It's used in other places in this repo so I'd prefer to use it here too

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @var array
+     */
+    protected $cache = [];
+
+    /**
+     * Retrieve an item from the cache by key.
+     *
+     * @param  string|array $key
+     * @return mixed
```
LukeTowers: I know I said to remove the `@param $default` that was here earlier but I swear the `$default = null` parameter wasn't present in the method signature when I said that :) Could you re-add it?

src/Halcyon/MemoryRepository.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param  string  $key
+     * @param  mixed   $value
+     * @param  \DateTimeInterface|\DateInterval|float|int  $minutes
+     * @return void
+     */
+    public function put($key, $value, $minutes = null)
+    {
+        if (is_array($key)) {
+            $this->putMany($key, $value);
```
LukeTowers: See https://github.com/laravel/framework/pull/13868#issuecomment-223764289 for a reason why the `is_array` check should be removed entirely

LukeTowers: See https://github.com/laravel/framework/issues/23729 for a reason why `is_array` check should be removed entirely here.

src/Halcyon/MemoryCacheManager.php
Josniii: I don't think I can since this is a static function, unless I'm missing something? I used the facade because I had seen that other places as well.

src/Halcyon/MemoryRepository.php
Josniii: I think that was in `getFromMemoryCache` I removed it. So it's just missing here :)

src/Halcyon/MemoryCacheManager.php
LukeTowers: Fair enough, you'd have to implement it as a facade which would be a bit ridiculous

src/Halcyon/MemoryRepository.php
LukeTowers: @taylorotwell has decided to not remove it (even though it's not documented anywhere and the argument mapping doesn't line up) so we'll have to leave it here for compatibility with Laravel.

Josniii: Indeed, I agree it looks very strange, although I guess it technically works. You'd have to call it like `put(['foo' => 'bar'], 10)` with the `$value` param effectively being used as `$minutes`.

PR #310, opened 2018-03-06
Develop model trait behaviors

joakimbo:
This PR is in reference to [Issue 3427 in October repo](https://github.com/octobercms/october/issues/3427)


src/Database/Behaviors/Encryptable.php
```diff
@@ -0,0 +1,19 @@
...
+ * Encryptable trait as behaviour
+ *
+ * @package october\database
+ * @author JoakimBo
+ */
+class Encryptable extends ModelTraitBehavior
+{
+    use \October\Rain\Database\Traits\Encryptable;
+
+    public function __construct($model)
```
LukeTowers: You don't need the `__construct()` method in any of these behaviors because the ModelTraitBehavior takes care of it for you

JoakimBo:
Okay, I included them since I was in the mindset that we would override their boot methods.

src/Database/Behaviors/Encryptable.php
JoakimBo: Okay, I included them since I was in the mindset that we would override their boot methods.

LukeTowers: Right, do it piece meal though

JoakimBo:
@LukeTowers So now, the "only" thing stopping purgeable from working is the static extending, you mentioned that one could override the static extend func and access the model(and class) but that hurts on the inside.
```
public static function bootPurgeable()
{
    if (!property_exists(get_called_class(), 'purgeable'))
        throw new Exception(sprintf(
            'You must define a $purgeable property in %s to use the Purgeable trait.'
            , get_called_class()
        ));

    /*
     * Remove any purge attributes from the data set
     */
    static::extend(function($model){
        $model->bindEvent('model.saveInternal', function() use ($model) {
            $model->purgeAttributes();
        });
    });
 }
```

LukeTowers:
How many traits use the extend method?

JoakimBo:
Among the ones availible [here](https://octobercms.com/docs/database/traits) 9 out of 11, although
Sortable uses `static::created` and `static::addGlobalScope`.   
SoftDelete uses `static::restoring` and `static::restored`.
Dont know if they are also troublemakers.

JoakimBo:
@LukeTowers Unsure if latest commit is what we want but its how I solved it as a first try.

LukeTowers:
@JoakimBo classes that use the extendable trait (Extendable base) are supposed to use `$this->methodExists()` and `$this->propertyExists()` I believe

JoakimBo:
@LukeTowers Should be fixed now.

JoakimBo:
@LukeTowers So I'm sure that some of the code in its current state is flawed but I'm hoping to be able to test it soon.
With regards to dynamic properties being pushed to the database, purging them isnt something that we should automate in the behaviors? I dont mean to say that all dynamic properties should get purged as default I'm just wondering if we should leave this entirely up to the developer in the context of these behaviors?

LukeTowers:
I'd say it would be simpler to leave it up to the developer, perhaps with a detailed explanation in the documentation when using these behaviours of that requirement.

LukeTowers:
@joakimbo did you mean to close this?

PR #315, opened 2018-03-11
Code to work with the pull request 'Add support for more than one nes…

SiyuQian:
…ted folder in the CMS file browser' #3441

Support max two levels nested folders in the Partial folder

Please see more details in https://github.com/octobercms/october/pull/3441

Cheers,
Siyu Qian

### DIFF
--- a/src/Halcyon/Datasource/FileDatasource.php
+++ b/src/Halcyon/Datasource/FileDatasource.php
@@ -100,7 +100,8 @@ class FileDatasource extends Datasource implements DatasourceInterface
         }
 
         $it = new RecursiveIteratorIterator(new RecursiveDirectoryIterator($dirPath));
-        $it->setMaxDepth(1); // Support only a single level of subdirectories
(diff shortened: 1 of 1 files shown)


LukeTowers:

bennothommo:
@LukeTowers just to note, this circles back to our discussion on the `select` method from `FileDatasource`, where it only allows one level of subdirectories. If that was allowed extra levels, I believe this issue would be resolved, although, I assume @daftspunk had a reasoning for limiting it to one level, whether it's a technical or UI reasoning.

LukeTowers:
@bennothommo yep. I believe the reason was for the UI, if we can implement a refactor the UI that supports arbitrary folder depth cleanly (I think the asset list already does this, but changes would need to be ported over) then we can remove the limit.

bennothommo:
@LukeTowers cool, I'll take a look at this in 467.

bennothommo:
Oops, this is a PR, not an issue. I'll leave myself assigned as a reminder.

PR #318, opened 2018-04-07
Add transaction

immeyti:


src/Auth/AuthException.php
```diff
@@ -34,7 +35,7 @@ public function __construct($message = "", $code = 0, Exception $previous = null
         $this->softErrors = !Config::get('app.debug', false);
 
         if ($this->softErrors) {
-            $message = $this->errorMessage;
+            $message = Lang::has('backend::lang.account.details_not_match')  ? Lang::get('backend::lang.account.details_not_match') : $this->errorMessage ;
```
LukeTowers: This should be handled at a higher level (in the backend module itself), not in the Auth library directly.

LukeTowers:
This should be handled at a higher level (in the backend module itself), not in the Auth library directly.

PR #319, opened 2018-04-12
Update AuthException.php

datune:
Provide a translation for the single error message that gets displayed when the application is in production. @see https://github.com/octobercms/october/issues/2106


LukeTowers:
See https://github.com/octobercms/october/issues/2106#issuecomment-380870704

PR #320, opened 2018-04-17
Determine if required_if field is mandatory

FaZeRs:


src/Database/Traits/Validation.php
```diff
@@ -407,6 +407,16 @@ public function isAttributeRequired($attribute)
             return $this->isAttributeRequired($requiredWith);
         }
 
+        if (strpos($ruleset, 'required_if') !== false) {
+            $data = $this->getValidationAttributes();
+            $requiredIf = substr($ruleset, strpos($ruleset, 'required_if') + 12);
```
LukeTowers: What does the $ruleset string look like here? It seems to me from an initial look at it that 12 is an arbitrary number that is subject to breaking.

FaZeRs: 'required_if:anotherfield,value' and $requiredIf variable here is 'anotherfield,value'

LukeTowers: what if there is more than one rule in the rule set?

FaZeRs: It wont break if there is rule set before this, but it will break if there is rule set after this.

LukeTowers:
Please also provide unit tests for this; ideally unit test the whole isAttributeRequired method

bennothommo:

PR #322, opened 2018-05-17
Pass $data by reference

datune:
In order to allow for manipulation of the $data contents it is required to pass it by reference.

### DIFF
--- a/src/Mail/Mailer.php
+++ b/src/Mail/Mailer.php
@@ -35,8 +35,8 @@ class Mailer extends MailerBase
          * Extensibility
          */
         if (
-            ($this->fireEvent('mailer.beforeSend', [$view, $data, $callback], true) === false) ||
-            (Event::fire('mailer.beforeSend', [$view, $data, $callback], true) === false)
+            ($this->fireEvent('mailer.beforeSend', [$view, &$data, $callback], true) === false) ||
+            (Event::fire('mailer.beforeSend', [$view, &$data, $callback], true) === false)
         ) {
             return;
         }


tschallacka:
I would say no. The purpose of these events are to test if it's okay to send the email with the provided data, not to modify the data.

I would rather suggest you implement modification events specifically for modifying the data, but that's just my opinion.

LukeTowers:
@tschallacka why shouldn't $data be modified in the beforeSend event?

LukeTowers:
Also @datune what specifically do you need this for?

datune:
Why is there a need to provide additional events, I see absolutely no danger in passing the data by reference. Where does it say the purpose of these events is limited to data checking only? 

@LukeTowers Simply to allow for manipulation from within other plugins. Filters, ensuring certain content is present, stripping or replacing etc...

LukeTowers:
@datune could you provide a specific example of it's use that you need it for right now?

datune:
@LukeTowers No. I created this ticket in May, it's September. Sorry, I have forgotten the original reason. Lesson learned!

LukeTowers:
This can wait until we update the docblock for the API event docs then

tschallacka:
My initial comment was after a cursory glance over the functionality as a whole.

Say that if we pass it by reference then https://github.com/octobercms/library/blob/a1b385be84044b8cba8dabb999591f983254e3e9/src/Mail/Mailer.php#L305-L306 should be pass by reference too to remain consistent.

But my main gripe about passing it by reference is that it might pose a data integrity problem by a wrongly coded plugin, since by passing things by reference things can be removed. And one has to assume that when things are provided to the $data that they should not be removed only added or modified. Adding and modifying is fine, but overriding the value of $data totally might lead to exceptions.

Imagine the plugins that exist now that listen casually to this event and at some time reset `$data = []` so they have a variable to re-use by intent or just convenience coding. They can do that right now without any harm because the moment they modify the variable or array it's duplicated by the php engine.

I would rather have an instance of the array passed normally and have that returned and then merged with an array merge and if false is returned then it aborts.

Passing by reference is a powerfull tool that can too easely be harmed by someone who doesn't know what they are doing.

Just my thoughts about the matter about any pitfalls such a change might bring.

datune:
@tschallacka 

Your arguments are valid points, thank you. I admit I have not considered the case where (for whatever reason) other developers might reset $data. I like your suggestion of returning an array and merging it, it's the least obtrusive and should not introduce any possible breaking changes.

bennothommo:

PR #323, opened 2018-05-25
Provide a simple way to test that a field exists in a model

firemankurt:
When creating a behavior class where most models might have for example, a user_id field, it is very helpful to find out if the current $model you are working with has that field so you can fill it or process it in some cases.

For example:
CompanyBehavior.php
             
    protected function createModel() {
		$this->model = new $this->modelClass;
		// set company of new model
		$this->model->company_id = $this->user()->company_id;

		if ($this->hasAttribute('user_id')) {
			$this->model->user_id = $this->user()->id;
		}
		return $this->model;
	}  

This is a resubmital of: https://github.com/octobercms/library/pull/321

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -971,7 +971,7 @@ class Model extends EloquentModel
      * @param  string  $key
      * @return bool
      */
-    public function hasAttribute($attr) {
-        return array_key_exists($attr, $this->attributes);
+    public function hasAttribute($key) {
+        return array_key_exists($key, $this->attributes);
     }
 }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -967,11 +967,11 @@ class Model extends EloquentModel
     
     /**
      * Determine if an attribute exists in the model.
-     * note: nullable attributes must set raw value of null before this works
      * @param  string  $key
      * @return bool
      */
-    public function hasAttribute($key) {
-        return array_key_exists($key, $this->attributes);
+    public function hasAttribute($key) 
+    {
+        return array_key_exists($key, $this->attributesToArray());
     }
 }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -971,7 +971,7 @@ class Model extends EloquentModel
      * @param  string  $key
      * @return bool
      */
-	public function hasAttribute($attr) {
-		return array_key_exists($attr, $this->attributes);
-	}
+    public function hasAttribute($attr) {
+        return array_key_exists($attr, $this->attributes);
+    }
 }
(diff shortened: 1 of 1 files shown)


src/Database/Model.php
```diff
@@ -963,4 +963,15 @@ public function hasSetMutator($key)
...
     }
+    
+    
+    /**
+     * Determine if an attribute exists in the model.
+     * note: nullable attributes must set raw value of null before this works
+     * @param  string  $key
+     * @return bool
+     */
+	public function hasAttribute($attr) {
```
LukeTowers: four spaces not tabs

src/Database/Model.php
```diff
@@ -963,4 +963,15 @@ public function hasSetMutator($key)
     {
         return $this->methodExists('set'.Str::studly($key).'Attribute');
     }
+    
+    
+    /**
+     * Determine if an attribute exists in the model.
+     * note: nullable attributes must set raw value of null before this works
+     * @param  string  $key
```
LukeTowers: variable name here is $attr, not $key

LukeTowers: you could have just changed the docblock instead ;)

src/Database/Model.php
```diff
@@ -963,4 +963,15 @@ public function hasSetMutator($key)
...
+     * @return bool
+     */
+    public function hasAttribute($key) {
```
tschallacka: This is not PSR-4 compliant, the `{` should be on the next line.

tschallacka:
It seems that this has been denied for issues by the laravel repo several times   
https://github.com/laravel/framework/pull/22249  
https://github.com/laravel/framework/pull/22250  
https://github.com/laravel/framework/pull/23375  

You should also use `array_key_exists($key, $model->attributesToArray())` as per https://github.com/laravel/framework/pull/22250#issuecomment-349975338 to prevent skipping over the attributes that are retrieveable via getter.

LukeTowers:
@firemankurt if you could make the changes suggested by @tschallacka that would be appreciated. I don't mind being better than Laravel by adding this into October :) 
Will require documentation though.

LukeTowers:
@firemankurt could you add this to the documentation in the Models section?

LukeTowers:
@firemankurt are you still interested in this?

firemankurt:
Yes

LukeTowers:
@firemankurt could you add this to the documentation in the Models section?

firemankurt:
Hey Luke,
I'm still hoping this will be added to October.
Thanks

bennothommo:
@firemankurt we're still waiting on your documentation update for the Models docs. Did you happen to have a look at that?

PR #324, opened 2018-06-02
Adds Purgeable trait as behavior

joakimbo:
- Adds ability to store dynamicProperties in extensionData
- Adds ability to get extensionData
- Adds the Purgeable trait as a Behavior

### DIFF
--- a/src/Database/Behaviors/Purgeable.php
+++ b/src/Database/Behaviors/Purgeable.php
@@ -1,0 +1,125 @@
+<?php namespace October\Rain\Database\Behaviors;
+
+use Exception;
+
+class Purgeable extends \October\Rain\Extension\ExtensionBase
+{
+    /**
+     * @var array List of attribute names which should not be saved to the database.
+     *
+     * protected $purgeable = [];
+     */
+
+    protected $parent;
+
+    public function __construct($parent)
+    {
+        $this->parent = $parent;
+        $this->bootPurgeable();
+    }
+
+    /**
+     * @var array List of original attribute values before they were purged.
+     */
+    protected $originalPurgeableValues = [];
+
+    /**
+     * Boot the purgeable trait for a model.
+     * @return void
+     */
+    public function bootPurgeable()
+    {
+        if (!$this->parent->propertyExists('purgeable'))
+            throw new Exception(sprintf(
+                'You must define a $purgeable property in %s to use the Purgeable trait.', get_class($this->parent)
+            ));
+
+        /*
+         * Remove any purge attributes from the data set
+         */
+        $model = $this->parent;
+        $model->bindEvent('model.saveInternal', function() use ($model) {
+            $model->purgeAttributes();
+        });
+
+    }
+
+    /**
+     * Adds an attribute to the purgeable attributes list
+     * @param  array|string|null  $attributes
+     * @return $this
+     */
+    public function addPurgeable($attributes = null)
+    {
+        $attributes = is_array($attributes) ? $attributes : func_get_args();
+
+        $this->parent->purgeable = array_merge($this->parent->purgeable, $attributes);
+
+        return $this->parent;
+    }
+
+    /**
+     * Removes purged attributes from the dataset, used before saving.
+     * @param $attributes mixed Attribute(s) to purge, if unspecified, $purgable property is used
+     * @return array Current attribute set
+     */
+    public function purgeAttributes($attributesToPurge = null)
+    {
+        if ($attributesToPurge !== null) {
+            $purgeable = is_array($attributesToPurge) ? $attributesToPurge : [$attributesToPurge];
+        }
+        else {
+            $purgeable = $this->getPurgeableAttributes();
+        }
+
+        $dynamic_attributes = $this->parent->getExtensionData()['dynamicProperties'];
+
+        $attributes = array_merge($this->parent->getAttributes(), $dynamic_attributes);
+        $cleanAttributes = array_diff_key($attributes, array_flip($purgeable));
+        $originalAttributes = array_diff_key($attributes, $cleanAttributes);
+
+        if (is_array($this->originalPurgeableValues)) {
+            $this->originalPurgeableValues = array_merge($this->originalPurgeableValues, $originalAttributes);
+        }
+        else {
+            $this->originalPurgeableValues = $originalAttributes;
+        }
+
+        return $this->parent->attributes = $cleanAttributes;
+    }
+
+    /**
+     * Returns a collection of fields that will be hashed.
+     */
+    public function getPurgeableAttributes()
+    {
+        return $this->parent->purgeable;
+    }
+
+    /**
+     * Returns the original values of any purged attributes.
+     */
+    public function getOriginalPurgeValues()
+    {
+        return $this->originalPurgeableValues;
+    }
+
+    /**
+     * Returns the original values of any purged attributes.
+     */
+    public function getOriginalPurgeValue($attribute)
+    {
+        return isset($this->originalPurgeableValues[$attribute])
+            ? $this->originalPurgeableValues[$attribute]
+            : null;
+    }
+
+    /**
+     * Restores the original values of any purged attributes.
+     */
+    public function restorePurgedValues()
+    {
+        $this->parent->attributes = array_merge($this->parent->getAttributes(), $this->originalPurgeableValues);
+        return $this->parent;
+    }
+}
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -21,9 +21,10 @@ trait ExtendableTrait
      * @var array Class reflection information, including behaviors.
      */
     protected $extensionData = [
-        'extensions'     => [],
-        'methods'        => [],
-        'dynamicMethods' => []
+        'extensions'        => [],
+        'methods'           => [],
+        'dynamicMethods'    => [],
+        'dynamicProperties' => []
     ];
 
     /**
@@ -181,6 +182,8 @@ trait ExtendableTrait
             $this->{$dynamicName} = $value;
         }
 
+        $this->extensionData['dynamicProperties'][$dynamicName] = $value;
+
         self::$extendableGuardProperties = true;
     }
 
@@ -271,6 +274,11 @@ trait ExtendableTrait
             isset($this->extensionData['methods'][$name]) ||
             isset($this->extensionData['dynamicMethods'][$name])
         );
+    }
+
+    public function getExtensionData()
+    {
+        return $this->extensionData;
     }
 
     /**


src/Extension/ExtendableTrait.php
```diff
@@ -181,6 +182,8 @@ public function addDynamicProperty($dynamicName, $value = null)
             $this->{$dynamicName} = $value;
         }
 
+        $this->extensionData['dynamicProperties'][$dynamicName] = $value;
```
LukeTowers: @joakimbo why is this necessary?

LukeTowers: @joakimbo if you need to store a list of properties that have been added dynamically, then store that, but don't store their values as well because those values only represent the values that they have been initialized with, it doesn't include the values that they may have been changed to. So store them as `$this->extensionData['dynamicProperties'][] = $dynamicValue;` and access them as `$this->{$dynamicName}`

joakimbo: @LukeTowers Ok, will fix that. Just to be clear I had no problem accessing the properties by name, the problem arose when the Purgeable behavior needed access to all dynamicProperties without knowing their names.

src/Extension/ExtendableTrait.php
```diff
@@ -273,6 +276,11 @@ public function methodExists($name)
         );
     }
 
+    public function getExtensionData()
```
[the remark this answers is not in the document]
LukeTowers: Also, you could implement the value retrieval logic here as well:
```php
$result = [];
$properties = $this->extensionData['dynamicProperties'];
foreach ($properties as $prop) { 
     $return[$prop] = $this->{$prop};
}
return $result;
```

src/Database/Behaviors/Purgeable.php
```diff
@@ -0,0 +1,133 @@
...
+    public function purgeAttributes($attributesToPurge = null)
+    {
+        if ($attributesToPurge !== null) {
+            $purgeable = is_array($attributesToPurge) ? $attributesToPurge : [$attributesToPurge];
+        }
+        else {
+            $purgeable = $this->getPurgeableAttributes();
+        }
+
+        $dyn_prop_names = $this->parent->getExtensionData()['dynamicProperties'];
```
LukeTowers: Variables in October generally use camelCase, please change to `dynamicProperties` and just use the `$this->parent->getDynamicProperties()` method

src/Extension/ExtendableTrait.php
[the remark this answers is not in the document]
joakimbo: Ah, will fix

src/Database/Behaviors/Purgeable.php
```diff
@@ -0,0 +1,127 @@
...
+        if ($attributesToPurge !== null) {
+            $purgeable = is_array($attributesToPurge) ? $attributesToPurge : [$attributesToPurge];
+        }
+        else {
+            $purgeable = $this->getPurgeableAttributes();
+        }
+
+        $dynamicProperties = $this->parent->getDynamicProperties();
+
+        $attributes = array_merge($this->parent->getAttributes(), ($dynamicProperties));
```
LukeTowers: You can just move the call to `$this->parent->getDynamicProperties()` into the `array_merge`

src/Extension/ExtendableTrait.php
```diff
@@ -273,6 +276,17 @@ public function methodExists($name)
         );
     }
 
+    public function getDynamicProperties()
+    {
+        $result = [];
+        $propertyNames = $this->extensionData['dynamicProperties'];
+        foreach($propertyNames as $propName)
+        {
```
LukeTowers: opening brace on same line please for for loops

joakimbo:
@LukeTowers Would I assume correctly that you would want the docs under Services/Behaviors?

src/Extension/ExtendableTrait.php
```diff
@@ -273,6 +276,16 @@ public function methodExists($name)
         );
     }
 
+    public function getDynamicProperties()
```
LukeTowers: Still need a docblock for this, thanks!

joakimbo: Should be fixed now.

LukeTowers:
@joakimbo that would be correct, thanks!

src/Database/Behaviors/Purgeable.php
```diff
@@ -0,0 +1,125 @@
...
+     */
+    protected $originalPurgeableValues = [];
+
+    /**
+     * Boot the purgeable trait for a model.
+     * @return void
+     */
+    public function bootPurgeable()
+    {
+        if (!$this->parent->propertyExists('purgeable'))
```
LukeTowers: Just had an idea here, could you throw an exception if they don't include the `'purgeable'` entry in their `purgeable` property?

src/Extension/ExtendableTrait.php
```diff
@@ -273,6 +276,21 @@ public function methodExists($name)
         );
     }
 
+
+    /**
+     * Returns all dynamic properties
```
LukeTowers: "Returns all dynamic properties and their values"

src/Extension/ExtendableTrait.php
```diff
@@ -273,6 +276,21 @@ public function methodExists($name)
...
+     * @return array
```
LukeTowers: `@return array ['property' => 'value']`

src/Database/Behaviors/Purgeable.php
joakimbo: Sure, do you know on the top of your head why it tries to save it to the db? When just a property dynamically it threw no sql errors but as soon as the behavior is implemented it would.

LukeTowers: So you're saying that if you just `addDynamicProperty` it doesn't cause any issues but if you add the behavior itself as well it does cause issues?

joakimbo: Yup

joakimbo: Ah, @LukeTowers When defining the $purgeable property it can't be declared protected like in can for the trait, the behavior has no idea it's acting on behalf of a class.

joakimbo: I don't know a good reason why we would not be able to use a public var for this, even though I dont *like* it, it seems to be in line with how dynamic properties are declared?

joakimbo: Okay so implementing the behavior directly in a class seems to work when $purgeable is public, extending the class throws [here](https://github.com/octobercms/library/blob/0c45460e23f5b9916f4eecf4d8d3382d5cd03e15/src/Database/Model.php#L519)

LukeTowers:
@joakimbo so what's the state of this at the moment? Are you running into an issue that you can't solve?

joakimbo:
@LukeTowers The only problem remaining that I know of is that regardless of how the behavior is implemented (directly in a class or dynamically) any dynamically added arrays will [throw this exception](https://github.com/octobercms/library/blob/0c45460e23f5b9916f4eecf4d8d3382d5cd03e15/src/Database/Model.php#L519).
I don't yet understand why the arrays are not being caught there *without* the behavior but as soon as it's implemented they are.

joakimbo:
@LukeTowers If there isn't anything more from your side I *think* it's good to go.

joakimbo:
@LukeTowers Sorry for the spam and being "all over the place" with this, I've just had a hard time to wrap my head around everything but I think it's close now.
Dynamic properties seemed to be saved in the context of creating a new model but not when updating existing ones. I simply needed to add 'purgeable' as an entry directly for it to function like it normally does.

joakimbo:
@LukeTowers Adding the property if not existing could be done, however I have 2 questions about the other implementations.
1. Do we need to have the 'purgeable' entry in the actual $purgeable property? I have mixed feelings if that actually makes things more clear or confusing. Pushing it hiddes this fact and they dont have to think about it but in a way hides how the behavior needs to work.
2. Do we want to purge all dynamic properties? Are they not used for extending models?

LukeTowers:
@joakimbo I feel like it makes it more clear, I can't imagine a time where someone would actually want that property saved to the DB. 

Yes, you can extend models with dynamic properties, but when do those extensions line up with DB extensions? If you've already made changes to the DB for your extension then you don't need to add a dynamic property to the model in order to use that DB property you've added

joakimbo:
@LukeTowers If you mean the $purgeable property then I dont think so either, what I meant was that as of now the 'purgable' entry is added "temporarily" in the [purgeAttributes function](https://github.com/octobercms/library/pull/324/files#diff-f5a7e6239dd2da01646a3bc81c7c1b83R77) instead of being directly added to the property, this was if they read the property value if doesnt show up but is still functioning the same.

Ah okay, I see why adding the dynamic property keys make sense then.

LukeTowers:
@joakimbo it shouldn't, this only applies to models that implement this behavior. It doesn't remove any dynamic properties, just adds them to the purgeable array to prevent them from being saved into the DB

joakimbo:
@LukeTowers Does the latest commit satisfy? It should take into account any user defined entries and the dynamic properties

joakimbo:
@LukeTowers Should hopefully work as intended now.

LukeTowers:
@joakimbo are you able to add some unit testing for this functionality?

joakimbo:
@LukeTowers Not really familiar with it, but I think I could get something together though I'm not sure when. Should I just put it in another repo and link to it if I do?

LukeTowers:
@joakimbo it's already setup for this repo, see the tests/ directory. Take a look at the existing tests to get an idea for how to implement a test for this functionality. You'd want to test at the very least the trait being dynamically added vs added in the definition of the model class

joakimbo:
@LukeTowers Ok, I'll try and get on that as soon as I can.

joakimbo:
@LukeTowers I've added the test now, not sure it's what were looking for since I haven't done it before but hopefully it's not far off.

LukeTowers:
@joakimbo and have you run the test cases? Are there any issues? I haven't setup travis yet and I don't have time to pull this down to run them myself atm

joakimbo:
@LukeTowers Couldn't until today and after a bit of tinkering to get it to work I could run it. The first test code https://github.com/octobercms/library/pull/324/commits/0fc703e22fc332c73e5ed387614575813f6ef365 resulted in:
```
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.EF..                                                               5 / 5 (100%)

Time: 158 ms, Memory: 10.00MB

There was 1 error:

1) PurgeableTest::testDirectImplementationWithoutProperty
Error: Class 'TestModelWithoutProperty' not found

(pointer to PurgeableTest.php line 17)

--

There was 1 failure:

1) PurgeableTest::testDynamicImplementation
null does not match expected type "array".

(pointer to PurgeableTest.php line 32)

ERRORS!
Tests: 5, Assertions: 5, Errors: 1, Failures: 1.
```
The error was due to me having renamed a class but not the call to it, the failure is more interesting since it points out that any model created before being extended does not recieve extension? Could this ever happen?
The updated code https://github.com/octobercms/library/pull/324/commits/249660406f8927c29c52b653c0d2b4ac13331da9 resulted in:
```
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.....                                                               5 / 5 (100%)

Time: 162 ms, Memory: 10.00MB

OK (5 tests, 8 assertions)
```

joakimbo:
Updated the test case to separate tests and their classes a bit more than I had before, latest test passed.
```
c:\xampp\htdocs\octobercms\vendor\october\rain\tests>phpunit database/behaviors/purgeabletest.php
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.....                                                               5 / 5 (100%)

Time: 166 ms, Memory: 10.00MB

OK (5 tests, 8 assertions)
```

src/Database/Behaviors/Purgeable.php
```diff
@@ -0,0 +1,127 @@
...
+        {
+            $this->parent->addDynamicProperty('purgeable', []);
+        }
+        
+        $this->parent->purgeable[] = 'purgeable';
+        $dynPropNames = array_keys(array_diff_key($this->parent->getDynamicProperties(), ['purgeable'=> 0]));
```
LukeTowers: missing space between `'purgeable'` and `=>`

joakimbo:
@LukeTowers Anything more I can do to help this get closer to merge?

PR #325, opened 2018-06-07
Fixes section parser.

SeriousKen:
The section parser currently treats Markdown headers of the form ==== as section breaks, this is the cause of [an issue in October](https://github.com/octobercms/october/issues/3581). The parser is fairly niave but this fix makes the check for sections more strict so should cause less of a problem.


SeriousKen:
@LukeTowers I've added tests, should be all OK now

LukeTowers:
Awesome, thanks @SeriousKen!

PR #328, opened 2018-06-26
Enhance deferred_bindings table query performance

polly3d:
We had a manyToMany(belongsToMany) relation, e.g movies, movie_lists.
We used relation behavior to display the form.
There’re almost 1 millions record in movies table.
It costed 30 second when display create form page every time.
So we make a change to enhance the query performance.
Just change the exists query to in query.
It costed 3 second when display create form page.
It might be help to other guy.

### DIFF
--- a/src/Database/Relations/DeferOneOrMany.php
+++ b/src/Database/Relations/DeferOneOrMany.php
@@ -53,11 +53,10 @@ trait DeferOneOrMany
             /*
              * Bind (Add)
              */
-            $query = $query->orWhereExists(function($query) use ($sessionKey) {
+            $query = $query->orWhereIn($this->getWithDeferredQualifiedKeyName(), function($query) use ($sessionKey) {
                 $query
-                    ->select($this->parent->getConnection()->raw(1))
+                    ->select('slave_id')
                     ->from('deferred_bindings')
-                    ->whereRaw('slave_id = '.$this->getWithDeferredQualifiedKeyName())
                     ->where('master_field', $this->relationName)
                     ->where('master_type', get_class($this->parent))
                     ->where('session_key', $sessionKey)


LukeTowers:
@polly3d what's the difference in the generated SQL statements between the original approach and the new approach?

polly3d:
@LukeTowers Just changed the where exists to where in.

Here is the picture of performance and SQL.

Before:
before

After:
after

SQL before:
```sql
select
   `movies`.*,
   `movies`.* 
from
   `movies` 
where
   (
      exists 
      (
         select
            1 
         from
            `deferred_bindings` 
         where
            slave_id = movies.id 
            and `master_field` = 'movies' 
            and `master_type` = 'Company\\Pro\\Models\\Movielist' 
            and `session_key` = 'amSIR1Y1y79jj4KTrPspe2NXrbJ6pDyRyT37rMr5' 
            and `is_bind` = '1'
      )
   )
   and not exists 
   (
      select
         1 
      from
         `deferred_bindings` 
      where
         slave_id = movies.id 
         and `master_field` = 'movies' 
         and `master_type` = 'Company\\Pro\\Models\\Movielist' 
         and `session_key` = 'amSIR1Y1y79jj4KTrPspe2NXrbJ6pDyRyT37rMr5' 
         and `is_bind` = '0' 
         and id > ifnull((
         select
            max(id) 
         from
            deferred_bindings 
         where
            slave_id = movies.id 
            and master_field = 'movies' 
            and master_type = 'Company\\Pro\\Models\\Movielist' 
            and session_key = 'amSIR1Y1y79jj4KTrPspe2NXrbJ6pDyRyT37rMr5' 
            and is_bind = '1' ), 0)
   )
order by
   `id` desc limit 10 offset 0
```

SQL after:

```sql
select
   `movies`.*,
   `movies`.* 
from
   `movies` 
where
   (
      `movies`.`id` in 
      (
         select
            `slave_id` 
         from
            `deferred_bindings` 
         where
            `master_field` = 'movies' 
            and `master_type` = 'Company\\Pro\\Models\\Movielist' 
            and `session_key` = 'vF8oWoiJMxIwjblhNuGfjs9icGcfWzHFOpaVWi9F' 
            and `is_bind` = '1'
      )
   )
   and not exists 
   (
      select
         1 
      from
         `deferred_bindings` 
      where
         slave_id = movies.id 
         and `master_field` = 'movies' 
         and `master_type` = 'Company\\Pro\\Models\\Movielist' 
         and `session_key` = 'vF8oWoiJMxIwjblhNuGfjs9icGcfWzHFOpaVWi9F' 
         and `is_bind` = '0' 
         and id > ifnull((
         select
            max(id) 
         from
            deferred_bindings 
         where
            slave_id = movies.id 
            and master_field = 'movies' 
            and master_type = 'Company\\Pro\\Models\\Movielist' 
            and session_key = 'vF8oWoiJMxIwjblhNuGfjs9icGcfWzHFOpaVWi9F' 
            and is_bind = '1' ), 0)
   )
order by
   `id` desc limit 10 offset 0
```

Different:

```sql
where
   (
      exists 
      (
         select
            1 
         from
            `deferred_bindings` 
         where
            slave_id = movies.id 
            and `master_field` = 'movies' 
            and `master_type` = 'Company\\Pro\\Models\\Movielist' 
            and `session_key` = 'amSIR1Y1y79jj4KTrPspe2NXrbJ6pDyRyT37rMr5' 
            and `is_bind` = '1'
      )
   )
```

```sql
where
   (
      `movies`.`id` in 
      (
         select
            `slave_id` 
         from
            `deferred_bindings` 
         where
            `master_field` = 'movies' 
            and `master_type` = 'Company\\Pro\\Models\\Movielist' 
            and `session_key` = 'vF8oWoiJMxIwjblhNuGfjs9icGcfWzHFOpaVWi9F' 
            and `is_bind` = '1'
      )
   )
```

LukeTowers:
@polly3d awesome, thanks for that! Out of curiosity, why exactly is there such a large performance difference between doing `where exists` and `where in` in this case? Doing some mild googling on the topic seems to come up with the suggestion that they should take the same amount of time.

polly3d:
@LukeTowers 

All I know is:

- If the sub-query results are very large, then EXISTS will be faster than IN
- If the sub-query results are small, then IN will be faster than EXIST.

There is a large performance difference,  because `deferred_bindings` is small table, and our `deferred_bindings` table is almost always less than business table(e.g movies table)

LukeTowers:
Awesome, thanks for your contribution!

PR #330, opened 2018-06-28
Regenerate slugs if missing on existing models.

mrelevance:
This will allow slugs to be regenerated on save even if the model already exists so that if the slug attribute is missing it will be regenerated.

### DIFF
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -25,14 +25,10 @@ trait Sluggable
         }
 
         /*
-         * Set slugged attributes on new records
+         * Set slugged attributes on new records and existing records if slug is missing.
          */
         static::extend(function($model) {
             $model->bindEvent('model.saveInternal', function() use ($model) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Now what happens if a model with a slug set already is saved again? Wouldn't this cause it to regenerate the slug?

joakimbo:
@LukeTowers Would this not be gated here? https://github.com/octobercms/library/blob/6eff938b1d4d8c794893e5ab35d4f8a817bf627b/src/Database/Traits/Sluggable.php#L62

LukeTowers:
@joakimbo it would still run this: https://github.com/octobercms/library/blob/6eff938b1d4d8c794893e5ab35d4f8a817bf627b/src/Database/Traits/Sluggable.php#L80, although it would exclude itself here: https://github.com/octobercms/library/blob/6eff938b1d4d8c794893e5ab35d4f8a817bf627b/src/Database/Traits/Sluggable.php#L109-L111 so I suppose that's fine

PR #332, opened 2018-07-05
Fix halcyon dirty logic

SeriousKen:
Fix for issue https://github.com/octobercms/october/issues/3611

Takes into account empty values and leading/trailing whitespace in strings when determining if a Halcyon model is dirty and needs saving.

### DIFF
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -872,44 +872,18 @@ class Model extends Extendable implements ArrayAccess, Arrayable, Jsonable, Json
     /**
      * Get the attributes that have been changed since last sync.
      *
-     * @param  array|null  $attributes
-     * @param  array|null  $original
      * @return array
      */
-    public function getDirty($attributes = null, $original = null)
+    public function getDirty()
     {
         $dirty = [];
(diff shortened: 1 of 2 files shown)


SeriousKen:
Seems this breaks removing components, need to go away and re-think this one.

SeriousKen:
Working on this has shown up an issue in how the backend works and that missing attributes are ignored in determining if a model is dirty (octobercms/october#3628)

SeriousKen:
Now considers removed attributes when calculating the dirty attributes array. Please test with other plugins that rely on the core CMS objects such as Page and Content (eg Static Pages)

SeriousKen:
@LukeTowers A quick question, do you think purgable attributes be ignored when considering if a model is dirty? From what I understand, they are just temporary/calculated fields such as parsedMarkup but I don't know if there is any other purpose where they might be used.

LukeTowers:
In my opinion purgeable attributes shouldn't be considered as making the model dirty because they are never stored. Perhaps do some googling on the issue first though

bennothommo:

PR #335, opened 2018-07-13
Enable use of unencrypted cookies

pikanji:
Currently OctoberCMS encrypts all cookies, and thus it always tries to decrypt them on read.
This PR gives developers a chance to specify cookies that should not be decrypted.

Even though `\Illuminate\Cookie\Middleware\EncryptCookies` provides `$except` field and `disableFor` method to specify a list of cookies that shouldn't be decrypted, current implementation of october lib does not make use of them.

With this modification, developer will be able to specify unencrypted cookies in `config/cookie.php` like below.
```
<?php
return [
    "unencryptedCookies" => [
        "my_cookie",
    ],
];
```

This is useful when a developer wants to create cookies in frontend to pass data to backend.

### DIFF
--- a/src/Cookie/Middleware/EncryptCookies.php
+++ b/src/Cookie/Middleware/EncryptCookies.php
@@ -8,7 +8,7 @@ class EncryptCookies extends \Illuminate\Cookie\Middleware\EncryptCookies
     public function __construct(EncrypterContract $encrypter)
     {
         parent::__construct($encrypter);
-        $except = Config::get('cookie.unencryptedCookies');
+        $except = Config::get('cookie.unencryptedCookies', []);
         $this->disableFor($except);
     }
 }
--- a/src/Cookie/Middleware/EncryptCookies.php
+++ b/src/Cookie/Middleware/EncryptCookies.php
@@ -8,7 +8,7 @@ class EncryptCookies extends \Illuminate\Cookie\Middleware\EncryptCookies
     public function __construct(EncrypterContract $encrypter)
     {
         parent::__construct($encrypter);
-        $except = Config::get('cookie.unencrypted_cookies');
+        $except = Config::get('cookie.unecryptedCookies');
         $this->disableFor($except);
     }
 }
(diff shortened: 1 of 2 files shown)


LukeTowers:
@pikanji thanks for the submission! I'll need a PR to octobercms/october adding config/cookie.php with a detailed comment following the style of the other config files on how this works (using the key `unecryptedCookies`) and then if you could also add some information to octobercms/docs that would be perfect.

pikanji:
@LukeTowers thanks for your response!
I changed the key to `unecryptedCookies`, and added `config/cookie.php` and some information in docs.

src/Cookie/Middleware/EncryptCookies.php
```diff
@@ -8,7 +8,7 @@ class EncryptCookies extends \Illuminate\Cookie\Middleware\EncryptCookies
     public function __construct(EncrypterContract $encrypter)
     {
         parent::__construct($encrypter);
-        $except = Config::get('cookie.unencrypted_cookies');
+        $except = Config::get('cookie.unecryptedCookies');
```
LukeTowers: Typo: `unecryptedCookies` should be `unencryptedCookies` (typo was in my original comment)

pikanji:
@LukeTowers I addressed your point including ones in the other PRs related. thanks!

src/Cookie/Middleware/EncryptCookies.php
```diff
@@ -0,0 +1,14 @@
...
+
+use Config;
+use Illuminate\Contracts\Encryption\Encrypter as EncrypterContract;
+
+class EncryptCookies extends \Illuminate\Cookie\Middleware\EncryptCookies
+{
+    public function __construct(EncrypterContract $encrypter)
+    {
+        parent::__construct($encrypter);
+        $except = Config::get('cookie.unencryptedCookies');
```
LukeTowers: Please add a default value of an empty array here.

PR #336, opened 2018-08-15
Code cleanups

nathan-van-der-werf:
Just like the one in the octobercms/october repository (https://github.com/octobercms/october/pull/3690), a lot more can be done to cleanup the codebase, but this is a good starting point.

In this repository it would be also nice to have a static code analyser and code sniffer (see: https://github.com/octobercms/october/pull/3690#issuecomment-413305893).


daftspunk:
Thanks!

PR #338, opened 2018-08-17
Fixes wrong config item

SeriousKen:
`cms.pluginsDir` should be `cms.pluginsPath` otherwise `$` always resolves to `\plugin` even when a custom value is set in the `config/cms.php`


SeriousKen:
@LukeTowers sorry about the wrong branch, my mistake I was in a rush and didn't check. I don't understand your comment about separate items or why this wasn't merged. cms.pluginsDir isn't referenced anywhere in october/octobercms and only referenced once in october/library so I assume that it is incorrect in october/library.

LukeTowers:
My comment is that you should add it to the config/cms.php file. PLUGINSPATH is the url, pluginsdir is the directory, it's just not documented in the config file yet

PR #340, opened 2018-08-23
Corrects typo in ExtensionTrait

SeriousKen:
extensionIsHiddenField should be extensionIsHiddenMethod

### DIFF
--- a/src/Extension/ExtensionTrait.php
+++ b/src/Extension/ExtensionTrait.php
@@ -26,7 +26,7 @@ trait ExtensionTrait
 
     protected $extensionHidden = [
         'fields' => [],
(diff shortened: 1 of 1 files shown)


src/Extension/ExtensionTrait.php
```diff
@@ -26,7 +26,7 @@
 
     protected $extensionHidden = [
         'fields' => [],
-        'methods' => ['extensionIsHiddenField', 'extensionIsHiddenField']
+        'methods' => ['extensionIsHiddenMethod', 'extensionIsHiddenField']
```
LukeTowers: This is extremely pedantic I know but I would prefer that the order of these two are reversed because that's the order they appear in this file. :)

PR #341, opened 2018-08-23
Validation of encrypted values fails incorrectly

pmhou:
`getOriginalEncryptableValues()` called in the Validation trait will only return values that have passed through `makeEncryptableValue()` which in turn adds them to the `Encryptable::$originalEncryptableValues` array property . 

When updating an encrypted model's values, only the attributes passed in to update will run through `makeEncryptableValue` which will remove any other encrypted values on the following line:

`$encryptedAttributes = array_intersect_key($this->getOriginalEncryptableValues(), $data);`

**An example scenario illustrating the bug**

```php
// Model with encrytable & validation trait
public $rules = [
    'birthday' => 'required',
    'phone' => 'required',
];

protected $encryptable = [
    'birthday',
    'phone'
];

// Create instance
$model = new Model;
$model->birthday = '01/01/1970';
$model->phone = '0123456789';
$model->save();

// Update instance
$model->update(['phone' => '9876543210']);
// October/Rain/Database/ModelException with message 'The birthday field is required.'
```

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -217,7 +217,15 @@ trait Validation
              * only if they are part of the data being validated.
              */
             if (method_exists($this, 'getEncryptableAttributes')) {
-                $cleanAttributes = array_diff_key($data, array_flip($this->getEncryptableAttributes()));
+                $encryptableAttributes = $this->getEncryptableAttributes();
+                $cleanAttributes = array_diff_key($data, array_flip($encryptableAttributes));
+
+                foreach ($encryptableAttributes as $attribute) {
+                    if (isset($data[$attribute])) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@pmhou could you add a test case for this specific issue to prevent regressions? Also, does this cause any encrypted value to be seen as dirty after the validation is performed?

pmhou:
@LukeTowers How would you go about testing this? The actual piece of code is nestled deep within the validate method of the Validation trait and also relies on the Encryption trait. Thanks.

LukeTowers:
@pmhou I'd create a model stub that implements validation and encryption and then the test would just reference that model. 

Do you know if this change makes encrypted values always appear "dirty" (i.e. Laravel thinks they have been changed?)

bennothommo:

PR #342, opened 2018-08-25
Correcting file extension checking in Attach\File, adding the ability to upload a file by URL

ebashu-on-holidays:
`pathinfo() !== null` will always return `true`, because if no extension is specified, this method will return an empty string: `""`

This can be verified using the following code:
```php
$path_parts = pathinfo('/path/emptyextension.php', PATHINFO_EXTENSION);
var_dump($path_parts);

$path_parts = pathinfo('/path/noextension', PATHINFO_EXTENSION);
var_dump($path_parts);
```

Added the ability to upload files to the site by link.

To upload a file to a storage and attach to a model - you can use the following code:
```php
$file = new \System\Models\File;
$file->fromLink('https://..../file.jpg');
$user->avatar()->add($file);
```

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -2,6 +2,7 @@
 
 use Storage;
 use File as FileHelper;
+use October\Rain\Network\Http;
 use October\Rain\Database\Model;
 use October\Rain\Database\Attach\FileException;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
@@ -146,10 +147,10 @@ class File extends Model
 
     /**
      * Creates a file object from url
-     * @param $url string Filename
+     * @param $url string URL
      * @return $this
      */
-    public function fromLink($url)
+    public function fromUrl($url)
     {
         if ($url === null) {
             return;
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -152,10 +152,6 @@ class File extends Model
      */
     public function fromUrl($url)
     {
-        if ($url === null) {
-            return;
-        }
-
         $data = Http::get($url);
 
         if ($data->code != 200) {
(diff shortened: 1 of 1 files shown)


src/Database/Attach/File.php
```diff
@@ -144,6 +144,28 @@ public function fromData($data, $filename)
         return $file;
     }
 
+    /**
+     * Creates a file object from url
+     * @param $url string Filename
+     * @return $this
+     */
+    public function fromLink($url)
```
LukeTowers: method should be `fromUrl`, also `Http` needs to be imported at the top of the file.

ebashu-on-holidays:
@LukeTowers, I remembered that I have another small edit of this class:

```php
public function getWidthAttribute()
{
    if ($this->isImage()) {
        return getimagesize($this->getLocalPath())[0];
    }
}

public function getHeightAttribute()
{
    if ($this->isImage()) {
        return getimagesize($this->getLocalPath())[1];
    }
}
```

Can I add this code to this PR or do I need to create a new PR?

src/Database/Attach/File.php
```diff
@@ -144,6 +145,28 @@ public function fromData($data, $filename)
...
+     */
+    public function fromUrl($url)
+    {
+        if ($url === null) {
+            return;
+        }
+
+        $data = Http::get($url);
+
+        if ($data->code != 200) {
```
LukeTowers: I feel like a failed request should throw an exception because otherwise people are going to have to do something like
```php
$file->fromUrl('http://example.com/');
if (!$file) {
    // request went wrong
}
``` 
everywhere they use it

src/Database/Attach/File.php
```diff
@@ -144,6 +145,24 @@ public function fromData($data, $filename)
         return $file;
     }
 
+    /**
+     * Creates a file object from url
+     * @param $url string URL
+     * @return $this
+     */
+    public function fromUrl($url)
```
LukeTowers: Please also add a `$filename = false` optional parameter that can override the default filename generated from the URL basename. Don't forget to add it to the documentation as well.

PR #343, opened 2018-08-29
Prevent NestedTree generating onSave event for children

SeriousKen:
Reordering a NestedTree generates onSave events for child nodes but not on the node being moved which could cause performance issues or unintended consequences of the event being triggered.

This PR keeps the same functionality but doesn't generate the onSave event on child nodes.


LukeTowers:
@SeriousKen is this still relevant?

daftspunk:
Thanks!

PR #344, opened 2018-08-29
Fixes incorrect order on listsNested()

SeriousKen:
listsNested should return an indented array of nodes in the correct order to represent the tree but it actually returns the array in index order. This PR adds an order by clause to the query to fix this.


LukeTowers:
@SeriousKen is this still relevant?

bennothommo:
Closing this as there's not been any other reports of incorrect ordering (AFAIK) and a lot of work has been done to the codebase from when this PR was submitted.

PR #346, opened 2018-08-30
Fixes storeNewParent logic

SeriousKen:
Under some circumstances the model will think it needs to update its parent when it doesn't causing an exception. This fixes the logic to determine when the model needs to be moved in the tree.

See https://github.com/octobercms/october/issues/3740


daftspunk:
Thanks!

PR #347, opened 2018-09-05
Swoole support: add support in october rain library

tamer-hassan:
see: https://github.com/octobercms/october/issues/3746

depends on: https://github.com/octobercms/october/pull/3763

### DIFF
--- a/src/Database/DatabaseServiceProvider.php
+++ b/src/Database/DatabaseServiceProvider.php
@@ -60,10 +60,14 @@ class DatabaseServiceProvider extends DatabaseServiceProviderBase
             return new Dongle($this->getDefaultDatabaseDriver(), $app['db']);
         });
 
-        // Disable memory cache when running in console environment. This should
-        // prevent daemon processes from handling stale data in memory, however
-        // it should be kept active for the purpose of accurate unit testing.
-        if ($this->app->runningInConsole() && !$this->app->runningUnitTests()) {
+        /**
+         * Disable memory cache when running in console environment. This should
+         * prevent daemon processes from handling stale data in memory, however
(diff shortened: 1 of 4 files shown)


src/Database/DatabaseServiceProvider.php
```diff
@@ -60,10 +60,14 @@ public function register()
...
-        // Disable memory cache when running in console environment. This should
-        // prevent daemon processes from handling stale data in memory, however
-        // it should be kept active for the purpose of accurate unit testing.
-        if ($this->app->runningInConsole() && !$this->app->runningUnitTests()) {
+        /**
+         * Disable memory cache when running in console environment. This should
+         * prevent daemon processes from handling stale data in memory, however
+         * it should be kept active for the purpose of accurate unit testing.
+         *
+         * Also disable memory cache for Swoole http server.
```
LukeTowers: Why is the memory cache being disabled for the Swoole server? Is there no way to force a full cache clear at the end of every request so that we can retain the deduplication features of the memory cache?

tamer-hassan: There is a a way, and I thought it would cause unnecessary overhead.
Reverted disabling of DB Memory cache for Swoole and made default config in OctoberFly to flush the DB memory cache in every request:
https://github.com/tamer-hassan/october-fly/commit/9f14b7a11305d1883ae663eb400d8306e9585816
And the average request time is actually slightly less, according to my test especially on a complex page with lots of database data and many deep relations. So that's great! Thanks.

daftspunk:
This one is fixed by fd839047c2d7757f225a8ed60673ca8a20fe7df4

Thanks!

PR #339, opened 2018-08-22
Add relation.afterAttach, relation.afterDetach events in BelongsToMany relations

kharanenka:
Related [issue](https://github.com/octobercms/october/issues/3675).
I add new events for BelongsToMany relation
  * **relation.afterAttach**
  * **relation.afterDetach**
Usage example:
```php
Product::extend(function ($obModel) {
    $obModel->bindEvent('relation.afterAttach', function ($sRelationName, $arElementIDList, $arInsertData) {
        //$sRelationName == 'category'
        //$arElementIDList == [1,3,15]
        //$arInsertData = [['product_id' => 10, 'category_id' => 1],['product_id' => 10, 'category_id' => 3],['product_id' => 10, 'category_id' => 15, 'pivot_field' => 'test']];
    });
    $obModel->bindEvent('relation.afterDetach', function ($sRelationName, $arElementIDList) {
        //$sRelationName == 'category'
        //$arElementIDList == [1,3,15]
    });
});
```

PR #349, opened 2018-09-12
Get attache ID list, if detach() method was called without parameters

kharanenka:


kharanenka:
Related [PR](https://github.com/octobercms/library/pull/339)

src/Database/Relations/BelongsToMany.php
```diff
@@ -163,6 +163,9 @@ public function attach($id, array $attributes = [], $touch = true)
     public function detach($ids = null, $touch = true)
     {
         $attachedIdList = $this->parseIds($ids);
+        if (empty($attachedIdList)) {
+            $attachedIdList = $query = $this->newPivotQuery()->lists($this->relatedPivotKey);
```
LukeTowers: Why do you also assign it to `$query` as well?

kharanenka:
Sorry, I corrected this stupid mistake

PR #350, opened 2018-09-19
Add theme and plugin url helper

tschallacka:
As requested by @LukeTowers on https://github.com/octobercms/october/pull/3794#issuecomment-422817893 a few new helper functions.

`themes_url()` will point to the base themes directory

`theme_url()` will point to the current active theme directory

`plugin_url()` will point to the current calling plugin directory. When called from class that's not actually part of a plugin things will go awry but let's assume people will mainly call it either in co-junction with a plugin class like a model or component or controller, or from those. But if someone wishes to use a screwdriver to hammer in a nail, they're welcome to try.

All these functions can be called with a path argument to a file.

I chose fort he debug_backtrace support in plugin_url because the cost is neglible and allows for easy use.

### DIFF
--- a/src/Support/helpers.php
+++ b/src/Support/helpers.php
@@ -247,7 +247,7 @@ if (!function_exists('plugins_url'))
      */
     function plugins_url($path = '')
     {
-        $base = implode('/', [config('app.url'), config('cms.pluginsPath'));
+        $base = implode('/', [config('app.url'), config('cms.pluginsPath')]);
         
         return url($base . ($path ? '/'.$path : $path));
     }
--- a/src/Support/helpers.php
+++ b/src/Support/helpers.php
@@ -211,8 +211,8 @@ if (!function_exists('themes_url'))
     /**
      * Get the public url to the themes folder.
      *
-     * @param  string  $path
-     * @return string
+     * @param  string  $path A relative path to a file in the themes directory
+     * @return string the full public url(http://example.com/themes/theme/file.txt) to the file in the themes directory.
      */
     function themes_url($path = '')
     {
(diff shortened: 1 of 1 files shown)


LukeTowers:
While I appreciate the ingenuity I'd prefer that we only supported `plugins_url()` and not `plugin_url()`. There's just too much that can potentially go wrong with not enough benefit, I can't think of a single case where I need the URL to the current plugin but I also don't know what that current plugin is to use with plugins_url(). Unless you can think of a good example, I'd prefer that we drop that and only use `plugins_url`, `themes_url`, and `theme_url`.

tschallacka:
I changed the plugins_url, to your satisfaction now?

LukeTowers:
@tschallacka do your tests covering each of those config items both starting and ending with a slash?

tschallacka:
@LukeTowers it does now.

LukeTowers:
@tschallacka Ping me if I forget to get to this, I'll have to spin up an octodock instance to run the unit tests since cross repo changes don't play nice with our travis setup

tschallacka:
I get what you mean. I just ended up copying this file in a dummy repo to be able to test it.

Teranode:
lemme uhhh, ping you there @LukeTowers

LukeTowers:
@tschallacka @Teranode have you both tested these thoroughly?

tschallacka:
> have you both tested these thoroughly?

I read that as: it's really a bother to set up a testing enviroment for cross repository tests.

I can see about doing a re-test but last time I checked everything was in order

tschallacka:
@LukeTowers @Teranode  I built up the following commands for easy testing reproduction

```
git clone https://github.com/octobercms/october.git
cd october
git fetch origin pull/3804/head:itscomplicated
#                     ^^^^ pull request number here
git checkout itscomplicated
composer install
cd vendor/october/rain
rm -rf *
rm .gitignore
git clone https://github.com/octobercms/library.git .
git fetch origin pull/350/head:itscomplicatedtoo
#                     ^^^ pull request number here
git checkout itscomplicatedtoo
cd ../../..
cd tests/unit
phpunit
```

but for the love of me, if I run it like that I get all kinds of double slash problems. I need to dig a bit in it if it's an encoding issue or that I switched to php 7.2 or something else why I suddenly get unexpected double slashes.

LukeTowers:

PR #353, opened 2018-10-11
Update MediaLibrary reference

acasar:
My log is full of deprecation warnings:

image

This is the only reference to the Cms\Classes\MediaLibrary I could find, so I suspect this is the reason.


LukeTowers:
@acasar good catch! Thanks very much :)

PR #358, opened 2018-11-21
In Mailer::queue is callback optional

Rike-cz:
In Mailer::queue is $callback optional (`public function queue($view, $data = null, $callback = null, $queue = null)...`) so `buildQueueMailable` have to reflect this. And as well `send` method.


LukeTowers:
@Rike-cz thanks!

PR #361, opened 2018-11-26
Update StylesheetMinify

Samuell1:
Hey,

I updated a minify method that handles minification of CSS. This should help a lot with performance and CSS file sizes. If anybody have another ideas to improve this just comment this pr.

### DIFF
--- a/src/Parse/Assetic/StylesheetMinify.php
+++ b/src/Parse/Assetic/StylesheetMinify.php
@@ -26,16 +26,39 @@ class StylesheetMinify implements FilterInterface
      */
     protected function minify($css)
     {
-        $css = preg_replace('#\s+#', ' ', $css);
+        // Normalize whitespace
+        $css = preg_replace('/\s+/', ' ', $css);
+
+        // Remove spaces before and after comment
+        $css = preg_replace('/(\s+)(\/\*(.*?)\*\/)(\s+)/', '$2', $css);
+
+        // Remove comment blocks, everything between /* and */
         $css = preg_replace('#/\*.*?\*/#s', '', $css);
-        $css = str_replace('; ', ';', $css);
-        $css = str_replace(': ', ':', $css);
-        $css = str_replace(' {', '{', $css);
-        $css = str_replace('{ ', '{', $css);
-        $css = str_replace(', ', ',', $css);
-        $css = str_replace('} ', '}', $css);
-        $css = str_replace(';}', '}', $css);
-        $css = str_replace('}', '}'.PHP_EOL, $css);
+
+        // Remove ; before }
+        $css = preg_replace('/;(?=\s*})/', '', $css);
+
+        // Remove space after , : ; { } */ >
+        $css = preg_replace('/(,|:|;|\{|}|\*\/|>) /', '$1', $css);
+
+        // Remove space before , ; { } ( ) >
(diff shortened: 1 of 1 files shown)


w20k:
@Samuell1 Could you run it against css library, like bootstrap, and see if minification didn't break anything?

Samuell1:
@w20k I can try with boostrap, i tried it with boostrap grid only and looks like everything works fine.

w20k:
Could you @Samuell1  also double check, if it's not used for 'scss or sass' files?

Samuell1:
@w20k Shouldn't be minify used only for CSS files?

w20k:
Sorry, @Samuell1 mislead you a bit 😄Had an idea.

w20k:
Testing finished. No issues fond. Want to re-run it once again on Monday, and ping @LukeTowers.

LukeTowers:
Sounds good, let me know on Monday @w20k

w20k:
@LukeTowers, PR will be postponed a bit!

@Samuell1 , could you fix the issue with brackets? Added files, and screenshots.

Using demo theme.
With patch:
screenshot 2018-12-10 at 15 57 40

Current dev:
screenshot 2018-12-10 at 15 58 57


File to compare:
[withoutPR.css.txt](https://github.com/octobercms/library/files/2663384/withoutPR.css.txt)
[withPR.css.txt](https://github.com/octobercms/library/files/2663385/withPR.css.txt)

w20k:
Looks good with demo theme. Need to make one more test with bootstrap samples and would be ready to merge @LukeTowers 😉

w20k:
@LukeTowers,  Testing Status - ✅
Looks great with the latest bootstrap.

LukeTowers:
@Samuell1 how much does removing newlines affect the final size / performance? I would prefer we kept them because it's much easier to diff compiled files when newlines exist which is better for maintainers and better for security (harder to slip malicious code into compiled files when diffs are easier to check)

Samuell1:
@LukeTowers I think its a lot of difference in size and maybe parse time for browser can be faster. But can css have milicious code? Shouldnt be css recompiled before releasing, testing it? to prevent this things?

mjauvin:
I would tend to agree with @samuell1 on this. I think it would be worth it.
Can you offer a size comparison?

--
Marc

LukeTowers:
Well a single CRLF byte sequence is two bytes, so depending on the newlines being used it's either going to be a saving of 1 or 2 bytes per line. I suppose if you have massive CSS files you could save 1KB per 1,000 lines, but for the largest CSS file October has (storm.css, 3,000 lines) that's a saving of 3KB for a dramatic reduction in diff checking abilities.

CSS by itself is tricky to do anything malicious in, but it is possible to do some evil things (see https://css-tricks.com/css-keylogger/). Asides from the security concerns around not being able to clearly see what has been changed in minified stylesheets there's also the concern about changes just being incorrect, for example if someone just manually edited the minified file instead of running the asset compiler (for any number of reasons) and there was a minor bug in their changes.

Thoughts?

LukeTowers:
Additionally, there might? be some concerns with large file content all being on one line: https://stackoverflow.com/a/10194153/6652884, just something else to think about.

w20k:
Maybe by default we could leave the newlines be? And, for example, if you add --min or --full param it wil remove those.

PR #362, opened 2018-11-30
Allow optional wildcard parameters

mariavilaro:
A wildcard parameter can be made optional using the syntax `url = "/blog/:slug?*"`

So this would work:

url/blog
url/blog/category
url/blog/category/subcategory

For using a default value the syntax is `url="/blog/:slug?all*"`

Note: this was already the behaviour when passing a param with ?*, but the parameter was parsed like this:

param name: slug?
value: *

(the optional character wasn't stripped from the parameter name, and the wildcard character was returned as the parameter value)

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -510,7 +510,7 @@ class Manager
 
         $this->user = null;
 
-        Session::forget($this->sessionKey);
+        Session::flush();
         Cookie::queue(Cookie::forget($this->sessionKey));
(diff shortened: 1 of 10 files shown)


mariavilaro:
sorry i made the pull request in master, closing and reopening

LukeTowers:
@mariavilaro you can always just change the base branch to develop instead by clicking on the edit button next to the title for future reference

PR #363, opened 2018-11-30
Allow optional wildcard parameteres

mariavilaro:
A wildcard parameter can be made optional using the syntax `url = "/blog/:slug?*"`

So this would work:

url/blog
url/blog/category
url/blog/category/subcategory

For using a default value the syntax is `url="/blog/:slug?all*"`

Note: this was already the behaviour when passing a param with ?*, but the parameter was parsed like this:

param name: slug?
value: *

(the optional character wasn't stripped from the parameter name, and the wildcard character was returned as the parameter value)

### DIFF
--- a/src/Router/Helper.php
+++ b/src/Router/Helper.php
@@ -147,6 +147,10 @@ class Helper
         $regexMarkerPos = mb_strpos($name, '|');
 
         if ($wildMarkerPos !== false) {
+            if ($optMarkerPos !== false) {
+                return mb_substr($name, 0, $optMarkerPos);
+            }
+
             return mb_substr($name, 0, $wildMarkerPos);
         }
 
@@ -201,10 +205,14 @@ class Helper
         }
 
         $regexMarkerPos = mb_strpos($segment, '|');
+        $wildMarkerPos = mb_strpos($segment, '*');
         $value = false;
 
         if ($regexMarkerPos !== false) {
             $value = mb_substr($segment, $optMarkerPos+1, $regexMarkerPos-$optMarkerPos-1);
+        }
+        elseif ($wildMarkerPos !== false) {
+            $value = mb_substr($segment, $optMarkerPos+1, $wildMarkerPos-$optMarkerPos-1);
         }
         else {
             $value = mb_substr($segment, $optMarkerPos+1);


LukeTowers:
@mariavilaro could you add unit tests to this for the cases that you found it exhibiting unintentional behaviour? Also, do the docs need to be updated for this?

mariavilaro:
Hi! yes I was not able to run the tests in my computer, I will try again now and tell you something

mariavilaro:
@LukeTowers "could you add unit tests to this for the cases that you found it exhibiting unintentional behaviour?" what do you exactly mean with this?

mariavilaro:
No luck, this is the error I get when I try to run `phpunit tests` inside octobercms library

`C:\repos\octobercms-library>phpunit tests`

`Fatal error: Class 'TestCase' not found in C:\repos\octobercms-library\tests\Config\ConfigWriterTest.php on line 5`

composer install ran perfectly and vendor folder is correctly populated, any clue about what is happening?

w20k:
@mariavilaro could you check if this will work. Try changing from `TestCase to PHPUnit_Framework_TestCase` (C:\repos\octobercms-library\tests\Config\ConfigWriterTest.php).

w20k:
@mariavilaro could you try in TestCase.php -> change extend to `\PHPUnit\Framework\TestCase or PHPUnit_Framework_TestCase` and before re-running do composer update for the folder 👍

mariavilaro:
I just added the test to october repo https://github.com/octobercms/october/pull/3964

mariavilaro:
> 
> 
> @mariavilaro could you try in TestCase.php -> change extend to `\PHPUnit\Framework\TestCase or PHPUnit_Framework_TestCase` and before re-running do composer update for the folder 👍

No luck, same error

mariavilaro:
Hi @LukeTowers I added the required unit tests in october main repo, for the current behaviour and the behaviour with this fix applied https://github.com/octobercms/october/pull/3964

LukeTowers:
@mariavilaro thanks! @w20k could you test these?

w20k:
@LukeTowers, I hope will today!

w20k:
@mariavilaro, I've found a bug or might be a 'feature' to fix 😄 

If you use url pattern, which you've suggested = `url="/blog/:slug?all*"`:
Your slug param would be a string, I'd guess it should be an array. Each slug should be an item in an array. 

Screenshot:
screenshot 2018-12-17 at 17 57 07

mariavilaro:
Right now a string is returned when there is a wildcard parameter in the url. I don't think we should change this behaviour, it'd be a breaking change.

w20k:
@LukeTowers this one is re-tested. 
Didn't find any issues. Could be merged!

Status - ✅ 
Side Note: Tests (https://github.com/octobercms/october/pull/3964) should be merged only after this PR.

PR #364, opened 2018-12-07
HasMany memory performance solution - pluck on QueryBuilder instead o…

viamage:
…f Collection

Creating collection here causes memory issues. More details at https://github.com/octobercms/october/issues/3969


w20k:
Tested works fine 😉

PR #365, opened 2018-12-10
Allow validator rule objects

Flynsarmy:
This PR adds support for object instances in rule definitions:
https://laravel.com/docs/5.5/validation#using-rule-objects

This functionality would be useful because it provides a way of extending the validator in a non-global manner.

### Example Usage
In a model add:
```php
public function beforeValidate()
{
    $this->rules['myField'][] = new \Path\To\ClassExists;
}
```
With example class definition:
```php
<?php

namespace Path\To;

use Illuminate\Contracts\Validation\Rule;

class ClassExists implements Rule
{
    public function passes($attribute, $value)
    {
        return class_exists($value);
    }

    public function message()
    {
        return 'The :attribute class does not exist.';
    }
}
```

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -369,6 +369,12 @@ trait Validation
              */
             foreach ($ruleParts as $key => $rulePart) {
                 /*
+                 * Allow rule objects
+                 */
+                if (is_object($rulePart)) {
+                    continue;
+                }
+                /*
                  * Remove primary key unique validation rule if the model already exists
                  */
                 if (starts_with($rulePart, 'unique') && $this->exists) {


Flynsarmy:
While this PR works as advertised, it would be a nice addition to provide the field name somehow. In my example above the string _'The :attribute class does not exist.'_ evaluates to:

> The settings.non channel relation class does not exist.

when it should say

> The Model class does not exist 

with a form field definition of:
```yaml
settings[nonChannelRelation]:
    label: Model
    type: text
```

I'm not sure how to make a change like this myself however.

LukeTowers:
@Flynsarmy are you using the latest develop branch? I'm sure I added support for the formcontroller inserting labels for attribute names in validation messages

Flynsarmy:
Just the live build. not sure how to use dev branch.

Flynsarmy:
On develop branch now. With _settings.nonChannelRelation_ added to the _attributeNames_ array, the issue is still occurring.

alpapan:
Hello Flynsarmy

When I use 
```
$this->rules['myField'][] = new \Path\To\ClassExists;
```

Then I get 
image

While if it is
```
 $this->rules['myField'] = new \Path\To\ClassExists;
```
then it complaints it cannot explode the object... is this doc only valid for the dev branch?

alpapan:
Turns out you need this in the Rule class (from https://github.com/mpociot/laravel-apidoc-generator/issues/247#issuecomment-353553829).....

```
public function __toString()
{
  return 'Name of Rule';
}
```

LukeTowers:
Should we also include an abstract base class RuleBase for developers to extend from that handles the __toString()?

bennothommo:
Work on this feature is now being done in https://github.com/octobercms/library/pull/424 as this appears to be abandoned.

Flynsarmy:
This PR is identical to https://github.com/octobercms/library/pull/424 and yet mine gets closed?

PR #371, opened 2019-01-11
Added validateUserModel method

tobias-kuendig:
Removed dependency on `SoftDelete` trait and added `validateUserModel` method to provide the ability to reject a user model.

see https://github.com/octobercms/october/issues/4046 and https://github.com/octobercms/october/pull/4051

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -89,6 +89,7 @@ class Manager
         $model = $this->createUserModel();
         $query = $model->newQuery();
         $this->extendUserQuery($query);
+
         return $query;
     }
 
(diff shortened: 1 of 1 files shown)


src/Auth/Manager.php
```diff
@@ -238,6 +242,18 @@ public function findUserByCredentials(array $credentials)
         return $user;
     }
 
+    /**
+     * Perform additional checks on the user model.
+     *
+     * @param $user
+     *
+     * @return boolean
```
LukeTowers: please get rid of the extra line between `@param` and `@return`

LukeTowers:
@tobias-kuendig no need to force push commits generally btw, I always squash and merge PRs (simple ones at least) into one commit to keep the history clean.

PR #376, opened 2019-01-28
Fix hasMany where not using primary key as "key"

nathan-van-der-werf:
Pull request #364 created a bug in the `HasMany` relation.

Create a model, for example: `User`
- id
- email

Create a model, for example: `File`
- id
- sender

Have an `HasMany` relation in the `User` model:
```
public $hasMany = [
    'files' => [File::class, 'key' => 'sender', 'otherKey' => 'email'],
];
```

Render the partial in the backend:

fields.yaml
```
tabs:
    fields:
        files:
            type: partial
            tab: Files
            context: update
```

config_relation.yaml
```
files:
    label: Files
    view:
        list: ~/plugins/acme/base/models/files/columns.yaml
        recordUrl: acme/base/files/update/:id
    emptyMessage: backend::lang.list.no_records
```

_files.htm
```
<?= $this->relationRender('files') ?>
```

The query generated before the fix is:
```
select `email` from `files` where `files`.`sender` = example@example.com and `files`.`sender` is not null
```

Where `email` is not existing in the `files` table... (resulting in a column not found exception)

Before pull request #364 the select was an asterix (*), the query wouldn't fail because of that. Instead of the `localKey` the pluck should take the `foreignKey`, this will generate the following query:

```
select `files`.`sender` from `files` where `files`.`sender` = 'example@example.com' and `files`.`sender` is not null
```

### DIFF
--- a/src/Database/Relations/HasMany.php
+++ b/src/Database/Relations/HasMany.php
@@ -81,7 +81,7 @@ class HasMany extends HasManyBase
         $relationName = $this->relationName;
 
         if ($relation = $this->parent->$relationName()) {
-            $value = $relation->pluck($this->localKey)->all();
+            $value = $relation->pluck($this->foreignKey)->all();
         }
 
         return $value;


mjauvin:
I'm confused... the following commit did sort of the reverse for HasOne relation:

   https://github.com/octobercms/library/commit/e502f3371e5f852989f4cccbcfdd10d67b54f49a

nathan-van-der-werf:
@mjauvin That one is using:

```
if ($this->parent->{$relationName}) {
```

Instead of:
```
if ($relation = $this->parent->$relationName()) {
```

Note the `()`. (Which might be the difference);

mjauvin:
Can you try with localKey but without the "()" to see if that really makes a difference??

nathan-van-der-werf:
> Can you try with localKey but without the "()" to see if that really makes a difference??

I tried this, this works since it reverts #364 , the only issues with this is that it will use an * in the select query, which is not good for the performance.

LukeTowers:
@nathan-van-der-werf can you submit the query that existed before the previous PR?

LukeTowers:
Also @viamage can you weigh in on this?

viamage:
Will take a look shortly and try to get my team's feedback. Test cases for that method would really help, quite hard to design I guess (especially as a lot of issues and possible issues are around efficiency here), but it's used in so many contexts that I'm afraid manual testing will never be enough here and it will return sooner or later after each change. 

@nathan-van-der-werf to not waste time, did you by any chance tested your PR with https://github.com/octoberrain/test-plugin and/or https://github.com/viamage/relationsbug ?

robinbonnes:
Having the same issue, due to the PR #364 my plugin breaks.

This: 

```
        if ($relation = $this->parent->$relationName) {
            $value = $relation->pluck($this->localKey)->all();
        }
```

Results in:

```
select * from table where table.key = '....'
```

And @nathan-van-der-werf solution:

```
        if ($relation = $this->parent->$relationName()) {
            $value = $relation->pluck($this->foreignKey)->all();
        }
```

Results in:

```
select table.foreignKey from table where table.key = '....'
```

Would be great if we can merge this in, because it causes a lot of issues for some plugins.

w20k:
@viamage @nathan-van-der-werf will try to test those! That might speed up the process a bit 😄

viamage:
Played a bit with it, fix seems to be working good.

viamage:
Definetely can confirm that this PR broke custom keys configs, fix worked on various apps I managed to pull up to test it and no pre-PR efficiency issues were reproduced, please kindly merge as soon as asignee @w20k confirms it's fine.

w20k:
@viamage will finish testing today!

LukeTowers:
@w20k just let me know when you're ready

w20k:
@LukeTowers done with testing, fix works perfectly!
Status - ✅

LukeTowers:
@nathan-van-der-werf Could you look into why https://github.com/octobercms/october/pull/4191 is failing now? I'll have to revert this if we can't fix the test soon.

w20k:
Hi, @nathan-van-der-werf @viamage - I'll revert your changes (https://github.com/octobercms/library/pull/376 and https://github.com/octobercms/library/pull/364). 

To reapply your PRs, please consider finding a proper solution for this failing test (for now it's skipped): https://github.com/octobercms/october/blob/master/tests/unit/plugins/database/HasManyModelTest.php#L64 ?

HammenWS:
@LukeTowers Looks like this PR is not reverted yet. And it breaks the relation form widget for a hasMany relation. Please don't ignore existing tests.

LukeTowers:
@w20k @bennothommo could you look into this?

bennothommo:
@LukeTowers there's a PR from @w20k [here](https://github.com/octobercms/library/pull/387) to revert the changes, but it looks to have been closed without being merged. @w20k, was there a reason it was closed?

HammenWS:
@LukeTowers @bennothommo I would love to make a PR for to revert the changes, but I don't have permission to push to this repository. Got it ready locally. 

When it is reverted, the test in octobercms https://github.com/octobercms/october/blob/7590dfd30fb08242c6c8e111115659d38ac4aeb8/tests/unit/plugins/database/HasManyModelTest.php#L64 can be unignored again (wasn't that scary to ignore such a test?).

I hope these commits can be reverted soon. A lot of relations for the site of my client are already broken, for using a core octobercms form widget. Yikes.

Kind Regards,
Ramiro Hammen

bennothommo:
@HammenWS If you want to make a PR, you are able to fork this repo using the Fork button at the top of the page, and then make a branch on your own copy of the repo. You can then create a PR to this repo using your branch. This page should have more info: https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

In this instance, however, we can restore @w20k's branch and merge his changes in if necessary.

w20k:
@bennothommo could you look at my PR quickly and we could merge. I think, did a small GitHub cleanup and removed 😄

HammenWS:
@w20k @LukeTowers Thanks! When can I expect a new release?

LukeTowers:
Updating the composer packages now, new build won't be out for a while though. Just use the develop dependencies through composer for now, or manually edit it if you want as it'll be in the next release

PR #377, opened 2019-02-12
Remove null values from event listener responses

SeriousKen:
No longer add null values from event listeners to the response list  https://github.com/octobercms/october/issues/4089

### DIFF
--- a/src/Events/Dispatcher.php
+++ b/src/Events/Dispatcher.php
@@ -232,10 +232,15 @@ class Dispatcher implements DispatcherContract
         foreach ($this->getListeners($event) as $listener) {
             $response = call_user_func_array($listener, $payload);
 
-            // If a response is returned from the listener and event halting is enabled
-            // we will just return this response, and not call the rest of the event
-            // listeners. Otherwise we will add the response on the response list.
-            if (! is_null($response) && $halt) {
+            // If the response is null continue to the next listener.
+            if ($response === null) {
+                continue;
+            }
+
+            // If event halting is enabled we will just return this response,
+            // and not call the rest of the event listeners. Otherwise we will
+            // add the response on the response list.
+            if ($halt) {
                 array_pop($this->firing);
 
                 return $response;


w20k:
@LukeTowers could you look-over?
For me and @tobias-kuendig - it's working! 

Status - ✅

mjauvin:
I'm sure it solves this particular issue, but I think we need to look in all the places handling events results to make sure the "null" value that used to be returned is not expected in somewhere else.

I think the fix in PR #4105 would have less risk to introduce a regression.

mjauvin:
@LukeTowers looks like @tobias-kuendig tested this (ref. #4105) and it's working great.

probably should be merged.

bennothommo:
I'm a little wary of this change - there are valid reasons for people to want to return `null` from a listener. This change will strip out any `null` values irrespective of whether it's from a listener that returns `null` or one that returns nothing at all.

I would be more inclined to accept a merge if this change was optional - for example, a fourth parameter to the dispatch method called `$ignoreNull`, defaulting to false to maintain the current dispatch behaviour.

SeriousKen:
@bennothommo I disagree, the logic in the dispatcher is working differently for local vs global events, ie one adds them to the list and the other doesn't. This is unexpected and inconsistent behaviour. All event dispatchers should be predictable and consistent and all bar this one disguard null values.

bennothommo:
@SeriousKen As much as there is a discrepancy (and I agree they should really have the same behaviour), this change will break compatibility with event dispatchers that are out there now that were either relying on, or working around, the fact that `null` values might be present in the resulting array.

SeriousKen:
@bennothommo in its current state it's already breaking event handling that expects there to be no nulls in the response (see Clockwork plugin which took me hours of development to discover it was the problem). There is also no difference between a handler that returns null or returns early because it should not be fired due to some condition. The correct behaviour for an event handler is to return a non null value for any special situation that needs handling outside the handler or null if it is to be ignored.

mjauvin:
I agree with both sides here, but I think @daftspunk's opinion on this would be really helpful.

bennothommo:
@SeriousKen what functionality in particular is it breaking? If it's a plugin that is dispatching the event and failing because a `null` is returned, I think that the plugin developer would need to resolve that within the plugin.

SeriousKen:
@bennothommo it is broken because the implementation in October is broken, it causes an exception in October's code when using 'depends on' in a backend form, there is nothing wrong with the plugin.

bennothommo:
@SeriousKen There is no difference in code between October's global event listener handling and Laravel 5.5's global event listener handling.

Compare:
https://github.com/octobercms/library/blob/develop/src/Events/Dispatcher.php#L232-L251
with:
https://github.com/laravel/framework/blob/5.5/src/Illuminate/Events/Dispatcher.php#L199-L216

Which leads me to believe that the same behaviour happens in Laravel too. It is only the local model events that differ with `October\Rain\Support\Traits\Emitter` being a different implementation to `Illuminate\Database\Eloquent\Concerns\HasEvents`. So I would be more inclined to say that it's the local events that are behaving wrong, however, it falls to the same issue as before where changing it now may break current implementations.

If `dependsOn` is breaking due to this, would I be able to suggest instead finding out if maybe `dependsOn` needs to be fixed to ignore `null` values? If it's simply impossible without this PR, then we will consider deprecating the behaviour.

SeriousKen:
@bennothommo good catch, seems I was wrong. I'll have another look at where the issue can be fixed.

SeriousKen:
The offending code is in the Form widget

https://github.com/octobercms/october/blob/4e92686c1a7de662840fa160b7348003f5ba817d/modules/backend/widgets/Form.php#L449-L453

It fails when the response is not a string, which is the case when Clockwork's wildcard listener is fired. Adding a check should solve this.

bennothommo:
@SeriousKen Thanks heaps for looking into that. Would you mind filling out a PR or issue in regards to fixing that in the [October repo](https://github.com/octobercms/october)? :)

SeriousKen:
@bennothommo I've submitted https://github.com/octobercms/october/pull/4331, this should sort this problem out with no BC issues.

bennothommo:
@SeriousKen awesome, thanks a lot! :)

PR #378, opened 2019-02-14
reduced sluggable max length to 175 and fix incorrect work with russi…

Nai4rus:
https://github.com/octobercms/october/issues/4133

### DIFF
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -53,9 +53,9 @@ trait Sluggable
      * @param int $maxLength Maximum length for the slug not including the counter.
      * @return string The generated value.
      */
-    public function setSluggedValue($slugAttribute, $sourceAttributes, $maxLength = 240)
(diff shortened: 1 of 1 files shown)


src/Database/Traits/Sluggable.php
```diff
@@ -53,9 +53,9 @@ public function slugAttributes()
      * @param int $maxLength Maximum length for the slug not including the counter.
      * @return string The generated value.
      */
-    public function setSluggedValue($slugAttribute, $sourceAttributes, $maxLength = 240)
+    public function setSluggedValue($slugAttribute, $sourceAttributes, $maxLength = 175)
```
LukeTowers: Why does the max length have to be 175 instead of 190? @Nai4rus

Nai4rus: @LukeTowers for counter -> `$maxLength Maximum length for the slug not including the counter.`
I think because of this, the old value was 240, not 255.

LukeTowers: Good point

PR #380, opened 2019-02-27
Change POSTFIELDS back to array

alxy:
In a previous commit, this has been changed to a string to ensure the correct argument separator to be set. However, starting from PHP 5.2.0, this disables us from uploading files, as for files an array argument is required. From the docs:
``If value is an array, the Content-Type header will be set to multipart/form-data. As of PHP 5.2.0, value must be an array if files are passed to this option with the @ prefix. As of PHP 5.5.0, the @ prefix is deprecated and files can be sent using CURLFile.
``

See: http://de2.php.net/manual/en/function.curl-setopt.php

### DIFF
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -279,7 +279,7 @@ class Http
             $requestDataQuery = http_build_query($this->requestData, '', $this->argumentSeparator);
 
             if (in_array($this->method, [self::METHOD_POST, self::METHOD_PATCH, self::METHOD_PUT])) {
-                curl_setopt($curl, CURLOPT_POSTFIELDS, $requestDataQuery);
+                curl_setopt($curl, CURLOPT_POSTFIELDS, $this->requestData);
             }
             elseif ($this->method == self::METHOD_GET) {
                 curl_setopt($curl, CURLOPT_URL, $this->url . '?' . $requestDataQuery);


LukeTowers:
@alxy can you link the commit that you commented on so @w20k can test this more easily?

alxy:
Sure, it's this one: https://github.com/octobercms/library/commit/e4e84783e9c65f9bf03a8171b3d08a9b894f7e66

I don't see the reason for this change anyway, as it was merely a result of a weird (wrong) php configuration of this guy on twitter.

w20k:
@alxy just a minor note, should it be just a `$this->requestData` or wrapped in `http_build_query($this->requestData)`?

alxy:
@w20k This is exactly the thing I reversed - `http_build_query` already trasnforms the array to string format and builds the request data string for us (for the request it needs to be regular string obviously).

Here is a usage example that doesnt work with the current setup, you can test it:
```
$result = Http::post('https://example.org', function($http)  {
            // Add the image data to the request
            $http->data('image', new \CurlFile('path/to/some/file.txt'));
        });
```

w20k:
@LukeTowers - testing is finished! Easy fix ;) 
Status - ✅

w20k:
@alxy could you look at this issue, I'm not sure, but seems like update process won't work with this modification.

screenshot 2019-03-02 at 16 15 53

PR #382, opened 2019-03-02
replace closure event handler with class method

mjauvin:
closures prevent an object from being serialized

### DIFF
--- a/src/Database/Behaviors/Purgeable.php
+++ b/src/Database/Behaviors/Purgeable.php
@@ -40,11 +40,13 @@ class Purgeable extends \October\Rain\Extension\ExtensionBase
          * Remove any purge attributes from the data set
          */
         $model = $this->parent;
-        $model->bindEvent('model.saveInternal', function() use ($model) {
-            $model->purgeAttributes();
-        });
-
+        $model->bindEvent('model.saveInternal', [$this, 'saveInternalHandler']);
     }
+
+    public function saveInternalHandler()
+    {
+        $this->parent->purgeAttributes();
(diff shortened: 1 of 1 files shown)


mjauvin:
I'd really like this one to go through before the next build. It's quite a straightforward replacement from a closure to a method reference, have tested it and it works. I frankly don't see any plausible side-effect.

w20k:
@LukeTowers what do you think? Could this be merged?

LukeTowers:
What's the reason for this change? The more detail the better.

mjauvin:
closures prevent an object from being serialized (like when caching it).

so any object using a formwidget that uses the Purgeable behavior, for example, cannot be cached.

LukeTowers:
example?

mjauvin:
I have a Model that has a MLFileUpload field (https://github.com/rainlab/translate-plugin/pull/442).

If I do the following in my page code section:

```
use StudioAzura\Test\Models\MyModel;

function onStart()
{
  $models = MyModel::all();
  Cache::put('MyTests', $models, 60);
  $this['tests'] = Cache::pull('MyTests');
}
```

I get an error "Serialization of 'Closure' is not allowed"

That PR fixes this.

mjauvin:
But any model which has closures in its event handlers will trigger that error.

LukeTowers:
@mjauvin hmm. Attempting to cache a model in that way will cause issues if you have listeners bound to any events. 

I think we should probably investigate a better way of achieving the end result you're looking for. Thoughts?

PR #388, opened 2019-03-14
Add withTrashed to prevent softDelete slug duplication

Adrion:
fix https://github.com/octobercms/october/issues/4192

@LukeTowers here it is :)

### DIFF
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -13,6 +13,12 @@ trait Sluggable
      */
 
     /**
+     * @var bool Allow trashed slugs to be counted in the slug generation.
+     *
+     * protected $allowTrashedSlugs = false;
+     */
+
+    /**
      * Boot the sluggable trait for a model.
      * @return void
      */
@@ -87,14 +93,10 @@ trait Sluggable
         $counter = 1;
         $separator = $this->getSluggableSeparator();
         $_value = $value;
-
-        if ($this->methodExists('withTrashed') ) {
-            $query = $this->newSluggableQuery()->withTrashed();
-        } else {
-            $query = $this->newSluggableQuery();
-        }
-
-        while ($query->where($name, $_value)->count() > 0) {
+        while (($this->methodExists('withTrashed') && $this->allowTrashedSlugs) ?
+            $this->newSluggableQuery()->where($name, $_value)->withTrashed()->count() > 0 :
+            $this->newSluggableQuery()->where($name, $_value)->count() > 0
+        ) {
             $counter++;
             $_value = $value . $separator . $counter;
         }
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -88,7 +88,7 @@ trait Sluggable
         $separator = $this->getSluggableSeparator();
         $_value = $value;
 
-        if (method_exists($this, 'withTrashed') ) {
+        if ($this->methodExists('withTrashed') ) {
             $query = $this->newSluggableQuery()->withTrashed();
         } else {
             $query = $this->newSluggableQuery();
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -88,7 +88,13 @@ trait Sluggable
         $separator = $this->getSluggableSeparator();
         $_value = $value;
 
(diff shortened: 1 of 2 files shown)


w20k:
@Adrion could you also provide Unit Tests? To cover this functionality :) Thanks!

Adrion:
Also 
> if the name column was set to Cheyenne, as a result the slug column would be set to cheyenne, cheyenne-1, or cheyenne-2, etc before the model is created.

counter start at 1 in `getSluggableUniqueAttributeValue` method. 
so `cheyenne-1` is never created. 

Do you want to start from 0 and get `cheyenne-1` or change the documentation ?

w20k:
@Adrion I'd go for documentation, less `pain` if someone is already using this behavior.

w20k:
@alxy most probably 😄 That's why I've asked for test)

Adrion:
Yes withTrashed method is only for the softDelete trait. 
I'm adding tests and fixing the issue

src/Database/Traits/Sluggable.php
```diff
@@ -88,7 +88,13 @@ protected function getSluggableUniqueAttributeValue($name, $value)
         $separator = $this->getSluggableSeparator();
         $_value = $value;
 
-        while ($this->newSluggableQuery()->where($name, $_value)->count() > 0) {
+        if (method_exists($this, 'withTrashed') ) {
```
LukeTowers: I think you should be using the `$this->methodExists('withTrashed')` method instead because that takes into account dynamic methods.

LukeTowers:
This pull request has been made to the wrong branch. Please review [the contributing guidelines](https://github.com/octobercms/october/blob/master/.github/CONTRIBUTING.md#pull-requests) as all PRs need to be made to the develop branch.

src/Database/Traits/Sluggable.php
w20k: @Adrion ^

Adrion:
That for guidelines. I made your modification request.

LukeTowers:
@w20k can you test please?

w20k:
@LukeTowers sure! Might take a few days, hopefully.

bennothommo:
Could this be made optional, perhaps via a property like `$includeTrashed` defaulted to true? I personally have used soft deletes for storing archived data with no UI functionality to restore it. In this instance, I would like the soft deleted slug to be made available again without necessarily removing the original from the DB.

Adrion:
Add `$ignoreTrashedSlugs` as  a property of Sluggable Trait can easily  be done. Then users will be able to override it in the model. Sluggable and Softdelete traits are really tied together in those cases. 

@w20k,  @LukeTowers  Do you want this option ?

w20k:
@Adrionit sounds like a good option to add, but it's additional labor for you 😄 @bennothommo has a point!

@LukeTowers what do you think?

LukeTowers:
@Adrion @bennothommo I like `allowTrashedSlugs` defaulting to false! Great thinking!

Adrion:
Ok this way, my behavior is optional and can be activated by setting
`protected $allowTrashedSlugs = true;`
By default deleted slugs are ignored, It won't break @bennothommo current behavior

LukeTowers:
@Adrion could you add tests for the `allowTrashedSlugs` property too? On vs. off sort of thing?

w20k:
@Adrion could you ping me when you're done ;)

Adrion:
I added tests for models with and without `$allowTrashedSlugs`:  
https://github.com/octobercms/library/pull/388/commits/b74c1cd1c67488c3f22fdb4cf86d16dd94938eb2#diff-4b55760faf2169b70e22ba9a6e07bd92R182

which cases do you want to cover ? 😄

LukeTowers:
That's fine, missed those on initial review. @w20k ready for your review.

w20k:
Sorry, guys for a delay will test it tomorrow after work!

w20k:
Works like charm! 
@LukeTowers, status - tested ✅ Ready to be merged.
@Adrion - awesome tests! 😄

LukeTowers:
One final request, can a line be added to the docs on the SoftDelete trait about how to use the allowTrashedSlugs property? @Adrion

Adrion:
@LukeTowers Yes, I'll make the PR today. 👍

Adrion:
I added the section in the sluggable part since it's a Sluggable trait property. 
However I also add a link inside the soft delete section. 

My english may needs a little review though 😄

PR #391, opened 2019-04-09
Added event after the model is reordered

jexme:


LukeTowers:
This can be done from the reorder controller itself, there is no need for it to be added directly to the library.

PR #394, opened 2019-04-17
Add missing svg mimetype

Samuell1:
Sometimes some system parse `image/svg` as file and start to download it, using `image/svg+xml` should work as image and it should prevent running js in it.


LukeTowers:
Can you provide an example of the problem this solves?

Samuell1:
@LukeTowers For example AWS S3 if content type is `image/svg` it downloads it. But if you change content type to `image/svg+xml` it works.

PR #395, opened 2019-04-18
Cache file exists checks if debug mode is disabled

Samuell1:
Fixes https://github.com/octobercms/october/issues/4260


LukeTowers:
Implemented in https://github.com/octobercms/library/commit/99c26d2bd06f42e8bec919cfa2b9ac2550326af7

PR #397, opened 2019-04-29
Rewrite config files with env calls

JoshuaBehrens:
After converting config files to .env files one cannot execute `october:install` anymore as it cannot rewrite config files with env-function calls. Because this leads to invalid php files.

To implement this I had to use a different replace-pattern so I had to be able to match regex and replace patterns.

### DIFF
--- a/src/Config/ConfigWriter.php
+++ b/src/Config/ConfigWriter.php
@@ -17,6 +17,7 @@ use Exception;
  * - booleans
  * - nulls
  * - single-dimension arrays
+ * - default values in env function calls
  *
  * To do:
  * - When an entry does not exist, provide a way to create it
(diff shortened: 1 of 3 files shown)


LukeTowers:
Why do you want to run october:install after replacing the config values?

JoshuaBehrens:
From other projects I am used to have .env files. So I used `october:env` to get my .env file (which led to an .env.example file in git). I cloned the project on a different machine and wanted to have everything setup. As `october:install` prepares everything like creating administration users etc it feels like the best command to me.

LukeTowers:
@JoshuaBehrens I would recommend using `october:up` instead. Does basically everything that `:install` does

JoshuaBehrens:
Didn't thought about `october:up`. I might have missed that in the docs or mixed that up with laravel's `up`. Thanks for the hint.

Beyond that. The component is about rewriting configuration files and env calls in configurations are common in there, aren't they? Is this change viable for this library?

LukeTowers:
Potentially, my main concern is that it still works fine without env() calls. Also, it's kind of weird to say that ConfigWriter supports env() when it isn't actually changing the value of that env() key, it's only able to change the default value of it. Thoughts?

JoshuaBehrens:
To support this I need to locate .env files. Do I want to change the values in .env or .env.example? In every? To make this possible I have to use a `preg_replace_callback` as I have to react on this only if I replace I replace an env call.

To make things simpler one could also just replace the whole call (not just the default value) and use `october:env` later to export it to the .env files again.

LukeTowers:
@JoshuaBehrens and here we get into the reason that we don't currently support this :) I don't think any of those solutions would work. 

It does say right in the documentation that `october:install` should not be run after `october:env`, do you have any problems with that?

JoshuaBehrens:
I found it now in https://octobercms.com/docs/console/commands#console-install-command. I wasn't aware of that hint back then. It is ok that it is documentation. RTFM I would say :sweat_smile: 

Well I had a problem that `october:install` reconfigures my files and `october:up` is instead the way to go for me.

LukeTowers:
Great! Thanks for your interest in helping to improve OctoberCMS though!

PR #398, opened 2019-05-01
Removed spaces from stubs so str_replace will place correct values

NathanBastiaans:
When creating a formwidget with artisan the generated javascript and stylesheet file will show the line 
`This is a sample JavaScript file used by {{ name }}` 
instead of the actual given name i.e.
`This is a sample JavaScript file used by MyFormWidget`
These changes fix this problem


LukeTowers:
Weird, stub files use str_replace? Why not Twig?

daftspunk:
Good catch, thanks!

PR #400, opened 2019-05-04
Add missing "fit" mode in class' DocBlock

matteotrubini:


bennothommo:
@matteotrubini Thanks for this. We are currently rolling out a new October CMS release, after that is completed, we will merge this.

PR #404, opened 2019-05-22
Implement "http_build_url" polyfill / wrapper into core

bennothommo:
Allows October sites to use the `http_build_url` method, either natively through the PECL HTTP extension, or through the userland polyfill implemented in this PR. This method is a counterpart to the [`parse_url`](https://php.net/parse_url) method provided in PHP - in essence, taking a structured array of URL components and making it a single URL string.

In addition, the PR extends the `Illuminate\Routing\UrlGenerator` class with a new class `October\Rain\Route\UrlGenerator` and uses this as the new endpoint for the `Url` facade.

### DIFF
--- a/src/Router/RoutingServiceProvider.php
+++ b/src/Router/RoutingServiceProvider.php
@@ -15,4 +15,40 @@ class RoutingServiceProvider extends RoutingServiceProviderBase
             return new CoreRouter($app['events'], $app);
         });
     }
+
+    /**
+     * Register the URL generator service.
+     *
+     * @return void
+     */
+    protected function registerUrlGenerator()
+    {
+        $this->app->singleton('url', function ($app) {
+            $routes = $app['router']->getRoutes();
+
+            // The URL generator needs the route collection that exists on the router.
+            // Keep in mind this is an object, so we're passing by references here
+            // and all the registered routes will be available to the generator.
+            $app->instance('routes', $routes);
+
+            $url = new UrlGenerator(
+                $routes, $app->rebinding(
+                    'request', $this->requestRebinder()
+                )
+            );
+
+            $url->setSessionResolver(function () {
+                return $this->app['session'];
+            });
+
+            // If the route collection is "rebound", for example, when the routes stay
+            // cached for the application, we will need to rebind the routes on the
+            // URL generator instance so it has the latest version of the routes.
+            $app->rebinding('routes', function ($app, $routes) {
+                $app['url']->setRoutes($routes);
+            });
+
+            return $url;
+        });
+    }
 }
--- a/src/Router/UrlGenerator.php
+++ b/src/Router/UrlGenerator.php
@@ -1,0 +1,69 @@
+<?php
+namespace October\Rain\Router;
+
+use Illuminate\Routing\UrlGenerator as UrlGeneratorBase;
+
+// PECL HTTP constant definitions
+if (!defined('HTTP_URL_REPLACE')) {
+    define('HTTP_URL_REPLACE', 1);
+}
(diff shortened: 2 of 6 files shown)


mjauvin:
Cool!

Could it be made as a static method so it could be used like:

`\October\Rain\Route\UrlGenerator::buildUrl()`

LukeTowers:
@mjauvin you should be able to do `Url::buildUrl()` with `use Url;` instead.

src/Router/UrlGenerator.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param array $newUrl If set, this will be filled with the array parts of the composed URL, similar to the return
+     *                      value of `parse_url`.
+     *
+     * @return string
+     */
+    public function buildUrl(array $url, array $parts = [], $flags = HTTP_URL_REPLACE, &$newUrl = []): string
+    {
+        if (function_exists('http_build_url')) {
+            return http_build_url($url, $parts, $flags, $newUrl);
+        }
```
LukeTowers: I'm not sure we should do that here. I'm imagining that we would define http_build_url() ourselves as a helper wrapper around Url::buildUrl() so this would be unnecessary / cause an infinite loop

mjauvin: @LukeTowers tried the Url::buildUrl() and it says it cannot find buildUrl method

mjauvin: @LukeTowers tried the Url::buildUrl() and it says it cannot find buildUrl method

mjauvin: @LukeTowers Not sure what problem you see here, looks fine to me. If the PECL http module method exists (http_build_url), it is used, otherwise we build the url ourselves with the code that follows.

bennothommo: @mjauvin It will be able to find the method, when it's finished ;)

@LukeTowers Good point. Will change it around.

bennothommo:
@LukeTowers @mjauvin It's now finished, if you guys would like to review it :)

w20k:
Awesome job @bennothommo, as always 👍

bennothommo:
@w20k Fixed the unit tests. It seems that when testing the Rain library in isolation, it does not know about facades :/. The unit tests were working fine for me when run with the entire codebase.

w20k:
@bennothommo Thanks for the comment, will add it to my list of Todos ;) To investigate.

src/Router/RoutingServiceProvider.php
```diff
@@ -15,4 +15,40 @@ protected function registerRouter()
...
+        $this->app->singleton('url', function ($app) {
+            $routes = $app['router']->getRoutes();
+
+            // The URL generator needs the route collection that exists on the router.
+            // Keep in mind this is an object, so we're passing by references here
+            // and all the registered routes will be available to the generator.
+            $app->instance('routes', $routes);
+
+            $url = new UrlGenerator(
+                $routes, $app->rebinding(
```
mjauvin: If I am not mistaken, the UrlGenerator class is not included in here.

bennothommo: @mjauvin You don't have to use a `use` case for a class in the same folder ;)

mjauvin: Ah, good to know! :)

mjauvin:
I confirm this works WITHOUT PECL HTTP module.

Can anyone install the PECL HTTP module to test this? Unfortunately I was not able to do so in my environment.

LukeTowers:
I'm fine with proceeding without verification with PECL HTTP - everything that would be a conflict is appropriately wrapped in the proper checks to make sure that nothing gets double defined so this should be fine.

PR #406, opened 2019-05-30
Various improvements to File handling logic

LukeTowers:
Improves File handling logic by running file commands on the disk that the file is on and centralizes the logic to generate the diskPath in one method.


mjauvin:
I confirm.

PR #407, opened 2019-06-01
Add sanitisation of HTML generated by Markdown parser.

bennothommo:
Fixes https://github.com/rainlab/translate-plugin/issues/484.

Adds sanitisation of the resulting HTML generated by the Markdown parser, preventing tags like `<script>` and `<style>` from being added to Markdown content.

The whitelist of HTML tags follows the tags provided in the Github-flavoured Markdown Spec:
https://github.github.com/gfm/#html-blocks

### DIFF
--- a/src/Parse/Markdown.php
+++ b/src/Parse/Markdown.php
@@ -93,7 +93,90 @@ class Markdown
         $this->fireEvent('parse', [$text, $data], false);
         Event::fire('markdown.parse', [$text, $data], false);
 
+        // Strip out certain HTML elements (eg. <script> or <style>)
+        $data->text = $this->sanitizeHtml($data->text);
+
         return $data->text;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Parse/Markdown.php
```diff
@@ -93,9 +93,92 @@ protected function parseInternal($text, $method = 'text')
...
+            'tfoot',
+            'th',
+            'thead',
+            'title',
+            'tr',
+            'track',
+            'ul'
+        ];
+
+        return strip_tags($text, implode('', array_map(function ($tag) {
```
w20k: @bennothommo I think using `strip_tags` not a good option for sanitizing values.
JS injection could be added directly into input fields as well. Also `strip_tags` doesn't validate Html, so non-closed tags can result in the removal of more text/data than expected. 

`htmlspecialchars` or `htmlentities` might be a better option.

PR #408, opened 2019-06-03
improve AddDynamicPoperty test

mjauvin:
- add test to verify existence of attribute after addDynamicProperty() call
- add comment about post-save test (dynamic properties should not persist to db)


bennothommo:
@mjauvin Just for posterity, would you be able to add an additional assertion to make sure that the dynamic property's value is the same as the one defined above it? No harm in testing that too.

PR #409, opened 2019-06-05
Add scssphp/scssphp

greatislander:
[scssphp has moved](https://github.com/scssphp/scssphp/issues/4). This PR updates the composer dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,12 @@
         "kriswallsmith/assetic": "~1.3",
         "linkorb/jsmin-php": "~1.0",
         "oyejorge/less.php": "~1.7",
-        "leafo/scssphp": "~0.6",
         "symfony/yaml": "~3.2",
         "twig/twig": "~2.0",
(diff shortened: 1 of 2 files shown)


LukeTowers:
@greatislander looking into this more I don't think we actually use that dependency directly anywhere in our own code, we just pull in Assetic and assetic uses it.

bennothommo:
@LukeTowers Looks like Assetic hasn't had a [commit](https://github.com/kriswallsmith/assetic/commits/master) since November 2016, so it's unlikely they'll change it.

LukeTowers:
It's unlikely that we'll change it either, I don't even think we use it directly, I think we can just remove it from our composer.json altogether

LukeTowers:
blocked until assetic updates their copy

bennothommo:
This is going to be resolved by the L6 upgrade, when we switch over to using the new [Assetic fork](https://github.com/assetic-php/assetic).

PR #410, opened 2019-06-19
Change parameter insertData from value to reference

patriknemecek:
Create possibility to set default values to pivot attributes by specific model

### DIFF
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -126,7 +126,7 @@ class BelongsToMany extends BelongsToManyBase
          *     });
          *
          */
-        if ($this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, $insertData], true) === false) {
+        if ($this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, &$insertData], true) === false) {
             return;
         }
 
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -118,7 +118,7 @@ class BelongsToMany extends BelongsToManyBase
          *
          * Example usage:
          *
-         *     $model->bindEvent('model.relation.beforeAttach', function (string $relationName, array $attachedIdList, array $insertData) use (\October\Rain\Database\Model $model) {
+         *     $model->bindEvent('model.relation.beforeAttach', function (string $relationName, array $attachedIdList, array &$insertData) use (\October\Rain\Database\Model $model) {
          *         if (!$model->isRelationValid($attachedIdList)) {
          *             throw new \Exception("Invalid relation!");
          *             return false;


LukeTowers:
@patriknemecek could you provide an example of how you'll be using this in the real world?

patriknemecek:
If a belongsToMany relation have custom attributes which not have default value set in MySQL table, assigning another models to this relation throw an error(column doesn't have a default value). For security reason we use global scope in every model and custom attributes like agency_id and agency_code. If insertData would be passed by reference I could set this custom values.

mjauvin:
Wouldn't changing the insertData argument type break existing code listening to the event?

patriknemecek:
This change probably not break the core.

mjauvin:
Not the core, but if someone wrote a Plugin listening to the event, it will break the plugin since the plugin is not expecting a reference to insertData...

LukeTowers:
@patriknemecek I need to see an actual code example of how you are going to be using this changed event.

patriknemecek:
Here is example how this feature will be used in behaviour

`
class AgencyBehavior extends ExtensionBase
{
    /**
     * @var \October\Rain\Database\Model Reference to the extended model.
     */
    protected $model;

    /**
     * Constructor
     * @param \October\Rain\Database\Model $model The extended model.
     */
    public function __construct($model)
    {
        $this->model = $model;

        $this->model->addGlobalScope(new AgencyScope);

        $this->model->bindEvent('model.beforeSave', function () use ($model) {
            $agency_id = Agency::agency_id();

            if ($model->agency_id && $model->agency_id != $agency_id) {
                throw new \Exception('Mismatched agency_id');
            }

            $model->agency_id = $agency_id;
        });

        $this->model->bindEvent('model.relation.beforeAttach', function (string $relationName, array $attachedIdList, array &$insertData) {
            $agency_id = Agency::agency_id();

            foreach ($insertData as $index => $value) {
                $insertData[$index]['agency_id'] = $agency_id;
            }
        });
    }
}
`

patriknemecek:
Here is example how this feature will be used in behaviour

```
class AgencyBehavior extends ExtensionBase
{
    /**
     * @var \October\Rain\Database\Model Reference to the extended model.
     */
    protected $model;

    /**
     * Constructor
     * @param \October\Rain\Database\Model $model The extended model.
     */
    public function __construct($model)
    {
        $this->model = $model;

        $this->model->addGlobalScope(new AgencyScope);

        $this->model->bindEvent('model.beforeSave', function () use ($model) {
            $agency_id = Agency::agency_id();

            if ($model->agency_id && $model->agency_id != $agency_id) {
                throw new \Exception('Mismatched agency_id');
            }

            $model->agency_id = $agency_id;
        });

        $this->model->bindEvent('model.relation.beforeAttach', function (string $relationName, array $attachedIdList, array &$insertData) {
            $agency_id = Agency::agency_id();

            foreach ($insertData as $index => $value) {
                $insertData[$index]['agency_id'] = $agency_id;
            }
        });
    }
}
```

patriknemecek:
Here is example how this feature will be used in behavior

```
class AgencyBehavior extends ExtensionBase
{
    /**
     * @var \October\Rain\Database\Model Reference to the extended model.
     */
    protected $model;

    /**
     * Constructor
     * @param \October\Rain\Database\Model $model The extended model.
     */
    public function __construct($model)
    {
        $this->model = $model;

        $this->model->addGlobalScope(new AgencyScope);

        $this->model->bindEvent('model.beforeSave', function () use ($model) {
            $agency_id = Agency::agency_id();

            if ($model->agency_id && $model->agency_id != $agency_id) {
                throw new \Exception('Mismatched agency_id');
            }

            $model->agency_id = $agency_id;
        });

        $this->model->bindEvent('model.relation.beforeAttach', function (string $relationName, array $attachedIdList, array &$insertData) {
            $agency_id = Agency::agency_id();

            foreach ($insertData as $index => $value) {
                $insertData[$index]['agency_id'] = $agency_id;
            }
        });
    }
}
```

LukeTowers:
Alright. If you update the docblock comment documenting the event to document that it's now passed by reference I'm fine with this. @mjauvin I don't think it'll be a huge issue, it was added 10 months ago and I don't think a lot of people even use the event.

patriknemecek:
No, this change doesn't break event listeners with value passing. I tested it. The docblock change is done.

mjauvin:
I was not aware of this, but it seems you don't need to "pass" by reference anymore, just declare the function arguments as a reference:

https://www.php.net/manual/en/language.references.pass.php

mjauvin:
So this would mean this PR would not be required, just declare your anonymous listener argument to use a reference...

LukeTowers:
@patriknemecek can you test @mjauvin's suggestion? If that works for you I would prefer you do that instead.

mjauvin:
I used this in one of my plugins, and it worked. But, please, confirm this
yourself.

--
Marc

patriknemecek:
> @patriknemecek can you test @mjauvin's suggestion? If that works for you I would prefer you do that instead.

I tried it before I made merge request. I got this type of error : "Parameter 3 to ...\Agency\Classes\ModelExtend::...\Agency\Classes\{closure}() expected to be a reference, value given" and when I make change in fire event it works.

mjauvin:
What version of php are you using?

--
Marc

patriknemecek:
@mjauvin I use PHP 7.1.9

mjauvin:
Indeed, you're right. I don't understand why the article I referenced above
says otherwise, this is odd.

--
Marc

patriknemecek:
This code is example why declare event listener argument as reference doesn't work.

```
function bindEvent($evetName, $callback) {
	$insertData = [];

	// 1. bindEvent is now implemented in OctoberCMS this way and throw error Parameter 3 to {closure}() expected to be a reference, value given
	$params = [null, null, $insertData];
	call_user_func_array($callback, $params);

	// 2. This works like I want
	// $params = [null, null, &$insertData];
	// call_user_func_array($callback, $params);

	// 3. This works like article you read
	// $callback(null, null, $insertData);


	var_dump($insertData);
}

bindEvent('model.relation.beforeAttach', function ($relationName, $attachedIdList, array &$insertData) {
	$insertData['work'] = true;
});		

```

mjauvin:
ok, so call_user_func_array() is to blame here... makes sense.

mjauvin:
Ok, so it's all good then!

bennothommo:
Yeah, it appears to be a unique behaviour to the `call_user_func` / `call_user_func_array` methods. The reference has to be defined both in the closure and the definition. I'm not sure if it's the same when using the [closure `call` method](https://www.php.net/manual/en/closure.call.php) but I don't think we should the event handling to that anyways.

I'd be wary of changing the event attribute to a reference though, even if it is a recent addition. If several plugins were to extend the same event, there's a great deal of scope for error if the insert data is modified by both. It seems this PR is trying to fix an issue that would be better suited to being fixed by defining some defaults in the database and changing the pivot data in the model prior to saving, or modifying the pivot data directly after the fact.

patriknemecek:
This problem is not be able fix with Schema default column value. I need set dynamic value for pivot columns. I tough I can set "default" value (maybe in belongsToMany relation definition) for this pivot columns when belongsToMany relation attaching models to pivot table, but I can't.

PR #411, opened 2019-07-08
Don't minify css rules inside parentheses

tobias-kuendig:
This is my solution to https://github.com/octobercms/october/issues/4438. 

By adding the negative lookahead `(?![^\(]*\))` everything inside of `()` is ignored. This includes `calc()` rules.

If someone can come up with a RegEx that only matches `calc()` that would be even better. I couldn't get it done within a few minutes so I've opted for this simpler solution.


w20k:
@tobias-kuendig looks good to me. I think we could leave it as it is right now because there is more than one function, like `calc`, that might be using the same approach (can't recall any 😄).

@bennothommo, @LukeTowers anything to add?

bennothommo:
@w20k @tobias-kuendig Looks good to me!

PR #418, opened 2019-08-05
Allow for using a static class method to return option data for variables

jaxwilko:
This change allows for the dev to pass `options="\Path\To\Class|method"` as part of a variable declaration.

The idea behind this is I would like to be able to select from a list of options kept in the database (i.e. blog posts). However this functionality is not limited to working with the database and could be used in anyway the developer sees fit, including linking to APIs or executing run time code to improve the options presented to the end user.

An example of how you could utilise this functionality is by defining a class and method to call:

```php
namespace Jack\Plugin;

class BlogPost extends Model
{
    public function getDropdownList()
    {
        $result = [];
        foreach (self::all() as $record) {
            $result[$record->id] = $record->title;
        }
        return $result; 
    }
}
```
And then populate your dropdown via:

```
{variable type="dropdown" name="blogPostId" label="Blog Post" options="\Jack\Plugin\BlogPost|getDropdownList"}{/variable}
```


w20k:
Awesome, thanks for Tests ;) Just wanted to comment about those ;)

JaxWilko:
My pleasure :)

JaxWilko:
I'm thinking about changing the order of operations to:

```diff
diff --git a/src/Parse/Syntax/FieldParser.php b/src/Parse/Syntax/FieldParser.php
index 3f890f5..6778f23 100644
--- a/src/Parse/Syntax/FieldParser.php
+++ b/src/Parse/Syntax/FieldParser.php
@@ -391,15 +391,8 @@ class FieldParser
 
         $result = [];
 
-        if (count($options) === 2 && class_exists($options[0])) {
-            if (!method_exists($options[0], $options[1])) {
-                throw new Exception(sprintf(
-                    'Invalid drop-down option: method `%s` not found in `%s`',
-                    $options[1],
-                    $options[0]
-                ));
-            }
-
+        if (count($options) === 2 && class_exists($options[0]) && method_exists($options[0], $options[1])) {
+          
             $result = $options[0]::{$options[1]}();
 
             if (!is_array($result)) {
```

As this would allow fall through in the case of things like:

```
{variable type="dropdown" name="class" label="Class Select" options="\Jack\Plugin\BlogPost|use blog post class"}{/variable}
```

That would allow users to still have options of classes and display names without throwing exceptions.

Does this sound like a good idea?

bennothommo:
@JaxWilko Is there any reason that you cannot just use a normal static call? ie. `\Jack\Plugin\BlogPost::getDropdownList`.

JaxWilko:
I spoke to Luke on Slack about this and we thought it would be nice to keep
it consistent with the current option format.

Happy to change it though, we also discussed using the scope resolution
operator as you suggested or the @ similar to Laravel routes.

I think the key aspect of this is what would be most logical for the
developer's using the feature, which would probably be the scope resolution
operator.

bennothommo:
@JaxWilko Ah fair enough. I feel it might make it more clear it's a method call with `::`, and it would eliminate the issue you posted where someone uses the namespace as an actual value: ie. `\Jack\Plugin\BlogPost|Blog Post Model`. @LukeTowers what do you think?

JaxWilko:
The more I think about it, the more I prefer `::` does anybody have objections if I make that change?

JaxWilko:
I kept the fallout handling so it'll revert to using the default if the class or method don't exist, however I think if we have found a `::` it may be better to assume that the dev wants to call a class and return them a better exception message along the lines of "class not found" or "method not found".

At present the fallout will lead to the exception:

```php
throw new Exception(sprintf(
     'Invalid drop-down option key: %s. Option keys can contain only digits, Latin letters and characters _ and -', $key
));
```

Because the class path would not match the key format.

It depends how the `::` is handled. At present if you have `options="test::ing"` the result is:

```
["options"]=>
array(1) {
   ["test"]=>
   string(4) ":ing"
}
```
Which seems makes sense.

A solution to keep both functionalities would be to set a variable on the static class method check fail and then changing the exception message if that variable is set.

Or we just leave it as is and let the error message be a little misleading.

LukeTowers:
@JaxWilko @bennothommo I'm fine with the `::` format, but whatever the final solution is we need to make sure that the Form widget getModelOptions (and Lists widget when that's done) matches the support here.

LukeTowers:
@JaxWilko I'm fine with `NonExistantClass::wrongMethod` leads to 
```php
[
    'NonExistantClass` => `:wrongMethod`
]
```
I feel like that offers the most support while also giving enough of a hint to developers that are seeing incorrect results to fix it.

LukeTowers:
This pull request has been made to the wrong branch. Please review [the contributing guidelines](https://github.com/octobercms/october/blob/master/.github/CONTRIBUTING.md#pull-requests) as all PRs need to be made to the develop branch.

JaxWilko:
Ah my bad, thank you very much. I shall make sure I remember next time

bennothommo:
@JaxWilko is this ready to test?

JaxWilko:
@bennothommo yes should be, if there is anything I need to do then just let me know

src/Parse/Syntax/FieldParser.php
```diff
@@ -377,17 +377,36 @@ protected function processTagsRegex($string, $tags)
     /**
      * Splits an option string to an array.
      *
-     * one|two           -> [one, two]
-     * one:One|two:Two   -> [one => 'One', two => 'Two']
+     * one|two               -> [one, two]
+     * one:One|two:Two       -> [one => 'One', two => 'Two']
+     * \Path\To\Class|method -> \Path\To\Class::method()
```
LukeTowers: @JaxWilko is this docblock up to date with the latest format?

jaxwilko: @LukeTowers ah no it's not been updated. Should read `\Path\To\Class::method -> \Path\To\Class::method(): array` I will push an update when I get on my main PC.

LukeTowers:
Perfect, just need some testing now. @JaxWilko could you also submit a PR to the relevant docs (not sure if it's in the October/docs repo or just discussed in RainLab.Pages or if both)

bennothommo:
@JaxWilko If you are still interested in getting this merged with build 460, we will need the update to the relevant docs and we'll also need to resolve the merge conflict with the `SyntaxParserTest` class above.

JaxWilko:
@bennothommo please see https://github.com/octobercms/docs/pull/414

Thanks :)

wordshop-git:
Trying to test this but after editing the files and adding the variable code to my page I'm getting this?

`{variable name="dropdown" label="Test Menu" tab="Test Menu" type="dropdown" options="\Path\To\Class::method"}{/variable}`

```
"Invalid drop-down option key: \Path\To\Class. Option keys can contain only digits, Latin letters and characters _ and -" on line 419 of C:\xampp\htdocs\gmv2\vendor\october\rain\src\Parse\Syntax\FieldParser.php
```

I'm completely lost in how and wherre I define this class/ model dealio

JaxWilko:
@wordshop-git you can use any class that is autoloaded by composer. Ideally you'd create a plugin or extend an existing one. Does that help?

wordshop-git:
> @wordshop-git you can use any class that is autoloaded by composer. Ideally you'd create a plugin or extend an existing one. Does that help?

Thanks for the quick reply! I wish it did, I want to prefill the dropdown with a list of static menus from the static pages plugin so I put the getDropdownList() function in rainlab/pages/classes/Menu.php

JaxWilko:
@wordshop-git really you shouldn't be modifying a plugin created by somebody else as next time they update their plugin it will overwrite your changes. You could create a very small plugin that just does what you need however. Check the docs here https://octobercms.com/docs/plugin/registration But all you should need is plugins/[authorname]/[pluginname]/Plugin.php & plugins/[authorname]/[pluginname]/classes/[MyMenuExtensionClassName].php.

Then you'll be able to call that class via \AuthorName\PluginName\Classes\MyMenuExtensionClassName::function

JaxWilko:
@wordshop-git Oh, either that or add the functionality to the Pages plugin and raise a pull request to get it merged in, it might be a useful feature for others :)

wordshop-git:
> @wordshop-git really you shouldn't be modifying a plugin created by somebody else as next time they update their plugin it will overwrite your changes. You could create a very small plugin that just does what you need however. Check the docs here https://octobercms.com/docs/plugin/registration But all you should need is plugins/[authorname]/[pluginname]/Plugin.php & plugins/[authorname]/[pluginname]/classes/[MyMenuExtensionClassName].php.
> 
> Then you'll be able to call that class via \AuthorName\PluginName\Classes\MyMenuExtensionClassName::function

I knew it was bad but I don't even know where to begin with extending plugins and all that jazz

So basically create a new plugin called menus for instance, that has two files (plugin.php, classes/menuext.php) and with the menuext.php extend the rainlab/pages plugin to pull the menus across into my own plugin? Then with menus/plugin.php call the getDropdownOptions() function and then reference the menus plugin with the variable?

wordshop-git:
> @wordshop-git Oh, either that or add the functionality to the Pages plugin and raise a pull request to get it merged in, it might be a useful feature for others :)

I'd love to but I don't know how to and don't really have the time to figure it out

I want a dynamic menu list in my back end variable to allow users to choose a different menu so I feel like this would be something other people would want to do

I have a layout that's used on 6 different sections of a site each section has a different menu but I didn't want to create 6 different layouts because then I'd have to make 1 change in 6 places over and over

JaxWilko:
@wordshop-git the Plugin.php file will just be there to register the plugin. Your MenuExt.php file can contain all your logic and can be called directly. Just remember to be case sensative with your file names :)

wordshop-git:
@JaxWilko Alright, then I suppose I'll give it a shot. Would doing this be classed as extending a plugin?

I don't really understand what it is even though I see it mentioned everywhere

JaxWilko:
@wordshop-git if you want you can join the octobercms slack and message me there, happy to give you a hand :)

LukeTowers:
@bennothommo what's the status of this?

wordshop-git:
> @bennothommo what's the status of this?

Can say it works pretty well with the menu list from the static pages plugin haven't really tested it other than this mind

bennothommo:
@LukeTowers I've still gotta give it a test - hopefully will get to that soon.

src/Parse/Syntax/FieldParser.php
```diff
@@ -377,17 +377,36 @@ protected function processTagsRegex($string, $tags)
...
     protected function processOptionsToArray($optionsString)
     {
+        $result = [];
+
+        if (strpos($optionsString, '::') !== false) {
+            $options = explode('::', $optionsString);
+            if (count($options) === 2 && class_exists($options[0]) && method_exists($options[0], $options[1])) {
+                $result = $options[0]::{$options[1]}();
+                if (!is_array($result)) {
+                    throw new Exception(sprintf(
```
bennothommo: @JaxWilko 
```suggestion
                    throw new ApplicationException(sprintf(
```

src/Parse/Syntax/FieldParser.php
```diff
@@ -377,17 +377,36 @@ protected function processTagsRegex($string, $tags)
...
+                        'Invalid drop-down option result: `%s::%s` invalid return type',
```
bennothommo: @JaxWilko 
```suggestion
                        'Invalid dropdown option array returned by `%s::%s`',
```

bennothommo (dismissed):

src/Parse/Syntax/FieldParser.php
jaxwilko: Pressed the wrong thing, will make the change in the morning :)

wordshop-git:
Hello all! What needs to be done to get this in the live build of October?

We're currently using this on a few of our sites and I'd really like it to go live soon so I'm more than happy to try and help out where possible!

src/Parse/Syntax/FieldParser.php
```diff
@@ -377,17 +377,36 @@ protected function processTagsRegex($string, $tags)
...
      *
      * @param  string $optionsString
+     * @throws Exception
      * @return array
      */
     protected function processOptionsToArray($optionsString)
     {
+        $result = [];
+
+        if (strpos($optionsString, '::') !== false) {
```
LukeTowers:
```suggestion
        if (str_contains($optionsString, '::')) {
```

PR #421, opened 2019-08-22
Add AV1 and aPNG formats and put in alphabetical order

ayumihamsaki:
Related to my comment here: https://github.com/octobercms/october/issues/4136#issuecomment-524084998

1. Added missing formats `av1` and `apng`

2. Put the file formats into alphabetical order, beforehand was a mess and all over the place

### DIFF
--- a/src/Filesystem/Definitions.php
+++ b/src/Filesystem/Definitions.php
@@ -97,48 +97,49 @@ class Definitions
     protected function defaultExtensions()
     {
         return [
-            'jpg',
-            'jpeg',
+            'apng',
+            'av1',
+            'avi',
             'bmp',
-            'png',
-            'webp',
-            'gif',
-            'svg',
-            'js',
-            'map',
-            'ico',
             'css',
-            'less',
-            'scss',
-            'ics',
-            'odt',
             'doc',
             'docx',
-            'ppt',
-            'pptx',
-            'pdf',
-            'swf',
-            'txt',
-            'xml',
-            'ods',
-            'xls',
-            'xlsx',
             'eot',
-            'woff',
-            'woff2',
-            'ttf',
             'flv',
-            'wmv',
-            'mp3',
-            'ogg',
-            'wav',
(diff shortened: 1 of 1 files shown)


bennothommo:
@ayumihamsaki in your linked discussion, they were discussing ways to *disable* WEBP due to it not being supported properly, and this particular PR is suggesting to implement 2 formats which have the same, if not less, support than WEBP. I'm not sure it's the greatest idea to allow people to upload them by default. If people wish to allow it on their site, they always have the `cms.fileDefinitions` config value. Thoughts?

ayumihamsaki:
@bennothommo I disagree with you saying less support.

`.apng` files are support by all major browsers right now the reason caniuse says 86% is because they are including: QQ and Baidu browsers which no one uses in china anyway (they are filled with Malware - I know I've tested them in virsustotal).

Support:

- Edge
- Chrome
- Firefox
- Opera
- Safari
- Android
- Samsung

The only two are UC Browser and IE11 not supported.

---

AV1 this is going to be the new standard of video on the internet, most of youtube now runs using AV1. It's made by Google and **all the browsers have agree to support it**.

AV1 is a next generation codec developed by the Alliance for Open Media. AV1 improves compression efficiency by greater than 30% over the current state-of-the-art video codec, VP9. 

I could go one step further and say AV1 will be the standard from now until end of 2021 and then Versatile Video Coding (VVC) (MPEG-I Part 3) maybe the new standard. This will be rolled out in 2022.

Webp is for images
Webm is for video

Google has already switched from recommending to use `webm` into `av1`.

---

That's all I can say really.

bennothommo:
@ayumihamsaki I've discussed this with the other maintainers and at this stage while we are still supporting IE11 and the actual current support for the AV1 format is low, we're not going to change the default file extensions to support these new formats. As per my previous comment, if a developer wants to use either formats on their site, they can modify the `cms.fileDefinitions` configuration value to add them in.

PR #423, opened 2019-09-05
Fire event after impersonate

rubenvanerk:
I would like to redirect after impersonating as per https://github.com/rainlab/user-plugin/issues/400.

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -712,6 +712,8 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
         if (!$this->isImpersonator()) {
             Session::put($this->sessionKey.'_impersonate', $oldSession);
         }
+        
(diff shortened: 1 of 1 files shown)


bennothommo:
@rubenvanerk It looks like `afterImpersonate` is already an event, fired here: https://github.com/octobercms/library/blob/807ae284a80f371d78f44b98ac141a1b26e81126/src/Auth/Manager.php#L742

You may need to call yours `impersonate` instead.

rubenvanerk:
@bennothommo You're correct, changed the event name.

LukeTowers:
@rubenvanerk why can't you use the existing events: https://github.com/octobercms/library/commit/dc5952d5b010602116012383ae9f1c6bf02d9182? Also @bennothommo we need to update the docs on those, I copy pasted them from model.afterBoot it looks like

LukeTowers:
Provided an alternative solution https://github.com/rainlab/user-plugin/issues/400#issuecomment-528925399

PR #424, opened 2019-09-17
Allow using Rule objects in validation rules

filipac:
Right now, if we do something like this in a model:
```php
public function beforeValidate()
    {
        $this->rules = array_merge($this->rules, [
            'prices' => ['required', new CustomRule()],
        ]);
    }
```

We get an error because the Validation trait assumes everything there is a string. We can allow this usage of rules by checking if `$rulePart` is an object, and if that's true, continue the loop. This is in function `processValidationRules`.

```php
                if(is_object($rulePart)) {
                    continue;
                }
```

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -355,6 +355,12 @@ trait Validation
 
         foreach ($rules as $field => $ruleParts) {
             /*
+             * Rule class instances should not be checked here
+             */
+            if(is_object($rulePart)) {
+                continue;
+            }
+            /*
              * Trim empty rules
              */
             if (is_string($ruleParts) && trim($ruleParts) == '') {
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -355,12 +355,6 @@ trait Validation
 
         foreach ($rules as $field => $ruleParts) {
             /*
-             * Rule class instances should not be checked here
-             */
-            if(is_object($rulePart)) {
-                continue;
-            }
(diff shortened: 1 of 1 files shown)


bennothommo:
@filipac this same change was proposed in this PR: https://github.com/octobercms/library/pull/365.

If you would like to see it introduced, you may need to ensure that validation rules allow the attribute labels to be correctly inserted in the error messages, as was suggested in that PR.

src/Database/Traits/Validation.php
```diff
@@ -373,6 +373,12 @@ protected function processValidationRules($rules)
              * Analyse each rule individually
              */
             foreach ($ruleParts as $key => $rulePart) {
+                /*
+                 * Rule class instances should not be checked here
+                 */
+                if(is_object($rulePart)) {
```
bennothommo: @filipac This should check that the `$rulePart` is a `Rule` object, and not just any arbitrary object.

Also, please add a space between the `if` and opening bracket. :)

filipac: Afaik any class can be used there, not only those implementing Rule. Am i wrong? I am on the mobile now

bennothommo: @filipac All objects were being skipped before your change, so I don't believe so. I would rather us ensure only `Rule` objects are used instead of string values just to keep consistency with Laravel.

bennothommo:
Closing this in favour of #455. I've implemented your change, @filipac, and will make sure you get some credit for it.

PR #425, opened 2019-09-17
Fixed Error in Comment

mahony0:
Http library throws error if you use setOption key as string like **$http->setOption('CURLOPT_RETURNTRANSFER', true);**

> "curl_setopt_array(): Array keys must be CURLOPT constants or equivalent integer values" on line 261 of .../vendor/october/rain/src/Network/Http.php"

Both README and Http.php comments fixed by changing array key to the CURLOPT constant.

### DIFF
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -50,7 +50,7 @@
  *       $http->toFile('some/path/to/a/file.txt');
  *
  *       // Sets a cURL option manually
- *       $http->setOption('CURLOPT_SSL_VERIFYHOST', false);
+ *       $http->setOption(CURLOPT_SSL_VERIFYHOST, false);
  *
  *   });
  *
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -520,7 +520,22 @@ class Http
             return;
         }
 
-        $this->requestOptions[$option] = $value;
+        if (is_string($option) && defined($option)) {
+            $optionKey = constant($option);
+            $this->requestOptions[$optionKey] = $value;
+        } elseif (is_int($option)) {
+            $constants = get_defined_constants(true);
+            $curlOptConstants = preg_grep('/^CURLOPT_/', array_flip($constants['curl']));
+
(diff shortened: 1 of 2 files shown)


bennothommo:
@mahony0 In addition to your docs changes, would it be worth converting a string to a `CURLOPT_` constant by using `defined()` to see if the constant exists, and use `constant()` to extract the value, if a string key is given?

mahony0:
Sure, it seems pretty logical. I'll update the pull request

mahony0:
Sent PR. Also created an issue related to this library on: https://github.com/octobercms/october/issues/4624

src/Network/Http.php
```diff
@@ -520,7 +520,22 @@ public function setOption($option, $value = null)
...
+            $optionKey = constant($option);
+            $this->requestOptions[$optionKey] = $value;
+        } elseif (is_int($option)) {
+            $constants = get_defined_constants(true);
+            $curlOptConstants = preg_grep('/^CURLOPT_/', array_flip($constants['curl']));
+
+            if (isset($curlOptConstants[$option])) {
+                $this->requestOptions[$option] = $value;
+            } else {
+                throw new \Exception('Array keys must be CURLOPT constants or equivalent integer values');
```
bennothommo: @mahony0 Would you mind changing the Exception type to `October\Rain\Exception\ApplicationException`, to reflect that it is an implementation error by the developer?

bennothommo: In addition, would you also mind changing the message to the following?
```suggestion
                throw new \Exception('$option parameter must be a CURLOPT constant or equivalent integer');
```

src/Network/Http.php
```diff
@@ -520,7 +520,22 @@ public function setOption($option, $value = null)
...
+            }
+        } else {
+            throw new \Exception('Array keys must be CURLOPT constants or equivalent integer values');
```
bennothommo: @mahony0 Same as above.

LukeTowers:
@bennothommo could you add some tests for this?

PR #426, opened 2019-09-20
Regenerated resizer fixtures

JaxWilko:
Regenerated the fixtures for the resizer so that the tests don't fail.


bennothommo:
@JaxWilko if the tests are failing for you, then it's likely something different with your GD setup or development environment. The tests are not failing in our automated tests.

JaxWilko:
@bennothommo yeah i noticed that after I pushed, I'll take a look at my setup although I did test this on both Ubuntu 18.04 php7.2 & php7.3 and Ubuntu 19.04 php7.2 & php7.3 and had the same results.

I'll take a look at the travis report and see what the difference in set up is. I did talk to Luke about changing the fail thresh hold to account for machine to machine render differences, I think changing it to 2-2.5% resolves the issue, but again might just be one of those "it works on my machine" things :)

Thanks for your time!

bennothommo:
@JaxWilko Thanks for letting me know. FWIW, I use Ubuntu 18.04 as well, but my tests are running file. Images are just one of those things that are a bit finicky to test.

JaxWilko:
@bennothommo ah cool, good to know and 100% agree, even with small binary differences you can have big margins of difference in the end result.

PR #427, opened 2019-09-22
Improved .gif resizing in Resizer class

bennothommo:
In diagnosing some issues with GitHub Actions (similar to issues spotted by @JaxWilko in https://github.com/octobercms/library/pull/426), I investigated ways of improving transparent GIF resizing that would at least make the tests consistently pass.

After testing out the `imagescale()` method made available since PHP 5.5, it appears to not only consistently pass the tests, but it also appears to have a better image quality.

I've produced the two sets of images generated by the tests before and after the change:

**Source image:**
portrait

**Before changes (using imagecopyresampled)**
Screenshot from 2019-09-22 22-25-34

**After changes (using imagescale)**
Screenshot from 2019-09-22 22-26-25

This change is only applied for GIF images, all other formats use the original `imagecopyresampled` method as it still appears to beat `imagescale` for quality.


LukeTowers:
LGTM!

PR #428, opened 2019-09-24
Conform library files to our code style

bennothommo:
In preparation for rolling out code style checks with GitHub Actions.


tests/Parse/IniTest.php
```diff
@@ -244,7 +244,7 @@ public function testMultilinesValues()
         $vars = [
             'var' => "\\Test\\Path\\",
             'editorContent' =>
-'<p>Some
+        '<p>Some
```
LukeTowers: What's the reason for this change?

tests/Support/EmitterTest.php
```diff
@@ -102,17 +110,35 @@ public function testBindPriority()
...
+        }, 40);
+        $emitter->bindEvent('event.test', function () use (&$result) {
+            $result .= 'the ';
+        }, 30);
+        $emitter->bindEvent('event.test', function () use (&$result) {
+            $result .= 'lazy ';
+        }, 20);
+        $emitter->bindEvent('event.test', function () use (&$result) {
+            $result .= 'dog';
+        }, 10);
```
LukeTowers: hmm, the earlier version was definitely easier to read, is there anyway we can allow for this?

tests/fixtures/config/sample-config.php
```diff
@@ -15,7 +15,7 @@
 
     'debug' => true,
 
-    "debugAgain"  =>  FALSE ,
+    "debugAgain"  =>  false ,
```
LukeTowers:
```suggestion
    "debugAgain"  =>  false,
```
there's an extra space there.

bennothommo: @LukeTowers I have a feeling that the spacing there is intentional, as it is here:
https://github.com/octobercms/library/blob/4099148a026b3d11091a76056811124db2c31aa7/tests/fixtures/config/sample-config.php#L54
I would imagine it's testing to see that the config values are still read correctly even when someone mangles the spacing.

tests/Parse/IniTest.php
bennothommo: PHPCS sees the above as incorrect indentation. I've formatted the string content there to match the original spacing but also keep the correct indentation.

https://github.com/octobercms/library/pull/428/commits/c487ef37cbf84844c92aa46fc296ff69672ba93a

tests/fixtures/config/sample-config.php
LukeTowers: In that case we should leave it as FALSE and exclude this file from coding style checks @bennothommo

bennothommo: Done :)

LukeTowers:
@bennothommo is this good to go now?

bennothommo:
@LukeTowers yep, good to go :)

PR #431, opened 2019-09-27
Removed the clearing of CURLOPT_POSTFIELDS to allow for users to manually overwrite

JaxWilko:
The line `curl_setopt($curl, CURLOPT_POSTFIELDS, '');` removes the ability to manually set the curl opt `CURLOPT_POSTFIELDS`. 

There can be instances where the post fields of the curl may not be query string (for instance calling a soap API via curl).

The line removed will have no effect if `CURLOPT_POSTFIELDS` is not manually set. If somebody sets `CURLOPT_POSTFIELDS` and calls `$http->data(['foo' => 'bar'])` then the `data()` values will overwrite the `CURLOPT_POSTFIELDS` value. If no `data()` values are passed, then the `CURLOPT_POSTFIELDS` value will be used.

For these reason it seems to be safe and useful to make this change.

Please let me know if you have any questions.


bennothommo:
@JaxWilko Makes sense, and looks good to me. Would you be willing to add in some unit tests to https://github.com/octobercms/library/blob/develop/tests/Network/HttpTest.php to test the different configurations?

src/Network/Http.php
```diff
@@ -278,13 +277,11 @@ public function send()
...
         if ($this->requestData) {
-            $requestDataQuery = http_build_query($this->requestData, '', $this->argumentSeparator);
-
             if (in_array($this->method, [self::METHOD_POST, self::METHOD_PATCH, self::METHOD_PUT])) {
-                curl_setopt($curl, CURLOPT_POSTFIELDS, $requestDataQuery);
+                curl_setopt($curl, CURLOPT_POSTFIELDS, $this->getRequestData());
             }
             elseif ($this->method == self::METHOD_GET) {
-                curl_setopt($curl, CURLOPT_URL, $this->url . '?' . $requestDataQuery);
+                curl_setopt($curl, CURLOPT_URL, $this->url . '?' . $this->getRequestData());
```
bennothommo: @JaxWilko Sorry for the delay in getting back to you on this PR. Just one thing to note - this line here for GET requests should only accept request data provided through the proper method calls - it should not allow `CURLOPT_POSTFIELDS` to provide data, however, your new `getRequestData` method can potentially allow POSTFIELDS data through if the `requestData` array is empty.

PR #433, opened 2019-10-01
Clear classes compile cache

jacobdekeizer:
Remove the classes compile cache instead of the services cache.
The services cache gets cleared in the parent::handle().


LukeTowers:
@jacobdekeizer was there a bug that was triggered by this?

Samuell1:
Yes i had issue with this file not deleting on clearing cache.

jacobdekeizer:
The classes cache didn't get cleared which led to some errors when classes don't exist anymore.

PR #434, opened 2019-10-10
Add selectConcat method to query builder

bennothommo:
Adds a new `selectConcat` method to the October Rain query builder, that takes an array of concatenation parts and makes them aliased to a specific field in the model. This will use the correct type of concatentation based on the database type, ie. CONCAT() for MySQL, || for other types.


src/Database/Query/Grammars/Grammar.php
```diff
@@ -0,0 +1,94 @@
...
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $columns
+     * @return string|null
+     */
+    protected function compileColumns(Builder $query, $columns)
+    {
+        // If the query is actually performing an aggregating select, we will let that
+        // compiler handle the building of the select clauses, as it will need some
+        // more syntax that is best handled by that function to keep things neat.
+        if (! is_null($query->aggregate)) {
```
LukeTowers:
```suggestion
        if (!is_null($query->aggregate)) {
```

LukeTowers:
@bennothommo when this is polished up can you also submit a PR to Laravel? If they accept it into Laravel 6 then I'd like to just wait for us to move to 6 so that the maintenance burden is on them.

multiwebinc:
> Ready for review. The only thing I'm not sure of is this line:
> 
> https://github.com/octobercms/library/pull/434/files#diff-654762bea79e1fd19fbc793ef9990585R85
> 
> The idea is that column names should be wrapped in column quotes, and everything else is interpreted as a string, however, I couldn't find a method in the Builder scope that works out whether a given string is a column name. If anyone has any ideas on how that should be properly done, please feel free to point me in the correct direction.

This was what I was running into as well, which is why I was thinking quoted quotes. The problem with the current regex is that both strings and column names can contain practically any characters (including upper-case for column names) when wrapped correctly with quotes/backticks.

Here are 3 solutions I can think of:

- Match for `/^\'.*\'$/`. If the string starts and ends with a single quote, treat it as a string, or else treat it as a column.
- Use multi-dimensional arrays:

        $query = $model
            ->newQuery()
            ->select(['id'])
            ->selectConcat([
                [
                    'field',
                    'type' => 'identifier',
                ],
                [
                    ' ',
                    'type' => 'string',
                ],
                [
                    'cast'
                    'type' => 'identifier',
                ]
                'as' => 'full_cast'
            ]);

- Append each string with the type:

        $query = $model
            ->newQuery()
            ->select(['id'])
            ->selectConcat(['field|i', ' |s', 'cast|i'], 'full_cast');

Also, for both DB types, I would recommend quoting strings with single quotes and column names with backticks for MySQL and double quotes for SQLite. See [here](https://stackoverflow.com/questions/1992314/what-is-the-difference-between-single-and-double-quotes-in-sql).

tests/Database/SelectConcatTest.php
```diff
@@ -0,0 +1,53 @@
...
+            ->newQuery()
+            ->select(['id'])
+            ->selectConcat(['field', ' ', 'cast'], 'full_cast');
+
+        $this->assertEquals(
+            'select "id", "field" || " " || "cast" AS "full_cast" from "revisions"',
+            $query->toSql()
+        );
+    }
+}
```
multiwebinc, line 145: Consider testing strings and identifiers in upper/lower case, as well as identifiers using reserved keywords (e.g. `select "order" from tablename`).

bennothommo:
@LukeTowers do you want me to refactor our implementation and proceed with including it in the Rain library? I have a feeling that Laravel are not interested in it.

LukeTowers:
@bennothommo sure, but hold off on merging it just yet

LukeTowers:
@bennothommo could you add this to our DB docs?

bennothommo:
@LukeTowers done - https://github.com/octobercms/docs/commit/3d8ab788b355bd4a150d56113ac0ada6d7bcd863

PR #435, opened 2019-10-19
Force CSRF cookie to be decrypted

bennothommo:
Related PR: https://github.com/octobercms/october/pull/4701.

The CSRF token cookie has to be decrypted in order to be used by the JS framework. Instead of forcing October site devs to add the cookie to their `cookine.unencryptedCookies` config, this will force an override for just this token.


bennothommo:
Will be superceded by the work done in https://github.com/octobercms/october/pull/4720.

PR #436, opened 2019-10-21
Added event for translation helper function

SebastiaanKloos:
I've added an event to listen for translations. With this event you can manipulate the content before displaying it.

``` php
Event::listen('trans.beforeDisplay', function ($id, $result) {
    if ($id !== 'codecycler.test::lang.test') {
        return;
    }

    return 'Test';
});
```


LukeTowers:
There is already an event to accomplish this: https://octobercms.com/docs/api/translator/beforeresolve

SebastiaanKloos:
Allright didn’t know that. I’ll make sure I check the API docs before making a pr next time👍

PR #437, opened 2019-10-21
update Halcyon Model's function comment

wenlong-date:


wenlong-date:
https://github.com/octobercms/library/blob/4bda08c826fc3ac1efba32fa8799d25ced14191f/src/Halcyon/Model.php#L536-L551
https://github.com/octobercms/library/blob/4bda08c826fc3ac1efba32fa8799d25ced14191f/src/Halcyon/Model.php#L1342-L1353

Maybe the return comment of function on should be `\October\Rain\Halcyon\Model` 
and function newQuery should be `\October\Rain\Halcyon\Builder` ?

PR #438, opened 2019-10-21
Prevent infinite loop with grouped repeater fields in Syntax Parser

grahambrown11:
Fixes `Maximum function nesting` error by checking for null istead of PHPs boolean automagic - an empty array is considered false


LukeTowers:
@grahambrown11 can you walk us through the bug that's causing the problem please?

grahambrown11:
@LukeTowers the bug I logged on the October repo kinda explains the issue, however here is some more unpacked...
We just started using the `Campaign Manager` email plugin from Responsiv (@daftspunk) and I wanted to reuse some of the definitions we created for the `Static Pages` plugin.
We use the variable definition with the repeater type:
`{variable name="sections" type="repeater" prompt="Add Section" groups="~/themes/mailer/sections.yaml"}`
I see the `Campaign Manager` `Message` class uses the `SyntaxModelTrait` and it calls the `makeSyntaxFields` function which calls `getFieldValues` on the `Parser` which calls `getDefaultParams` on the `FieldParser`, the `$fields` param is not supplied so it uses `$this->fields` then loops over the fields and when it gets to the `repeater` type is calls itself with an empty array as there is no `fields` in that type of repeater definition. so `!$fields` when `$fields = []` evaluates to false and instead of returning an empty `$defaults` it uses `$this->fields` again and does the same loop and calls itself until eventually the heap is too big and PHP throws a `Maximum function nesting` error. 

The simple change from `!$fields` to `is_null($fields)` fixes this issue, I applied the patch to my dev environment and I can now use `{variable name="sections" type="repeater" prompt="Add Section" groups="~/themes/mailer/sections.yaml"}` and no more errors

LukeTowers:
@w20k can you confirm this issue and fix, and test to ensure that it doesn't break anything else?

w20k:
@LukeTowers, sure, will do!

daftspunk:
LGTM, thanks!

PR #439, opened 2019-10-23
Move registering of routes to boot method of ServiceProvider

Samuell1:
Related PR: https://github.com/octobercms/october/pull/4714

And note from laravel docs - https://laravel.com/docs/5.5/providers#writing-service-providers
> You should never attempt to register any event listeners, routes, or any other piece of functionality within the register method.


daftspunk:
There doesn't seem to be any reason to change the location of this registration logic. Even though Laravel says its a bad idea and the reasoning makes sense, it has existed this way for a long time. This will need testing to ensure it doesn't break anything

Samuell1:
@daftspunk Yes, i think best way will be to update this with L6 upgrade.

daftspunk:
Fixed in v2+

PR #440, opened 2019-11-02
No longer serialize cookies

daftspunk:
This is based on a security patch that was released by Laravel. We implemented their workaround for the time being, this fixes it properly by disabling cookie serialization all together.

Ref: https://laravel.com/docs/5.6/upgrade#upgrade-5.6.30

> Since this vulnerability is not able to be exploited without access to your application's encryption key, we have chosen to provide a way to re-enable encrypted cookie serialization while you make your application compatible with these changes.


daftspunk:
It is very important to note the `Cookie::set` method no longer accepts anything but a scalar value as the first argument.

```
// This is OK
Cookie::set('string')

// This is NOT OK anymore
Cookie::set(['array', 'array']);
```

PR #442, opened 2019-11-10
Prevent double saving of models in relation manager widget.

bennothommo:
In some cases the relationship has already been associated outside this workflow, this change prevents the save events from triggering twice. This works because it mimics Laravel's internals, where if nothing changes on a model, the database is never engaged, otherwise it would produce a nil update query

Replaces https://github.com/octobercms/october/pull/4723. Fixes https://github.com/octobercms/october/issues/4246

### DIFF
--- a/src/Database/Relations/HasOneOrMany.php
+++ b/src/Database/Relations/HasOneOrMany.php
@@ -57,7 +57,10 @@ trait HasOneOrMany
     {
         if ($sessionKey === null) {
             $model->setAttribute($this->getForeignKeyName(), $this->getParentKey());
-            $model->save();
+
+            if (!$model->exists || $model->isDirty()) {
+                $model->save();
+            }
 
             /*
              * Use the opportunity to set the relation in memory


src/Database/Relations/HasOneOrMany.php
```diff
@@ -57,7 +57,10 @@ public function add(Model $model, $sessionKey = null)
     {
         if ($sessionKey === null) {
             $model->setAttribute($this->getForeignKeyName(), $this->getParentKey());
-            $model->save();
+
+            if (!$model->exists || $model->isDirty()) {
```
LukeTowers: @bennothommo @daftspunk wouldn't this prevent existing models from being added to the relationship?

```php
$comment = new Comment();
$comment->save();

$post->add($comment);
```

bennothommo: @LukeTowers I haven't tested that particular case, but I think the line above it, setting the related attribute to the new parent model, would make the child model dirty, so it would still save.

LukeTowers: Ah, right. I got confused by thinking that `||` was being processed like `&&`, where it would short circuit if the first value was false, but it's actually the opposite so this is fine. Thanks for the response!

PR #443, opened 2019-11-22
[Laravel 6] Bring code to a state where unit tests pass

bennothommo:
This PR brings the state of the Library code, running Laravel 6 as the foundation, so that the unit tests pass. Whilst it is definitely no guarantee that the code will work with Laravel 6, it's at least a good start. This work forms part of the work towards https://github.com/octobercms/october/issues/4381.

### Changes made:
- Now using Laravel 6 as the foundation framework for the library.
- Now using PHPUnit 8 as opposed to 7, as Laravel 6 uses [PHPUnit 8](https://github.com/laravel/framework/blob/6.x/composer.json#L88) as a dev dependency, and it will also be supported for longer. Refactored several tests to stop using deprecated assertions, which should also theoretically make them compatible with PHPUnit 9.
- Now using [maintainer-run Assetic library repo](https://github.com/assetic-php/assetic/) for bringing in Assetic library and supporting the latest Symfony Process library. Temporarily using an explicit repository definition in `composer.json` whilst the library is still unpublished on Packagist.
- **Breaking change:** Changed `October\Rain\Halcyon\MemoryRepository` to use seconds instead of minutes, as per changes made to `Illuminate\Cache\Repository`. I didn't see this being used anywhere in the library, but I imagine this will likely result in needing changes done to October CMS core.
- Changed some `Input` facade calls to comparable `Request` facade calls, as the `Input` facade is no longer available in Laravel 6 (although we are making the facade available in October still).


bennothommo:
Submitted a PR to **meyfa/phpunit-assert-gd** to make the library compatible with PHPUnit 8, so we don't have to rewrite the Resizer unit tests. This should resolve the unit test errors reported by Actions.

https://github.com/meyfa/phpunit-assert-gd/pull/2

LukeTowers:
@bennothommo I've sent an email to @meyfa asking if they're interested in merging that PR. If they aren't we'll just fork it over to @octoberrain and use that instead.

src/Database/Concerns/HasRelationships.php
```diff
@@ -416,17 +429,17 @@ public function belongsTo($related, $foreignKey = null, $parentKey = null, $rela
...
      */
-    public function morphTo($name = null, $type = null, $id = null)
+    public function morphTo($name = null, $type = null, $id = null, $ownerKey = null)
     {
         if (is_null($name)) {
             $name = $this->getRelationCaller();
         }
 
-        list($type, $id) = $this->getMorphs(Str::snake($name), $type, $id);
+        [$type, $id] = $this->getMorphs(Str::snake($name), $type, $id);
```
LukeTowers: @bennothommo what's the reason for this change?

src/Parse/Assetic/LessCompiler.php
```diff
@@ -1,13 +1,12 @@
 <?php namespace October\Rain\Parse\Assetic;
 
-use Event;
```
LukeTowers: @bennothommo @jaxwilko @christianWilling  is there value in moving this filter to the Assetic core? I'm not 100% sure what it does differently from the already included Assetic filters.

src/Parse/Assetic/ScssCompiler.php
```diff
@@ -1,12 +1,11 @@
 <?php namespace October\Rain\Parse\Assetic;
 
 use Event;
```
LukeTowers: @bennothommo @JaxWilko @christianWilling is there value in moving this filter to the Assetic core? I'm not 100% sure what it does differently from the already included Assetic filters.

@bennothommo also note that the Event facade is included but not used here.

src/Database/Concerns/HasRelationships.php
bennothommo: @LukeTowers Is "habit" an acceptable answer? :P

src/Parse/Assetic/LessCompiler.php
JaxWilko: From the looks of it the October one also add's in file version hashing functionality, could merge it in this filter or add in that functionality as an option for the Assetic project?

src/Database/Concerns/HasRelationships.php
LukeTowers: @bennothommo only if you can back it up with a reason 😛

src/Parse/Assetic/LessCompiler.php
LukeTowers: I haven't spent enough time on this filter or the core Assetic filter to understand the difference between those two options @jaxwilko, are you able to explain it in a bit more detail for me?

src/Database/Concerns/HasRelationships.php
bennothommo: @LukeTowers It's just the more "modern" way of defining a `list()` call. I probably did it when I went through that class, but I can change it back. It's purely cosmetic.

LukeTowers: I think I'd prefer the list call, at least with that if I don't understand what's happening I can google the function name and figure out how the list() magic works.

LukeTowers:
@bennothommo note that the assetic package is on packagist: https://packagist.org/packages/assetic/framework#dev-2.0/dev

bennothommo:
@LukeTowers sweet, will drop the repository definition in `composer.json`.

meyfa:
@LukeTowers FYI, I've released v2.0.0 of phpunit-assert-gd with the proposed changes.

LukeTowers:
@meyfa awesome, thanks very much for that! @bennothommo do you need to update the constraint in composer?

bennothommo:
Thanks heaps @meyfa. Yep, will adjust `composer.json` accordingly, @LukeTowers :)

bennothommo:
@LukeTowers Input facade has been re-added.

composer.json
```diff
@@ -22,11 +22,11 @@
         "php": "^7.2",
         "doctrine/dbal": "^2.6",
         "erusev/parsedown-extra": "~0.7",
-        "kriswallsmith/assetic": "~1.3",
+        "assetic/framework": "dev-2.0/dev",
         "patchwork/jsqueeze": "~2.0",
         "oyejorge/less.php": "~1.7",
         "leafo/scssphp": "~0.6",
-        "symfony/yaml": "^4.0",
+        "symfony/yaml": "^3.4",
```
LukeTowers: Why was this change required?

bennothommo: @LukeTowers Symfony/Yaml 4.0 and above do not allow numeric keys for arrays, which basically means our current plugin version files will no longer work. See here: https://github.com/symfony/yaml/blame/1ff992f3a74b4c77fea5fd9f382a083915c214aa/Parser.php#L215
They deprecated it in 3.4.0 as shown here: https://github.com/symfony/yaml/commit/240ab5d231719d40a20d8433914b9ea27981624a

composer.json
```diff
@@ -35,20 +33,20 @@
...
+        "illuminate/queue": "^6.11",
+        "illuminate/support": "^6.11",
+        "illuminate/log": "^6.11",
+        "illuminate/mail": "^6.11",
+        "illuminate/events": "^6.11",
+        "illuminate/pagination": "^6.11",
+        "illuminate/routing": "^6.11",
+        "illuminate/validation": "^6.11",
+        "illuminate/view": "^6.11",
+        "illuminate/cache": "^6.11",
```
LukeTowers: @bennothommo isn't it 6.0 that's the LTS?

bennothommo: @LukeTowers the entire v6 line is LTS, because of their new semantic versioning policy. I'm  trying to ensure we're using the latest release and testing against that.

src/Parse/Assetic/LessCompiler.php
LukeTowers: @JaxWilko ping

PR #444, opened 2019-11-22
Allow replacements to be made for unregistered strings

multiwebinc:
If you do `trans("hello :where", ['where' => 'world'])`, the `:where` parameter is only substituted if the `hello :where` string is registered in one of the language files. If not, parameter substitutions are not made.

### DIFF
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -138,7 +138,7 @@ class Translator implements TranslatorContract
         // that will be quick to spot in the UI if language keys are wrong or missing
         // from the application's language files. Otherwise we can return the line.
         if (!isset($line)) {
-            return $key;
+            return $this->makeReplacements($key, $replace);
         }
 
         return $line;


LukeTowers:
Why would we support unregistered strings?

multiwebinc:
@LukeTowers They already are supported in the sense that `trans('Some unregistered string')` will return the original string and doesn't produce any error. I personally find it easier to develop everything first and then move everything to translation files last, so I would see this PR as being more useful during development.

mjauvin:
interesting point

LukeTowers:
@multiwebinc I wouldn't really call that support, at the very least not intentional support. To be honest, I'm not sure we want to support that behaviour as it encourages improper use of the translation functions that can easily make it into production. I'll discuss it with @daftspunk though and get his input.

multiwebinc:
@LukeTowers If you decide *not* to implement this PR, maybe it would be better to go the other way and completely remove any support for unregistered strings, like making it throw an exception, return an empty string, or trigger an error, like E_DEPRECATED. I think it should either be fully supported (feature-wise), or not at all.

daftspunk:
This is a good idea

PR #445, opened 2019-11-24
add event documentation comments;

mjauvin:
- fix wrong argument provided to fireEvent() call ($attr -> $value)
- add event documentation comments

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1009,6 +1009,8 @@ class Model extends EloquentModel
          * @event model.getAttribute
          * Called after the model attribute is retrieved
          *
+         * Note: also triggered in October\Rain\Halcyon\Model
+         *
          * Example usage:
          *
          *     $model->bindEvent('model.getAttribute', function ((string) $key, $value) use (\October\Rain\Database\Model $model) {
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -178,6 +178,7 @@ class Model extends EloquentModel
             /**
              * @event model.afterBoot
              * Called after the model is booted
+             * > **Note:** also triggered in October\Rain\Halcyon\Model
              *
              * Example usage:
              *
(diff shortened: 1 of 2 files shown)


LukeTowers:
@mjauvin take a look at https://github.com/octobercms/library/blob/master/src/Database/Model.php#L974-L1024. That's what this is supposed to be mirroring. $attr is valid in that one. Specifically notice that these are local events, and actually we can't have the same event key here. We should just change the eloquent model's ones to note that they're also fired on Halcyon models and add a comment as such to the halcyon model events without actually adding a full event docblock.

mjauvin:
Definitely local events, my bad.

And by "We should just change the eloquent model's ones", I assume you mean change the *docblocks* (i.e. add a note to reflect the event is also fired on the Halcyon model, and vice-versa) ?

LukeTowers:
@mjauvin that's correct, change the docblocks to reflect the fact they're also fired on Halcyon models

mjauvin:
Is that acceptable?

src/Database/Model.php
```diff
@@ -1009,6 +1009,8 @@ public function getAttributeValue($key)
          * @event model.getAttribute
          * Called after the model attribute is retrieved
          *
+         * Note: also triggered in October\Rain\Halcyon\Model
```
LukeTowers: Should add this to all events here that are fired in the Halcyon model

mjauvin: Yes, but I wanted to know first if this is how you want it to appear.

LukeTowers: Gotcha. LGTM!

daftspunk:
The fix makes sense. The documentation is a little confusing, why does Halycon need to know about Database and vice versa?

mjauvin:
@daftspunk this is because the API doc will only document one or the other since they both use the same event id. So we added the references

daftspunk:
I see, thanks. I think the documentation was written with the intent to cover only global events. There might be something we can do to namespace the local ones, or include them in the class definitions. This should do for now, though!

mjauvin:
@daftspunk that would be a good thing to add, yes!

daftspunk:
Made a note of it 😉

PR #446, opened 2019-11-28
Allow using custom or built-in Laravel/Symfony error responses

marcogrueter:
Fixes https://github.com/octobercms/october/issues/3416.

This PR allows to use fully custom responses (Symfony\Component\HttpFoundation\Response) from custom error handlers and fixes the bug mentioned in octobercms/october#3416 and octobercms/october#3127  also allows throwing other Exceptions.

While this is not directly described as a possibility in the docs (https://octobercms.com/docs/services/error-log#exception-handling), where the only example is returning a string which works regardless, I think it's a good thing and also closes octobercms/october#3416 and octobercms/october#3127, which is a bug that occurs when using an error response method that throws an exception.

Checking if the response from a custom handler is a "Response" object fixes the effect that the headers of the first response gets rendered into the content.

### DIFF
--- a/src/Foundation/Exception/Handler.php
+++ b/src/Foundation/Exception/Handler.php
@@ -68,6 +68,10 @@ class Handler extends ExceptionHandler
         $response = $this->callCustomHandlers($exception);
 
         if (!is_null($response)) {
+            if ($response instanceof \Symfony\Component\HttpFoundation\Response) {
+                return $response;
+            }
+
             return Response::make($response, $statusCode);
         }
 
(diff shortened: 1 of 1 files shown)


bennothommo:
Thanks @marcogrueter - would you mind fixing the code quality issues reported here?
https://github.com/octobercms/library/pull/446/checks?check_run_id=324807088#step:5:8

marcogrueter:
Thanks for checking, all done!

daftspunk:
Fix for what appears to be a long-standing bug. Thanks!

PR #447, opened 2019-11-28
Trim the list of available artisan commands

marcogrueter:
Trim the list of available artisan commands down to those that work.
Commands that are not listed in the array will not be registered,
can not be used on the command and will not appear on the command line when using 'php artisan'.

This PR is a part of an answer to https://github.com/octobercms/docs/issues/394.
I will also submit a small PR for the docs repo, that the 'make' commands from laravel are not supported.

List of disabled command for reference:

* auth
  * clear-resets
* db
  * seed
* event
  * geenerate
* cache
  * table
* make
  * auth
  * command
  * event
  * exception
  * factory
  * job
  * listener
  * mail
  * middleware
  * migrate
  * model
  * notification
  * policy
  * provider
  * request
  * resource
  * rule
  * seeder
  * test
* migrate
  * fresh
  * install
  * refresh
  * reset
  * rollback
  * status
* notifications
  * table
* preset
* queue
  * table
  * failed-table
* session
  * table
* storage
  * link

### DIFF
--- a/src/Foundation/Providers/ArtisanServiceProvider.php
+++ b/src/Foundation/Providers/ArtisanServiceProvider.php
@@ -7,6 +7,50 @@ use Illuminate\Foundation\Providers\ArtisanServiceProvider as ArtisanServiceProv
 class ArtisanServiceProvider extends ArtisanServiceProviderBase
 {
     /**
+     * The commands to be registered.
+     *
+     * @var array
+     */
+    protected $commands = [
+        'CacheClear'      => 'command.cache.clear',
+        'CacheForget'     => 'command.cache.forget',
+        'ClearCompiled'   => 'command.clear-compiled',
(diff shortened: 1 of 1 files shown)


marcogrueter:
image

List of available commands for reference

Samuell1:
Looks good, are tables for session, queue, failed queue and cache created by default? Maybe they should stay?

marcogrueter:
Migrations for those tables are all in https://github.com/octobercms/october/tree/master/modules/system/database/migrations
and the commands use the default migration files from laravel-framework. So it should be safe to remove them.

Samuell1:
@marcogrueter amazing, i think this is okay, hope nothing uses this commands.

I doesnt need to be removed when laravel removes it when ocms gets updated.

daftspunk:
Good idea

PR #451, opened 2020-01-06
Make translator implementation compatible with translator interface

jacobdekeizer:
The current implementation of the translator is not compatible with the interface.

The translator interface requires 3 parameters for method `trans`

```php
public function trans($key, array $replace = [], $locale = null);
```

The current implementation has 4 parameters.
```php
public function trans($id, array $parameters = [], $domain = 'messages', $locale = null)
{
     return $this->get($id, $parameters, $locale);
}
```

If I resolve the translator interface I can't set the locale, because it will set the domain parameter (which isn't used) instead of the locale parameter.

For example this will not work:
```php
/** @var Illuminate\Contracts\Translation\Translator $translator */
$translator = resolve(Illuminate\Contracts\Translation\Translator::class);
$translator->trans('my_key', [], 'nl');
```

The same goes for the method `transChoice`

### DIFF
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -245,11 +245,10 @@ class Translator implements TranslatorContract
      *
      * @param  string  $id
      * @param  array   $parameters
-     * @param  string  $domain
      * @param  string  $locale
      * @return string
      */
-    public function trans($id, array $parameters = [], $domain = 'messages', $locale = null)
+    public function trans($id, array $parameters = [], $locale = null)
     {
         return $this->get($id, $parameters, $locale);
     }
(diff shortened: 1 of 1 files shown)


bennothommo:
@jacobdekeizer It looks like that was a leftover from when the Translator extended Symfony's Translator.

I haven't seen any `trans` or `transChoice` calls in core where they even use the third or fourth parameters so I think this is safe to include, but I'll let @LukeTowers weigh in just in case there's an edge case I haven't thought of.

daftspunk:
Can confirm this is [where I missed a spot](https://github.com/laravel/framework/blob/5.1/src/Illuminate/Translation/Translator.php#L198) in the Laravel 5.1 -> 5.5 upgrade

@jacobdekeizer go easy on this fix, it's an antique! :joy:

mjauvin:
This is breaking backend translations. This needs to be reverted ASAP.

acasar:
Yeah, it broke a lot of my code as well and it took me lots of time to find this commit...

There should really be some better process of introducing breaking changes like that.

LukeTowers:
@acasar this should be fixed in 464.

PR #452, opened 2020-01-09
Guess extension for uploaded files if missing

aurelien-roy:
This change aims to guess file extension from MIME type if file is being uploaded without file extension. The guessed extension is then added to the generated filename.

Some contextual explanation: 

**Why would a file be uploaded without extension?**

It's the case for raw blob files. For instance, in my app I let the user choose a image file and then allow her/him to do some basic editing (like cropping) client-side. Thus, when uploading, I'm no longer uploading the raw file, but a blob file that don't have a filename and thus doesn't have any extension.

**Why storing file without extension is an issue?**

This can break other October features. For instance, the issue I had when uploading blob images, was that the getThumb function silently failed (because it didn't consider images without extension as images, and thus do not resize them). So typically, I was serving high-res image on my website just to display them as thumbnail and didn't notice it for a long time.

You could argue that a solution is to change my client-side code to manually give filename with extension to the blob. The thing is, even on a regular form input field, a user may attempt to upload an image without file extension (for whatever reason), that successfully passes the "image" validation rule (because this rule only check MIME type), but won't work with the getThumb function, which is counter-intuitive.

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -703,7 +703,7 @@ class File extends Model
         $ext = strtolower($this->getExtension());
         
         // If file was uploaded without extension, attempt to guess it
-        if(!$ext && $this->data instanceof UploadedFile) {
+        if (!$ext && $this->data instanceof UploadedFile) {
             $ext = $this->data->guessExtension();
         }
         
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -701,6 +701,12 @@ class File extends Model
         }
 
         $ext = strtolower($this->getExtension());
+        
+        // If file was uploaded without extension, attempt to guess it
+        if(!$ext && $this->data instanceof UploadedFile) {
+            $ext = $this->data->guessExtension();
+        }
+        
         $name = str_replace('.', '', uniqid(null, true));
 
         return $this->disk_name = !empty($ext) ? $name.'.'.$ext : $name;


36864:
There would need to be a way to opt out of this behavior.

aurelien-roy:
Any example example of why this behavior may be undesirable?
It's only happening on the function that generates a hash filename. So in any case, the original file name is lost. It won't happen on code that are preserving the filename for instance.

36864:
Would this not break existing files?

aurelien-roy:
No, this is an additional check happening when generating filename for newly uploaded files.
It doesn't affect already uploaded files.

36864:
My mistake then, I misunderstood the usage. Thanks for clarifying.

bennothommo:
@aurelien-roy I think, in the context you have provided, I would prefer a fix to `getThumb()` over trying to apply extensions to files programatically through MIME detection. There is a *potential* security issue here where someone forges a MIME type in a file, which your code would then assign it an extension which might be considered safe on the server, or in a plugin, even though it's not. Thoughts?

bennothommo:
I also think that in the normal course, people uploading blob files with no extensions is fairly low.

aurelien-roy:
@bennothommo 
Well sure a fix to getThumb would do the trick. However:

- I think this kind of extension auto-assignation is part of a input sanitization process, that could prevent other types of problems.
- Vanilla Laravel does this! See the function `hashName` in from Laravel FileHelpers uses `guessExtension` to generate a filename ([source](https://github.com/laravel/framework/blob/5eff676525f66773582f9e5b1f50f8716fdaaa29/src/Illuminate/Http/FileHelpers.php#L50)) and  it is that hashName function which it used to store uploaded file ([source](https://github.com/laravel/framework/blob/5eff676525f66773582f9e5b1f50f8716fdaaa29/src/Illuminate/Http/UploadedFile.php#L36))).
- I can't think how this could be a security issue. In any case, any kind of input should always be treated with cautious. (And I think this behavior wouldn't be in Laravel if there was a problem)

aurelien-roy:
Any comment on this?

LukeTowers:
@aurelien-roy where does `$this->data` get set?

aurelien-roy:
@LukeTowers 
On a File model creation, within the Attach / AttachMany relations:
https://github.com/octobercms/library/blob/7f8bc342bf805aaa0bf6fc1f4d22eaec04823932/src/Database/Relations/AttachOne.php#L43

aurelien-roy:
@LukeTowers 
Thanks for reviewing and merging :)
Also, what do you think of the « always rely on MIME approach » discussed [here](https://github.com/octobercms/library/pull/452#issuecomment-575750569) to be more consistent with Laravel behavior?

In short: With that PR, we use MIME to guess extension only when extension is not present in the uploaded file name, whereas in Laravel MIME is always used to determine extension (extension in file name is ignored).

LukeTowers:
@aurelien-roy I'm not sure about it tbh, there's a lot of things to consider. For one thing, MIME type guessing is not perfect, there are mime types that map to multiple extensions and there are files that MIME type guessing can't even come up with a guess for. 

The main concerns when it comes to files is preventing malicious files from being served or executed by the server or browsers. Since server software generally relies on file extensions to determine how to process a file (whether to execute it in the case of .php or just deliver it unmodified to the client as with .css, image files, etc). In that sense, we really just need to lock down file upload to only known safe extensions (which October does right now).

If we were to auto detect all uploads using that method, many classes of files would stop working entirely (that method returns null for a lot of plain text file types) and additionally we might even be opening ourselves up to future theoretical vulnerabilities where an image passes the initial extension check for upload but then helpfully gets auto renamed with a malicious extension by that method.

aurelien-roy:
Sorry for the delayed answer.

> In that sense, we really just need to lock down file upload to only known safe extensions (which October does right now).

I think this is wrong. Currently validation rule like « image » rely exclusively on the MIME type. That means, that you can make an image, and change its extension to « .exe » and it will uploads and validates without issue. (Just tested it).

> whether to execute it in the case of .php or just deliver it unmodified to the client as with .css, image files, etc

There is already a check for PHP file built in to MIME type validation rule that does that. But in any case, the provided .htaccess file blocks the execution of all php files except the « index.php ».

> If we were to auto detect all uploads using that method, many classes of files would stop working entirely (that method returns null for a lot of plain text file types)

In the current version of Laravel, there is a null check on guessExtension when naming a file. That check is not present on the version of Laravel currently used by October. In any case, this is a non-issue, because null is always casted to empty string when concatenating strings. So the only bad thing that could happen doing `$filename . "." . $extension` is having an useless trailing dot on filenames.


So I don’t think there is something different in October that would justify to change the default Laravel behavior. Plus, since all validation rules work with MIME file, for me it is safer to rename the file to match the MIME type that have been verified. Otherwise, you can end up with « .exe » files that passes image validation (with all the security implications it could potentially lead).

PR #455, opened 2020-01-16
Improved Rule object handling in Validator

### LINKED ISSUE #4783: When using Rule objects, the error message is not used
I am trying to add a rule to validate a reCAPTCHA response, and I copied the code under the section "Using Rule objects" from https://octobercms.com/docs/services/validation#custom-validation-rules. I can tell that the `validate()` and `passes()` methods are being executed because when I do `dd()` inside these messages, it works. However this is not so with the `message()` method in that class, which is not being executed at all. In the ajax response, it has a generic error message:

    g-recaptcha-response: ["validation.recaptcha"]

instead of using the custom error message.

bennothommo:
Fixes https://github.com/octobercms/october/issues/4783, replaces #365 and #424.

The Rule object support in the Validator is currently lacking a bit of finesse. A developer must define a `validate`, a `message` and a `passes` method to satisfy the Validator as well as the contract, but currently, the only method that *really* gets used is the `validate` method when following our [docs for custom validation rules](https://octobercms.com/docs/services/validation#custom-validation-rules). The `message` method is never used to determine a fallback message, requiring a user to define a custom replacer just to get a message across, or define the message manually everywhere where the rule is used.

This PR tackles a few things:
- It defines a new base Rule class in October Rain, which will cover the `validate` method, allowing a developer to simply define the `passes` method which will determine if a given value passes validation (of course, a developer is free to overwrite the `validate` method if they wish)
- It allows Rule objects to be used as Validator extensions correctly, and to be used when defining rules for models.
- Finally, it allows the `message` method of a Rule object to be used as the fallback message, whilst allowing developers to override the message if need be.

This change does replace Laravel's service provider for Validation with our own, however, for the most part our pipeline is just a carbon copy of Laravel's - so it should be backwards compatible.

### Still to do:
- [x] Unit tests
- [x] Update docs (https://github.com/octobercms/docs/pull/473)


bennothommo:
For example, the Reserved rule in use for the Builder plugin (https://github.com/rainlab/builder-plugin/blob/master/rules/Reserved.php) would now be able to be condensed to the following:

```php
<?php namespace RainLab\Builder\Rules;

use Lang;
use October\Rain\Validation\Rule;

class Reserved implements Rule
{
    protected $reserved = [
        // ....
    ];

    public function passes($attribute, $value)
    {
        return !in_array(strtolower($value), $this->reserved);
    }

    public function message()
    {
        return Lang::get('rainlab.builder::lang.validation.reserved');
    }
}
```

bennothommo:
Just to note, this is now targeting the Laravel 6 update, as we introduced the Validator service in https://github.com/octobercms/library/commit/97d4163041f68c23f6deb28832d860463bb233ee, which this PR was also going to do.

bennothommo:
@LukeTowers ready for final review.

LukeTowers (approved):
LGTM!

PR #456, opened 2020-01-20
Rebase October Rain library on Laravel 6

bennothommo:
Please see https://github.com/octobercms/october/pull/4893 for more information.

This PR represents the current state of the Rain library rebase to Laravel 6.


composer.json
```diff
@@ -12,38 +12,42 @@
...
-        "leafo/scssphp": "~0.6",
-        "symfony/yaml": "~3.2",
+        "symfony/yaml": "^3.4",
         "twig/twig": "~2.0",
         "league/csv": "~8.0",
         "jenssegers/date": "~3.0",
-        "laravel/tinker": "~1.0"
+        "laravel/tinker": "^1.0",
+        "laravel/helpers": "^1.0",
+        "fideloper/proxy": "^4.0"
```
LukeTowers: What is this dependency used for?

src/Halcyon/MemoryRepository.php
```diff
@@ -47,18 +47,18 @@ public function get($key, $default = null)
      *
      * @param  string  $key
      * @param  mixed   $value
-     * @param  \DateTimeInterface|\DateInterval|float|int  $minutes
+     * @param  \DateTimeInterface|\DateInterval|int  $seconds
      * @return void
      */
-    public function put($key, $value, $minutes = null)
+    public function put($key, $value, $seconds = null)
```
LukeTowers, line 53: Is there anywhere we need to ensure is updated to pass seconds instead of minutes to this?

src/Support/Facade.php
```diff
@@ -19,6 +19,7 @@ protected static function resolveFacadeInstance($name)
     {
         if (
             !is_object($name) &&
+            !is_null(static::$app) &&
```
LukeTowers, line 22: Why was this required?

LukeTowers (dismissed):

composer.json
bennothommo: @LukeTowers it's a trusted proxy library. It's required for Laravel 5.6, but it's only required by `laravel/laravel`, even though the framework uses it. So I had to make it a dependency for our library too.

src/Halcyon/MemoryRepository.php
bennothommo: @LukeTowers very likely, but that's still a task to complete in the list.

src/Support/Facade.php
bennothommo: @LukeTowers some unit tests fail without this, as no application is available (I believe only the plugin test cases initiate an application).

composer.json
LukeTowers: @bennothommo could you point me to where it's required by the framework?

bennothommo: @LukeTowers I was mistaken - for some reason, in my initial tests last year, it was erroring out if `fideloper/proxy` was missing. When I've done a fresh install, everything works now. The framework doesn't use it - only the `laravel/laravel` package does.

composer.json
```diff
@@ -12,38 +12,40 @@
...
+        "illuminate/support": "^6.0",
+        "illuminate/log": "^6.0",
+        "illuminate/mail": "^6.0",
+        "illuminate/events": "^6.0",
+        "illuminate/pagination": "^6.0",
+        "illuminate/routing": "^6.0",
+        "illuminate/validation": "^6.0",
+        "illuminate/view": "^6.0",
+        "illuminate/cache": "^6.0",
+        "meyfa/phpunit-assert-gd": "^2.0.0",
```
LukeTowers: Do you mind grouping all the PHPUnit deps next to each other? Additionally, why do we include all these laravel libraries manually if we include laravel/framework in the main project?

bennothommo: @LukeTowers my understanding is that the framework libraries are there so the library can be tested in isolation.

LukeTowers: @bennothommo fair enough, wasn't sure if the main project was being pulled in to run these tests or not

bennothommo: @LukeTowers not with the automated tests on this repo, no. I think @daftspunk has mentioned in the past that the library is sometimes used externally to October.

src/Database/Concerns/HasRelationships.php
```diff
@@ -299,6 +321,7 @@ protected function handleRelation($relationName)
                 $relationObj = $this->$relationType($relation[0], $relation['public'], $relation['key'], $relationName);
                 break;
 
+            case 'hasOneThrough':
```
LukeTowers, line 324: Was this added to the database docs?

bennothommo: @LukeTowers yep :) - https://github.com/octobercms/docs/blob/wip/laravel-6/database-relations.md#has-one-through

LukeTowers: Excellent, thanks!

src/Auth/Models/User.php
```diff
@@ -25,8 +25,8 @@ class User extends Model implements \Illuminate\Contracts\Auth\Authenticatable
      */
     public $rules = [
         'email' => 'required|between:3,255|email|unique:users',
-        'password' => 'required:create|min:2|confirmed',
-        'password_confirmation' => 'required_with:password|min:2'
+        'password' => 'required:create|min:4|confirmed',
+        'password_confirmation' => 'required_with:password|min:4'
```
bennothommo, line 29: @LukeTowers I'm not quite sure the reasoning that @daftspunk changed this back to `2` (I presume for backwards compatibility), but if we are going to change it, I personally would rather this be `min:6` or `min:8` at the least. We should be following best practices for password selection, and 4 character passwords are too easy to brute-force in this day and age. (I understand we have throttling, but it can be disabled).

LukeTowers: @bennothommo I checked the git history and the value was originally 4, so I don't know why @daftspunk felt the need to bump it down to 2. I would also prefer the minimum be set to 8, but I believe we had a conversation about this a while ago and it was decided that people could always just bump up the rules themselves. Additionally, this is somewhat tied into the autogeneration of admin passwords PR, was that one ever finished?

bennothommo: @LukeTowers I do recall that conversation. On further consideration though, I feel if anything we should be bumping it high and forcing people to work to make the value lower, so that they are fully responsible for the consequences.

Re: auto-generation of admin passwords, it's more or less [finished](https://github.com/octobercms/october/pull/4866), but on hold until after L6.

LukeTowers: I'll have another convo with @daftspunk about it and get back to you

LukeTowers: Spoke to Sam and he re-iterated that
> it's best to leave it as 4 otherwise it's a breaking change. Password policies are a complex topic and so it's best not to have any strong opinions and leave it up to a plugin or attend to the entire problem holistically
> there are legit applications that use short passwords
> Namely demo apps and temporary passwords
> A config item would probably appease whatever is driving this. Default without a breakage

So we could add a new config item in `auth.minPasswordLength` that defaults to `8` in the file but in the Config::get() it defaults to the existing 4. That way existing installs won't break, and new ones will default to 8.

Also we might want to look at namespacing the auth config under auth.backend instead to better support any future changes bringing us closer to Laravel's auth system.

.github/workflows/code-quality-pr.yaml
```diff
@@ -10,13 +10,10 @@ jobs:
...
         uses: actions/checkout@v1
-      - name: Install PHP
-        uses: shivammathur/setup-php@master
+      - name: Install PHP and PHP Code Sniffer
+        uses: shivammathur/setup-php@v1
         with:
-          php-version: 7.2
-      - name: Install Composer dependencies
-        run: composer install --no-interaction --no-progress --no-suggest
+          php-version: '7.3'
```
LukeTowers, line 16: What's with this change? Do our tests still support 7.2?

src/Support/helpers.php
```diff
@@ -210,13 +227,31 @@ function temp_path($path = '')
     }
 }
 
+if (!function_exists('e')) {
```
LukeTowers, line 230: Did Laravel get rid of this helper and now we have to implement it ourselves?

.github/workflows/code-quality-pr.yaml
bennothommo: Yeah, this is just for the Code Quality tests which run externally. It has no effect on the unit tests.

src/Support/helpers.php
bennothommo: No, they didn't remove it, but they set the `$doubleEncode` parameter to `true` by default in 5.6 which stuffed up a lot of October functionality, so I opted to define the helper ourselves with the original `false` default.

bennothommo: https://github.com/octobercms/library/commit/adda4366544e9e690b1a2b6d0cf3011df8a9ca3a

PR #457, opened 2020-01-20
Add missing options in plugin scaffolding

Samuell1:


src/Scaffold/Console/plugin/plugin.stub
```diff
@@ -42,6 +42,54 @@ class Plugin extends PluginBase
     {
 
     }
+    
+    /**
+     * Registers scheduled tasks that are executed on a regular basis.
+     *
+     * @param string $schedule
```
jacobdekeizer: Hi @Samuell1, the register method gives the Schedule object instead of string :), see octobercms/october#4871. Could you change that?

bennothommo:
@Samuell1 any reason why this was closed?

Samuell1:
@bennothommo I talked with luke that this shouldnt be in default scafollding to dont add unnecessary methods.

PR #460, opened 2020-02-06
Add getRelationTypeDefinition methods

### LINKED ISSUE #4930: Add model relationship retrieval method to HasRelationships
October's HasRelationships trait currently loops through its relationTypes arrays directly when retrieving a relation type. Here's an example:
https://github.com/octobercms/library/blob/d00a3fb7b956bcf5d3eaa1d7e8591a99ae18ef84/src/Database/Concerns/HasRelationships.php#L173-L176
but I'd like to have relationships defined in there based on data in the model. I'd like relation information to be retrieved through a new method rather than directly from the arrays.

## Use Case

I'd like relations dynamically placed onto my model instance based on data loaded into it.

Say you have a two-way polymorphic relation. WordPress is a great example of where something like this exists. You have Posts (of type _post, page, product_ etc) in a many to many polymorphic relationship with taxonomies (_category, tag_ etc).

Normally you'd extend your Post model to add the polymorphic relation like so:
```php
Post::extend(function (\Model $model) {
    $model->hasMany['taxonomy_category'] = [
        'My\Plugin\Models\TaxonomyTerm',
        'name' => 'taxonomable',
        'table' => 'my_plugin_taxonomables'
    ];
});
```

However this would add categories to every post type. I only want it to be added to the 'product' post type.

I propose adding two methods:
```php
public function getRelationTypeDefinitions($type)
{
    if (in_array($type, static::relationTypes)) {
        return $this->{$type};
    }

    return [];
}

public function getRelationTypeDefinition($type, $name)
{
    $definitions = $this->getRelationTypeDefinitions($type);

    if (isset($definitions[$type])) {
        return $definitions[$type];
    }

    return null;
}
```

So for example instead of the following in your code
```php
if (isset($this->{$type}[$name])) {
```
you'd call
```php
$definition = $this->getRelationTypeDefinition($type, $name);
if ($definition) {
```

These functions would then be overrideable in a plugin model like so:
```php
public function getRelationTypeDefinitions($type)
{
    if ($this->post_type == 'product') {
        return ['my_custom_definitions'];
    }

    return parent::getRelationTypeDefinitions($type);
}
```

If this is acceptable I can submit a PR. This change would be entirely backwards compatible with no breaking changes.

Flynsarmy:
Rather than read the relation arrays directly, use a methods to retrieve the list of defined relations. These overrideable methods are useful for creating relations on the fly based on data in your model. Fixes  https://github.com/octobercms/october/issues/4930

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -142,7 +142,7 @@ trait HasRelationships
     /**
      * Returns all defined relations of given type.
      * @param string $type Relation type
-     * @return array
+     * @return array|string|null
      */
     public function getRelationTypeDefinitions($type)
     {
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -154,7 +154,7 @@ trait HasRelationships
     }
 
     /**
-     * Returns all defined relations of given type.
+     * Returns the given relation definition.
      * @param string $type Relation type
      * @param string $name Relation name
      * @return array
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -135,7 +135,36 @@ trait HasRelationships
     public function getRelationDefinition($name)
     {
         if (($type = $this->getRelationType($name)) !== null) {
-            return (array) $this->{$type}[$name] + $this->getRelationDefaults($type);
+            return (array) $this->getRelationTypeDefinition($type, $name) + $this->getRelationDefaults($type);
+        }
+    }
+
+    /**
+     * Returns all defined relations of given type.
+     * @param string $type Relation type
+     * @return array
+     */
+    public function getRelationTypeDefinitions($type)
+    {
+        if (in_array($type, static::$relationTypes)) {
+            return $this->{$type};
+        }
+
+        return [];
+    }
+
+    /**
+     * Returns all defined relations of given type.
+     * @param string $type Relation type
+     * @param string $name Relation name
+     * @return array
+     */
+    public function getRelationTypeDefinition($type, $name)
+    {
+        $definitions = $this->getRelationTypeDefinitions($type);
+
+        if (isset($definitions[$name])) {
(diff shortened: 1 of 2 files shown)


bennothommo:
@Flynsarmy could you provide some unit tests for this functionality, to give us a clearer idea of the potential use cases for this change?

Flynsarmy:
Tests added

bennothommo:
@Flynsarmy thanks for that. However, after looking at it a bit more, I feel it is a bit more verbose than just querying the arrays. @LukeTowers @w20k any thoughts on this?

Flynsarmy:
It's useful for overriding what relations are defined based on data in the model. You can't do that easily by modifying the arrays directly.

Flynsarmy:
Ping

Flynsarmy:
Ping

bennothommo:
@LukeTowers @daftspunk I'm happy to merge this, unless you have any objections?

Flynsarmy:
Ping

src/Database/Concerns/HasRelationships.php
```diff
@@ -135,7 +135,36 @@ public function hasRelation($name)
...
+        }
+
+        return [];
+    }
+
+    /**
+     * Returns the given relation definition.
+     * @param string $type Relation type
+     * @param string $name Relation name
+     * @return array
```
bennothommo, line 160: @Flynsarmy your tests indicate that a string can be returned here, if someone is using the short-hand version of defining a relation. It also would return `null` if the definition type or name specified did not exist.
```suggestion
     * @return array|string|null
```

bennothommo: Could you also add a test case that shows an array being returned, just to tick that off as being tested?

bennothommo (changes requested):

Flynsarmy:
Added the extra test and updated the method annotation. Back to you @bennothommo

bennothommo:
@Flynsarmy could you provide a PR to the docs to demonstrate these new methods?

PR #461, opened 2020-02-25
Fix error using count helper on morph relations

BaNkIck:
On Laravel 5.5, the function getForeignKey no longer exists on relations. Some relations (e.g.: Attach) kept working because this function was overridden in October but, for example, on Morph, it was not so it now breaks.

The function getQualifiedForeignKeyName however, is available across all types of relations and is safer to use as it includes the table name in the returned value (which prevents possible ambiguous references on the final query).


bennothommo:
@BaNkIck Looks good - just needs to be tested.

LukeTowers:
@BaNkIck I can't confirm the original issue. Please submit a detailed issue report to the main october repo and we can look at this again.

PR #462, opened 2020-02-26
replace closure in favor of callback method

mjauvin:
This makes objects using Validation trait cacheable until Laravel cache permits closure to be serialized.

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -60,25 +60,28 @@ trait Validation
         }
 
         static::extend(function ($model) {
-            $model->bindEvent('model.saveInternal', function ($data, $options) use ($model) {
-                /*
-                 * If forcing the save event, the beforeValidate/afterValidate
-                 * events should still fire for consistency. So validate an
-                 * empty set of rules and messages.
-                 */
-                $force = array_get($options, 'force', false);
-                if ($force) {
-                    $valid = $model->validate([], []);
-                }
-                else {
-                    $valid = $model->validate();
-                }
-
-                if (!$valid) {
-                    return false;
-                }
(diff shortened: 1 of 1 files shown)


daftspunk:
This will be fixed in Laravel 6?

mjauvin:
Can @bennothommo confirm closures can be serialized in Laravel 6?

mjauvin:
But in any case, this change is harmless, only better, no?

bennothommo:
@mjauvin TBH I'm not sure - I haven't specifically seen any changes revolving around closure handling in the Cache.

LukeTowers:
@mjauvin @bennothommo I'm pretty sure there's something about them switching to opis serializable closures, can't remember when it was introduced though. We've had this discussion before though and it would be working in L6. @mjauvin can you provide example code for what you're currently trying to do, the error that occurs, and how it works on the L6 branch?

mjauvin:
Yes, I though I had seen this as well, but couldn't find it anywhere in their changelogs

LukeTowers:
@mjauvin looks like it was 5.7. Not sure if it's implemented in the cache layer though, although we could probably do so: https://github.com/laravel/framework/commit/556b948c41b273144e5f23f6b3b6167bf7e0c84e

mjauvin:
@LukeTowers unfortunately, cache is not covered here

LukeTowers:
@mjauvin would you be willing to implement support for it in our caching layer using their queue support as a reference?

mjauvin:
I can take a look, but can't promise anything.

LukeTowers:
Any luck @mjauvin?

mjauvin:
@LukeTowers no, sorry.

PR #463, opened 2020-03-04
Allow regular expressions on wildcard parameters

Flynsarmy:
This PR allows page URLs to include regular expressions on their wildcard parameters. For example the following URL is now valid:
> /blog/:page?*|^(page\/\d+)?$

## Use Case
This PR will allow October users to make a single Page work for the following URLs:
> /blog
> /blog/page/2
> /blog/page/4 etc

without it accidentally catching URLs that should load a different Page
> /blog/2020/10/my-post

But really the use cases for allowing regex for wildcard params would be the same as those for allowing it for non-wildcard params.

This PR is backwards compatible and tests have been provided.

### DIFF
--- a/src/Router/Helper.php
+++ b/src/Router/Helper.php
@@ -124,7 +124,7 @@ class Helper
      */
     public static function segmentIsWildcard($segment)
     {
-        $parts = mb_split("\|", $segment, 2);
+        $parts = mb_split('\|', $segment, 2);
         return mb_strpos($parts[0], ':') === 0 && mb_substr($parts[0], -1) === '*';
     }
 
--- a/src/Router/Helper.php
+++ b/src/Router/Helper.php
@@ -40,11 +40,12 @@ class Helper
     public static function segmentizeUrl($url)
     {
         $url = self::normalizeUrl($url);
-        $segments = explode('/', $url);
+        // Explode by unescaped / characters
+        $segments = mb_split('(?<!\\\)\/', $url);
 
         $result = [];
         foreach ($segments as $segment) {
-            if (strlen($segment)) {
+            if (mb_strlen($segment)) {
                 $result[] = $segment;
             }
         }
@@ -123,7 +124,8 @@ class Helper
      */
     public static function segmentIsWildcard($segment)
     {
-        return mb_strpos($segment, ':') === 0 && mb_substr($segment, -1) === '*';
+        $parts = mb_split("\|", $segment, 2);
+        return mb_strpos($parts[0], ':') === 0 && mb_substr($parts[0], -1) === '*';
     }
 
     /**
--- a/tests/Router/RouterTest.php
+++ b/tests/Router/RouterTest.php
@@ -167,6 +167,10 @@ class RouteTest extends TestCase
         $this->assertEquals('code/with', $params['largecode']);
 
         $rule = $router->reset()->route('testRuleId', '/color/:color/largecode/:largecode*|^[a-z]+\/[a-z]+$');
+        $result = $rule->resolveUrl('color/brown/largecode/code/100', $params);
+        $this->assertfalse($result);
+
+        $rule = $router->reset()->route('testRuleId', '/color/:color/largecode/:largecode*|^[a-z]+\/[a-z]+$');
         $result = $rule->resolveUrl('color/brown/largecode/code/with/slashes', $params);
         $this->assertfalse($result);
     }
--- a/tests/Router/RouterTest.php
+++ b/tests/Router/RouterTest.php
@@ -156,6 +156,19 @@ class RouteTest extends TestCase
         $this->assertArrayHasKey('largecode', $params);
         $this->assertEquals('brown', $params['color']);
         $this->assertEquals('code/with/slashes/edit', $params['largecode']);
+
+        $rule = $router->reset()->route('testRuleId', '/color/:color/largecode/:largecode*|^[a-z]+\/[a-z]+$');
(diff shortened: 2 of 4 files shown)


bennothommo:
@Flynsarmy @LukeTowers I'm not against it in theory, but regex is such a broad spectrum that I feel allowing the full power of it is going to be something that is: a) prone to error and b) a hassle to maintain.

Flynsarmy:
Regex is prone to error, sure, but that's on the people who can/want to use it IMO. In my use case adding it for wildcard params allows me to make a blog with wordpress style URLs - something not currently possible with October.

daftspunk:
This seems like it is asking for trouble from a maintenance perspective, I agree with Ben here.

The docs give a clear example on how you could approach this

```
url = "/blog/:slug?*"
```

And this route should take priority over it

```
url = "/blog/:year/:month/:slug"
```

Allowing the specified use case. If this is not the case, perhaps we should look at the route prioritization (longer routes take priority).

Flynsarmy:
Having the routes
```
/blog/:page?*
/blog/:year/:month/:day/:slug*
```
doesn't work. The top one takes priority and catches all.

It seems inconsistent that we'd support regex for non-wildcard routes but not wildcard routes. Also frustrating because wildcard routes will catch all the URLs we don't want in addition to the ones we do - I only want pagination. Without regex 404ing bad URLs I need to add conditional logic in my controller to do the same thing.

bennothommo:
@Flynsarmy Fair points, although if it were up to me, I'd nuke all regular expressions in routing from orbit and keep parameter validation separate.

src/Router/Helper.php
```diff
@@ -123,7 +124,8 @@ public static function replaceParameters($object, $string)
      */
     public static function segmentIsWildcard($segment)
     {
-        return mb_strpos($segment, ':') === 0 && mb_substr($segment, -1) === '*';
+        $parts = mb_split("\|", $segment, 2);
```
bennothommo: @Flynsarmy 
```suggestion
        $parts = mb_split('\|', $segment, 2);
```

tests/Router/RouterTest.php
```diff
@@ -156,6 +156,19 @@ public function testResolveUrl()
         $this->assertArrayHasKey('largecode', $params);
         $this->assertEquals('brown', $params['color']);
         $this->assertEquals('code/with/slashes/edit', $params['largecode']);
+
+        $rule = $router->reset()->route('testRuleId', '/color/:color/largecode/:largecode*|^[a-z]+\/[a-z]+$');
```
bennothommo: @Flynsarmy Could you add another test like this one that tests a URL that would fail the validation explicitly (ie. it has numbers in it), and make sure that is correctly handled?

bennothommo (changes requested):

tests/Router/RouterTest.php
Flynsarmy: https://github.com/octobercms/library/pull/463/files/1f5014bad4b7e52cb129f00159ec2da825c8419b#diff-da99d28fe49830941fd5098bd79edc52R169

Is this what you're looking for?

bennothommo: @Flynsarmy nah, that one is just testing if there's an additional segment. I was hoping to have a test that covers a URL that fails the actual regular expression.

Flynsarmy: Test added

bennothommo (approved):

bennothommo:
@github-actions run

<details>
<summary>✔ Ready for testing</summary>

```js
(async () => {
  // Get issue/PR number
  const number = context.issue.number;

  // Add Testing Needed label
  await githubClient.issues.addLabels({
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: number,
    labels: ['Testing Needed'],
  });

  // Comment body
  const commentBody = 'This pull request has been marked as **Ready for Testing**. To expedite this pull request being'
    + ' merged, we encourage members of the community to test the changes and confirm they are working. You may view our'
    + ' [Testing a Pull Request](https://github.com/octobercms/october/blob/develop/.github/CONTRIBUTING.md#testing-a-pull-request)'
    + ' section in the Contibution Guide to access the changes in your own copy of October CMS.'
    + '\n\n'
    + 'If you have [Docker](https://www.docker.com/), you can quickly run a copy of October CMS with this pull request\'s'
    + ' changes in place via the following command:'
    + '\n\n'
    + '```\n'
    + 'docker run --rm -p 80:80 -e GIT_MERGE_PR=' + number + ' aspendigital/octobercms:develop\n'
    + '```';

  // Add comment
  await postComment(commentBody);
})();
```

LukeTowers:
@bennothommo nothing happened

bennothommo:
@LukeTowers That would be because the Action isn't installed for the library yet, haha :man_facepalming:

Flynsarmy:
Ping

mjauvin:
Note: the documentation will need to be updated for this, specifically this section:

https://octobercms.com/docs/cms/pages#url-syntax

Flynsarmy:
Up @LukeTowers

daftspunk:
Wrote some additional tests to work these new changes. Seems like they don't quite cover all scenarios. The biggest issue is that wildcard support is already stretching the logic (only 1 wildcard is supported in a route). Adding regex would therefore allow a second wildcard to be introduced and the router did not handle it well

I've kept the tests but commented out in case we ever revisit this idea. Discussed with @Flynsarmy and he's ok with it! Thanks mate 🙏

PR #464, opened 2020-03-06
Add static method typehints to schema facade

jacobdekeizer:
Add typehinting to schema facade like laravel does.

From this:
image

To this:
image

### DIFF
--- a/src/Support/Facades/Block.php
+++ b/src/Support/Facades/Block.php
@@ -9,7 +9,7 @@ use October\Rain\Support\Facade;
  * @method static void endBlock(bool $append = false)
  * @method static void set(string $name, string $content)
  * @method static void append(string $name, string $content)
- * @method static void placeholder(string $name, string $default)
+ * @method static string placeholder(string $name, string $default = null)
  * @method static void get(string $name, string $default = null)
  * @method static void reset()
  *
--- a/src/Support/Facades/Block.php
+++ b/src/Support/Facades/Block.php
@@ -3,18 +3,22 @@
 use October\Rain\Support\Facade;
 
 /**
- * Http Block Facade
+ * @method static void put(string $name)
+ * @method static void startBlock(string $name)
+ * @method static void endPut(bool $append = false)
+ * @method static void endBlock(bool $append = false)
+ * @method static void set(string $name, string $content)
+ * @method static void append(string $name, string $content)
+ * @method static void placeholder(string $name, string $default)
+ * @method static void get(string $name, string $default = null)
+ * @method static void reset()
  *
- * @package october\support
- * @author Alexey Bobkov, Samuel Georges
+ * @see \October\Rain\Html\BlockBuilder
  */
 class Block extends Facade
 {
     /**
      * Get the registered name of the component.
-     *
-     * Resolves to:
-     * - October\Rain\Html\BlockBuilder
      *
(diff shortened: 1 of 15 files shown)


jacobdekeizer:
@bennothommo @LukeTowers If this PR is accepted, I'm willing to add docblocks for all facades :)

LukeTowers:
Is this not something that's handled by using the IDE Helper plugin? I would rather not have these IDE specific comments littered all over the code base purely for the sake of making one particular IDE slightly nicer to use.

jacobdekeizer:
@LukeTowers Why should we use a plugin which comes with other disadvantages (warnings like multiple definitions exist for class x in phpstorm), instead of fixing it in the core so we improve the development experience of octobercms for everyone.

LukeTowers:
@jacobdekeizer because this specific usage is not improving the development experience for everyone, it's only improving it with those people that run IDEs that can parse this information. We use regular docblocks pretty much everywhere in October, but that's because they serve a dual purpose in that they're human and machine readable. 

I think this specific example is getting a bit far into the realm of catering to one specific development platform at the expense of the overall code clarity and maintainability (now these docblocks have to be maintained along with the rest of the actual code). I'd like to hear @daftspunk's input on this first however.

jacobdekeizer:
@LukeTowers I understand what you mean, but it's proven to be liked by many developers in the laravel community laravel/ideas#284. It indeed brings costs extra work to maintain, but in my opinion it brings massive improvements to de development experience for a lot of developers.

LukeTowers:
@jacobdekeizer I understand there is a benefit to some developers, otherwise this wouldn't have been requested in the first place after all 😄 Nevertheless I'd like to get @daftspunk's input on it since this sort of thing is toeing the line with October's ethos that "things should be as simple as possible, no simpler" and this sort of functionality could be provided by a plugin, even if the plugin needs more work to be more helpful to developers.

jacobdekeizer:
@LukeTowers In my opinion installing a plugin for development and removing that on a production environment is not "simpler" than integrating docblocks in the core of octobercms :), but maybe that is a difference in opinions.

daftspunk:
This does appear to be something that has been [introduced in Laravel](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Support/Facades/Schema.php) and will make it easy for us to maintain moving forward. Please make sure it follows the same standard, the current proposal is missing:

```
 * @see \Illuminate\Database\Schema\Builder
```

rajakhoury:
plugins break and lead to more problems.
It does improve the developer experience by having a faster and clearer access to available methods in a certain facade for example.

LukeTowers:
@jacobdekeizer Had a discussion with @daftspunk and it'd be great if you could do the other facades too. Thanks for working with us to get this introduced!

jacobdekeizer:
@LukeTowers I already made them in a different branch, should I merge them in this pr or create a separate pr :)

LukeTowers:
@jacobdekeizer merge them into this PR please

jacobdekeizer:
@LukeTowers I've merged the branch

jacobdekeizer:
@bennothommo I generated them manually. I applied all your changes 👍

bennothommo:
@jacobdekeizer Perfect! Thanks heaps :)

PR #471, opened 2020-03-16
Use "path.config" app variable for loading configuration

bennothommo:
This allows the config path to be changed dynamically, ie. for tests.


mjauvin:
@bennothommo where is path.config defined? Should we set a default value for it?

bennothommo:
@mjauvin It's defined in Laravel here:

https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Application.php#L291

and here:

https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Application.php#L340

Their default works for us, as we use the same.

PR #472, opened 2020-03-21
Add throws docblocks to blockbuilder

jacobdekeizer:
Add throws docblocks to blockbuilder

### DIFF
--- a/src/Html/BlockBuilder.php
+++ b/src/Html/BlockBuilder.php
@@ -26,6 +26,7 @@ class BlockBuilder
     /**
      * Begins the layout block.
      * @param string $name Specifies the block name.
+     * @return void
      */
     public function startBlock($name)
     {
@@ -36,6 +37,7 @@ class BlockBuilder
     /**
      * Helper for endBlock and also clears the output buffer.
      * @param boolean $append Indicates that the new content should be appended to the existing block content.
+     * @return void
      * @throws Exception if there are no items in the block stack
      */
     public function endPut($append = false)
@@ -50,6 +52,7 @@ class BlockBuilder
     /**
      * Closes the layout block.
      * @param boolean $append Indicates that the new content should be appended to the existing block content.
+     * @return void
      * @throws Exception if there are no items in the block stack
      */
     public function endBlock($append = false)
@@ -73,6 +76,7 @@ class BlockBuilder
      * Sets a content of the layout block.
      * @param string $name Specifies the block name.
      * @param string $content Specifies the block content.
+     * @return void
      * @throws Exception if there are no items in the block stack
      */
     public function set($name, $content)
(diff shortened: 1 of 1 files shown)


LukeTowers:
Minor quibble, but shouldn't the throws be above the return?

jacobdekeizer:
@LukeTowers my ide always adds them as last and so do I.

bennothommo:
@jacobdekeizer I'm wondering they should follow the typical "path" of a method, like the below:

```php
/**
 * Summary.
 * 
 * Long description.
 * 
 * // 1. Informational tags
 * @see Example
 * @deprecated 1.0.0
 * 
 * // 2. Parameters - the input into the functionality, the first step in any method.
 * @param string $example
 * @param bool|null $example2
 * 
 * // 3. Throws - exceptions happen DURING the functionality, so they should be listed afterwards.
 * @throws \Exception if something goes wrong
 * @throws \AnotherException if something else goes wrong
 * 
 * // 4. Return - the output of the functionality, the last step in any method.
 * @return bool
 */
```

jacobdekeizer:
@bennothommo Most of the times I see them defined as last. I think it's personal preference. But if you want them before the return statement I could change that :). Let me know.

nathan-van-der-werf:
Laravel and Symfony have @throws at the bottom, see:

https://laravel.com/docs/5.7/contributions#phpdoc
https://symfony.com/doc/3.4/contributing/code/standards.html#symfony-coding-standards-in-detail

bennothommo:
@jacobdekeizer @nathan-van-der-werf fair enough - leave them as is (at the bottom).

LukeTowers:
I personally prefer the approach of having them follow the flow / path of the method's execution. What's your personal preference @bennothommo? Maybe @daftspunk should weigh in too. I would rather we followed the preference of the core maintainers for this specific item :) I know it's a minor quibble but we should use this as an opportunity to set the project's standard for this.

bennothommo:
@LukeTowers I'm only suggesting my way because that's what me and my colleagues use at the office. In no way am I saying my way is right. Given that all the other frameworks do @throws at the bottom, I'd be quite happy for us to follow their lead too.

As a minor point though, I'm not a big fan of us being dictated by PHPStorm's way of doing things.

LukeTowers:
@daftspunk said he prefers at the end so that combined with other frameworks using that style means that's what we'll do from now on.

src/Html/BlockBuilder.php
```diff
@@ -37,6 +38,7 @@ public function startBlock($name)
      * Helper for endBlock and also clears the output buffer.
      * @param boolean $append Indicates that the new content should be appended to the existing block content.
      * @return void
+     * @throws Exception if there are no items in the block stack
```
bennothommo: @jacobdekeizer could you make sure this and the following throws declarations are using the root namespace `\Exception`?
```suggestion
     * @throws \Exception if there are no items in the block stack
```

bennothommo:
@jacobdekeizer Once you have resolved the other `Exception` @throws and added the forward slash to denote a root `Exception` class, I will merge this :)

PR #476, opened 2020-03-21
Add json extension to composer.json

jacobdekeizer:
afbeelding


bennothommo:
@jacobdekeiser I'm not sure this is required - despite what PHPStorm might be saying. JSON is installed by default in every version of PHP we support.

jacobdekeizer:
@bennothommo I always add it to my composer.json and Laravel does it too. See https://github.com/laravel/framework/blob/de01e2926a7560c43fb763d86de134a52b46928b/composer.json#L19.

bennothommo:
@jacobdekeizer if it's a default extension, do we need to cover it though? I mean, would we need to add `ext-pdo`, `ext-gd` etc?

jacobdekeizer:
If you wan't a "complete" composer.json then yes 😄. For example drupal has all the extensions listed in their composer.json. https://github.com/drupal/drupal/blob/9.1.x/core/composer.json

bennothommo:
@jacobdekeizer TBH we might as well have the whole lot in there, if we technically need it. But I'm not sure if that's overkill.

jacobdekeizer:
@bennothommo Laravel also doesn't include all extension. I think they only added the extensions which phpstorm complains about 😄. If someone encouters a problem with another missing extension, we could add that extension later in my opinion.

bennothommo:
@jacobdekeizer we have an opportunity to be ahead of the curve here though :P

jacobdekeizer:
@bennothommo Do you have a list of all the extensions octobercms uses?

LukeTowers:
Is there a dotfile we can include to just tell PHPStorm to shut up about specific non-issues? This is getting ridiculous. October has made several design decisions around how the project is structured to keep things as simple and approachable for beginners as possible (for example having all the modules in one repo and then subsplit behind the scenes into the various packages, having the rainlab plugins separate, all the other repositories under the octoberrain repository instead of the main one, etc). 

Adding these extensions to the composer.json file clutters it unnecessarily, especially since the extensions we require are included in the versions of PHP that we support by default. Again, I'll defer to @daftspunk on this, but that's my initial thoughts on the matter.

nathan-van-der-werf:
From the composer docs (https://getcomposer.org/doc/04-schema.md); also, not all setups are equal, it's a good thing to be strict about the requirements of a project.

> Note: It is important to list PHP extensions your project requires. Not all PHP installations are created equal: some may miss extensions you may consider as standard (such as ext-mysqli which is not installed by default in Fedora/CentOS minimal installation systems). Failure to list required PHP extensions may lead to a bad user experience: Composer will install your package without any errors but it will then fail at run-time. The composer show --platform command lists all PHP extensions available on your system. You may use it to help you compile the list of extensions you use and require. Alternatively you may use third party tools to analyze your project for the list of extensions used.

jacobdekeizer:
@LukeTowers An incomplete composer.json has nothing to do with design decisions of october or how simple and approachable the project is. Adding this can even prevent issues non related to october, but to the php installation.

LukeTowers:
@jacobdekeizer system requirements for October are listed here: https://octobercms.com/docs/setup/installation#system-requirements. I'm going to change the base branch to the WIP L6 one since we'll be putting that in develop shortly. Please put the extension requirements immediately below the PHP one.

martendej:
@LukeTowers In my opinion this is very simple: If October requires an extension, it should be added to the composer.json file. That is what requirements are for, and it only saves beginning users a lot of pain instead of the 'clutter' you described. Having vague runtime errors because the users' PHP build does not include some required extensions is way more frustrating than a clear error message by composer in advance.

jacobdekeizer:
@LukeTowers I've added the other php extensions

PR #477, opened 2020-03-23
Cleanup auth manager

jacobdekeizer:
As discussed with @daftspunk, I would make a pr for changes to the auth manager.

- Fix issues with auth manager so it follows the illuminate contracts.
- Fix autocompletion issues

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -64,11 +64,6 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
     public $ipAddress = '0.0.0.0';
 
     /**
-     * @var \Illuminate\Http\Request $request
-     */
-    private $request;
-
-    /**
      * Initializes the singleton
      */
     protected function init()
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -1,6 +1,7 @@
 <?php namespace October\Rain\Auth;
 
 use Cookie;
+use October\Rain\Database\Builder;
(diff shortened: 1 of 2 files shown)


bennothommo:
@jacobdekeizer Would the try/catch blocks you have used to catch `AuthException` instances result in different functionality, as those exceptions were passed through before?

jacobdekeizer:
@bennothommo yes it would. But the auth manager didn't follow the laravel contracts. If someone uses a package that is using these contracts it would not work, because this implementation would throw exceptions, where the contract says it only returns true or false. That could lead to unfixable errors.

bennothommo:
@jacobdekeizer I understand, however, given that the Auth in October has operated so long in a specific fashion, it is incredibly likely that people have based their own custom auth implementations around the way our auth functions right now. I know that several people use our Auth for API-driven developments, and I cannot discount that they may be relying on `AuthException` instances being thrown.

I'll have a chat with the team and work out if this is an acceptable BC break or not.

jacobdekeizer:
I think this is the right time to make such a change, we all have to check our code with the laravel 6 upgrade.

LukeTowers:
The L6 upgrade is not an excuse to put in breaking changes. As is right now the October auth system requires some level of custom work per project to make work with most Laravel auth packages (although a lot less than it used to) so I see no benefit of introducing the breaking change of no longer passing along exceptions for the sake of "some packages that might not expect an exception to be thrown".

PR #478, opened 2020-03-26
Add throws docblocks to file loader

jacobdekeizer:
Add throws docblocks and fix code style issues

### DIFF
--- a/src/Config/FileLoader.php
+++ b/src/Config/FileLoader.php
@@ -36,8 +36,8 @@ class FileLoader implements LoaderInterface
     /**
      * Create a new file configuration loader.
      *
-     * @param  \Illuminate\Filesystem\Filesystem  $files
-     * @param  string  $defaultPath
+     * @param \Illuminate\Filesystem\Filesystem $files
+     * @param string $defaultPath
      * @return void
      */
     public function __construct(Filesystem $files, $defaultPath)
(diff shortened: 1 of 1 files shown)


src/Config/FileLoader.php
```diff
@@ -183,35 +186,33 @@ protected function getPackagePath($package, $group, $env = null)
...
+        return $this->defaultPath . '/' . $file;
     }
 
     /**
      * Get the configuration path for a namespace.
      *
-     * @param  string  $namespace
-     * @return string
+     * @param string $namespace
+     * @return string|void
```
LukeTowers: When would this return void?

jacobdekeizer: If the elseif is not true

LukeTowers: What conditions would trigger that? I'm wondering if we should have better handling for it than just returning void

jacobdekeizer: If the namespace doesn't exist in $this->hints. How would you like me to solve it? add a return null to the bottom?

LukeTowers: > If the namespace doesn't exist in $this->hints

I know that 😛 

I'm curious to know what application state would cause that (as well as what happens if it does occur), and if it's at risk of happening in the course of normal operations (or even non-normal operations) should we be considering throwing an exception instead?

jacobdekeizer: @LukeTowers `return null` is the best option, see line 123. After `$path = $this->getPath($namespace)` an `is_null` check is done. `is_null` returns true voor type void. So it won't break any functionality :)

LukeTowers: Sounds good, lets make the switch then

jacobdekeizer: @LukeTowers done :)

PR #479, opened 2020-03-26
Cleanup database builder

jacobdekeizer:
- Remove unused import
- Add builder typehint to callbacks
- Fix code style

### DIFF
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -136,7 +136,7 @@ class Builder extends BuilderModel
         }
 
         $total = $this->toBase()->getCountForPagination();
-        $this->forPage((int)$currentPage, (int)$perPage);
+        $this->forPage((int) $currentPage, (int) $perPage);
 
         return $this->paginator($this->get($columns), $total, $perPage, $currentPage, [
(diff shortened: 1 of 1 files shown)


src/Database/Builder.php
```diff
@@ -69,30 +68,29 @@ protected function searchWhereInternal($term, $columns, $mode, $boolean)
         }
 
         if ($mode === 'exact') {
-            $this->where(function ($query) use ($columns, $term) {
+            $this->where(function (Builder $query) use ($columns, $term) {
```
LukeTowers: Is this referring to itself? Why is it required?

jacobdekeizer: Yes it's referring to itself, because a callback is default mixed typed.

LukeTowers: Right, but what benefit is this specific change adding?

jacobdekeizer: Autocompletion for ide's and let the developer know which type is returned

jacobdekeizer: @LukeTowers any update on this?

LukeTowers: LGTM, thanks!

PR #480, opened 2020-03-26
Add illuminate schema builder typehint

jacobdekeizer:


src/Database/DatabaseServiceProvider.php
```diff
@@ -84,7 +84,7 @@ protected function getDefaultDatabaseDriver()
      */
     protected function swapSchemaBuilderBlueprint()
     {
-        $this->app['events']->listen('db.schema.getBuilder', function ($builder) {
+        $this->app['events']->listen('db.schema.getBuilder', function (\Illuminate\Database\Schema\Builder $builder) {
```
LukeTowers: What's the benefit of this?

jacobdekeizer: Typehint for an ide, so we don't get errors in this code

PR #481, opened 2020-03-26
Cleanup data feed

jacobdekeizer:
- Remove extract so we don't have magic variables.
afbeelding
- Add docblocks

### DIFF
--- a/src/Database/DataFeed.php
+++ b/src/Database/DataFeed.php
@@ -247,7 +247,7 @@ class DataFeed
                 ->addSelect(Db::raw($sortSelect));
 
             // Union this query with the previous one
-            if ($lastQuery !== null) {
+            if ($lastQuery) {
                 if ($this->removeDuplicates) {
                     $cleanQuery = $lastQuery->union($cleanQuery);
                 } else {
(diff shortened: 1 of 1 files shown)


src/Database/DataFeed.php
```diff
@@ -228,9 +245,9 @@ protected function processCollection()
             $cleanQuery = $cleanQuery->select(Db::raw($idSelect))
                 ->addSelect(Db::raw($tagSelect))
                 ->addSelect(Db::raw($sortSelect));
-            
+
             // Union this query with the previous one
-            if ($lastQuery) {
+            if ($lastQuery !== null) {
```
LukeTowers: What was the reason for this change?

jacobdekeizer: More strict check, nothing else

LukeTowers: If you want a more strict check then check to see if it's actually the object that it's expected to be, one that supports the `union()` and `unionAll()` methods.  Checking `if ($lastQuery)` vs `if ($lastQuery !== null)` provides no benefit and actually is worse because it allows falsey values like an empty array, string, or false boolean to slip through the cracks.

jacobdekeizer: I could remove it :)

LukeTowers: Revert it or make it properly strict, whichever you prefer. !== null won't be merged though 😉

jacobdekeizer: @LukeTowers reverted null check

PR #483, opened 2020-03-31
Lang support

SaifurRahmanMohsin:
Switched from using hardcoded string to language strings.

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -1,5 +1,6 @@
 <?php namespace October\Rain\Auth;
 
+use Lang;
 use Cookie;
 use Session;
 use Request;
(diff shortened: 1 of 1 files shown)


LukeTowers:
This has been proposed before (see octobercms/library#318) and will not be accepted for the following reasons:

1. This translation should happen in the backend module layer, not the core authentication library
2. Specific reasons for authentication failure are **only** provided while the application is in debug mode and **only** for debugging purposes, they should **never** be used in production. Giving specific reasons for authentication failure in production only makes attackers lives easier.

LukeTowers:
@SaifurRahmanMohsin see https://github.com/octobercms/october/issues/2106 for the full discussion around this. Specifically what I mean by 1 is that if we were to translate the various reasons for auth failure that'd need to be done with different exception classes and detecting those classes in the backend. In this case it was an imperfect solution but easier to implement at the time.

PR #484, opened 2020-04-01
Pass current SCSS asset path down to children to use as new root path.

bennothommo:
Previously, this method would pass the same loadPath to all assets down the hierarchy, which would result in any relative paths not being able to be found. This change will take the current directory of the current asset and use that as a base for all child assets.

Will need to test this with a few SCSS-based themes.

### DIFF
--- a/src/Assetic/Filter/ScssphpFilter.php
+++ b/src/Assetic/Filter/ScssphpFilter.php
@@ -134,9 +134,10 @@ class ScssphpFilter implements DependencyExtractorInterface
         foreach (CssUtils::extractImports($content) as $match) {
             $file = $sc->findImport($match);
             if ($file) {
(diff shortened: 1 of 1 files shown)


mjauvin:
I confirm that fixes the bootstrap 4 scss with child imports.

LukeTowers:
@bennothommo or @JaxWilko could you port this over to https://github.com/assetic-php/assetic/blob/2.0/dev/src/Assetic/Filter/ScssphpFilter.php?

bennothommo:
@LukeTowers - @JaxWilko might need to implement it, I don't have write access to the Assetic repo.

LukeTowers:
@bennothommo thanks for letting me know, I've updated your permissions.

PR #485, opened 2020-04-01
add Less filter

mjauvin:
This allows using less assets in theme asset compiler


mjauvin:
That remaining Code Quality check error I just can't see what's wrong here... I went through every single line of that file and there is no extra spaces as it says. Help!

src/Assetic/Filter/LessFilter.php
```diff
@@ -0,0 +1,208 @@
+<?php namespace October\Rain\Assetic\Filter;
```
LukeTowers, line 1: We don't want this filter, just the LessphpFilter

mjauvin: Gone! 
Any idea about the Code Quality error n LessphpFilter.php ?

LukeTowers: No idea.

src/Assetic/Filter/LessphpFilter.php
```diff
@@ -0,0 +1,165 @@
...
+
+        if ($this->formatter) {
+            $lc->setFormatter($this->formatter);
+        }
+
+        if (null !== $this->preserveComments) {
+            $lc->setPreserveComments($this->preserveComments);
+        }
+
+        if (method_exists($lc, 'setOptions') && count($this->options) > 0 ) {
```
bennothommo, line 110: @mjauvin The code quality check is saying the issue lies on line 110:
https://github.com/octobercms/library/pull/485/checks#step:5:10
It's saying it expected 0 spaces *before* the closing bracket, but there is one there.

```suggestion
        if (method_exists($lc, 'setOptions') && count($this->options) > 0) {
```

mjauvin:
Ahhh, thanks Ben, I just didn't see it!!

LukeTowers:
Thanks @bennothommo, my brain's kinda fried today 😂

mjauvin:
What got me is the error mentioned a closing "bracket" but it was a parenthesis...

bennothommo:
@mjauvin all good - I can see how the error message would be misleading. I have PHPCS set up in VSCode which highlights things like this in the code, so I normally don't notice the wonky error wording - I just tweak it until my editor shuts up.

PR #486, opened 2020-04-11
rename parent property

mjauvin:
prevent conflict with SimpleTree/NestedTree traits


mjauvin:
This has been tested and it resolves the issue.

PR #487, opened 2020-04-13
Add sidebar layout to scaffold command

alxy:
This PR adds a `layout` option to the controller scaffold command. The syntax is as follows:

```
php artisan create:controller Acme.Blog Posts --layout=sidebar
```
This will automatically use the secondary fields and display them in a sidebar as described [here](https://octobercms.com/docs/backend/views-partials#layout-form-with-sidebar).

Additionally, I replaced the statix english phrases with the [corresponding translate strings](https://github.com/octobercms/october/blob/develop/modules/backend/lang/en/lang.php#L247-L307).

PR #490, opened 2020-05-06
Translate buttons in stub template

dead23angel:
Put translations in stub templates for the command create:controller


Samuell1:
I think @alxy already implemented this with this PR https://github.com/octobercms/library/pull/487

dead23angel:
Yes, it turns out already implemented, close

PR #491, opened 2020-05-08
Allow deep symlinks when translating paths to local URLs

bennothommo:
This extends the work done by @LukeTowers in https://github.com/octobercms/library/commit/a88b5fd3880edeac0c2bb3bf9e5297c508300055 by doing a search (once) for all folders within the base directory and determining symlinks within.

It will create an array of sources and targets in order to translate these to URLs.

This allows symlinks to be placed multiple levels deep, for example:

- `/home/test/projects/october/plugins/rainlab/pages` could be symlinked to `/home/test/projects/rainlab-pages`

In addition, this will respect the `cms.restrictBaseDir` config value, in that symlinks that link to folders outside the base directory will not be recorded if restricted to the base directory, but symlinks that remain within the base directory will still work.

The only issue with this change is a slight performance hit when reading the directory structure. While this is mitigated by only doing the search once when needed (an asset that does not link to a real path in the base directory), it could be further mitigated by caching the symlinks, although this is offset by having to clear the cache in some way when symlinks change.


LukeTowers:
@bennothommo are you wanting this in 466 or can we push it to 467?

bennothommo:
@LukeTowers Let's bump it to 467. It's all working and tested - I'm just not a fan of the ~30-50ms increase in load time, so I would like to implement some form of caching.

bennothommo:
Symlinks are now cached for 1 hour, so this should alleviate the slight performance hit.

bennothommo:
@LukeTowers at this stage, just when it tries to generate a URL for a local path (the `Filesystem::localToPublic()` method). I've been using it in conjunction with the `develop.decompileAssets` config to use symlinks for Rainlab plugins.

LukeTowers:
@bennothommo I'm not a huge fan of the performance hit, even if it's cached. Perhaps we could do uncached and then check for `develop.allowNestedSymlinks` instead?

bennothommo:
@LukeTowers I'm happy to do it that way. With the original functionality that this was replacing (allowing a first-level subdirectory to be symlinked), do you want that to be edited to respect `cms.restrictBaseDir`? It currently allows the symlink to go outside of the base directory which can be a security risk.

LukeTowers:
@bennothommo it should yes. If you're using symlinks outside the base dir you pretty much have to have `cms.restrictBaseDir` disabled anyways for anything else to work. It's not really that big of an issue right now because it only applies to generating a public URL to a file that's publicly accessible anyways, it's merely an internal tool for URL generation.

bennothommo:
Proposed setting for `config/develop.php`

```
    /*
    |--------------------------------------------------------------------------
    | Allow deep-level symlinks
    |--------------------------------------------------------------------------
    |
    | October CMS, by default, will allow symlinks within the first level of
    | subdirectories. When this feature is enabled, the system will allow
    | symlinks to be used at any directory level. This can be useful for
    | symlinking individual plugins or themes.
    |
    | Please note that this has a negative effect on performance. It also
    | still respects if "cms.restrictBaseDir" is enabled - if so, symlinks
    | cannot point to resources outside of the root folder.
    |
    | true  - allow deep symlinks
    |
    | false - do not allow deep symlinks (default)
    |
    */

    'allowDeepSymlinks' => false,
```

bennothommo:
@LukeTowers I've implemented your requested changes, and re-targeted this for L6. I'll add the proposed config variable above if you're happy for this PR to be merged.

src/Filesystem/Filesystem.php
```diff
@@ -93,14 +100,31 @@ public function localToPublic($path)
...
+                foreach ($this->symlinks as $source => $target) {
+                    if (strpos($path, $target) === 0) {
+                        $relativePath = substr($path, strlen($target));
+                        $result = str_replace("\\", "/", substr($source, strlen($publicPath)) . $relativePath);
+                        break;
+                    }
+                }
+            }
+        } elseif ($directories = $this->glob($publicPath . '/*', GLOB_NOSORT | GLOB_ONLYDIR)) {
             foreach ($directories as $dir) {
```
LukeTowers, line 122: Add a comment to this block explain its purpose like the one above it please

bennothommo:
@LukeTowers I've simplified the symlink check by using my new method to get just a single level of symlinks by default (if `develop.allowDeepSymlinks` is `false`), as this way it caches symlinks for that request in the `$symlinks` property and also respects `restrictBaseDir`.

src/Filesystem/Filesystem.php
```diff
@@ -445,8 +438,10 @@ protected function findSymlinks()
                     continue;
                 }
 
-                // Get subfolders
-                $iterator($directory->getPathname());
+                // Get subfolders if "develop.allowDeepSymlinks" is enabled.
+                if ($deep) {
+                    $iterator($directory->getPathname());
```
LukeTowers, line 443: Don't you need to pass the value of $deep to this iterator call too?

bennothommo: @LukeTowers nope - it uses the `$deep` variable from the outer scope. :)

LukeTowers: Ah, completely missed the use statement there :) Looks good!

PR #492, opened 2020-05-14
Add config value to control loading of discovered Laravel packages

bennothommo:
Adds the config value `app.loadDiscoveredPackages` (default: `false`) - if true, packages discovered through Laravel's autodiscovery feature will be loaded.

The reason for this control is that in some cases, packages which are autodiscovered belong to plugins. When the plugins themselves are disabled, the discovered package is still being loaded. See https://github.com/rainlab/debugbar-plugin/issues/34 for an example.

Proposed `config/app.php` change:

```php
    /*
    |--------------------------------------------------------------------------
    | Load automatically discovered packages
    |--------------------------------------------------------------------------
    |
    | By default, October CMS uses automatic discovery of Laravel packages.
    | In some cases, these packages will still operate even if the plugin
    | using them is disabled.
    |
    | Set this to `false` to disallow automatic loading of these packages. The
    | plugins will then be responsible for loading the packages manually.
    |
    | Please note that packages defined in `app.providers` will still be loaded
    | even if discovery is disabled.
    |
    */

    'loadDiscoveredPackages' => true,
```


LukeTowers:
@bennothommo what are your thoughts on having the default be true vs false?

bennothommo:
@LukeTowers just what was discussed in the Debugbar issue thread. Having it default to `false` means that it actually resolves the issue I made, namely, it won't load the discovered packages and puts the onus on the plugins to manually register them. @daftspunk believes we don't use package discovery at all (although turns out we do, with Carbon and Jenssegers\Date), so it should be safe to default to `false`, but I can't discount that some people are (knowingly or unknowingly) relying on the behaviour for their plugins.

Having it default to `true` will keep the same behaviour October has now intact, and basically nothing changes, except that people have the option to disable it if they wish (although, they already had that capability with [adding `dontDiscover`](https://laravel.com/docs/5.5/packages#package-discovery) to their `composer.json`).

LukeTowers:
Personally I would lean more towards disabling it by default since I would rather we encourage developers to manually register their dependencies, although it's hard to say for sure because I don't know how many people have been using it. Are we able to adjust the logic so that it can dynamically load from composer.json files in plugins (which would be required for plugin's dont-discover to function properly)? (although, this wouldn't work for any marketplace plugins since their composer.json file gets removed by the build process)

bennothommo:
@LukeTowers I'd probably not rely on the composer.json inside plugins exactly for the reason you specified. That being said, if they do not have composer.json files, then their libraries won't be included in the `installed.json` file that Laravel package discovery uses - so perhaps it might be safe to assume that plugins are loading their packages manually?

LukeTowers:
@bennothommo hmm, that would be true for marketplace plugins, I believe the plugin booting process takes care of adding dependencies from the plugin's vendor directory. So we could then in that case use composer.json files of plugins if present since that would be the case where they would be included in the project installed.json anyways.

bennothommo:
@LukeTowers I don't believe our plugin architecture does package discovery though in the bootstrap - it's purely Laravel doing that through the `installed.json` file. So I don't think there's any need to rely on the plugin's `composer.json`. My change only needs to handle people including libraries in the root `composer.json` file.

bennothommo:
Oh I see what you mean - you want us to check the plugin's Composer file for `dont-discover` definitions, is that right?

LukeTowers:
@bennothommo that's correct

LukeTowers:
Going to push this to 467

LukeTowers:
@bennothommo I've discovered an issue with how Laravel handles this right now that we should fix. If a developer adds a new package that provides autodiscovered providers / aliases via composer require, and then runs the package:discover command, that package's providers / aliases will be added to the generated files under /storage/framework. 

However, if that package is then removed from the project, the application will break as it will attempt to load the cached registered provider / aliases even though they no longer exist. See https://github.com/laravel/ideas/issues/796 & https://github.com/laravel/framework/issues/21626. 

The officially recommended solution is to avoid using any facades in ServiceProviders as using a  facade will cause that cache layer to be hit, thus attempting to load a now invalid cache file before package:discover can be run, preventing the cache from being invalidated through any method other than manual removal of the cached framework files. See https://github.com/mpyw/laravel-cached-database-stickiness/commit/ba9749f6a280345880a0531fef123fc45ebe3501?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8 for an implementation of this. 

That suggestion is unacceptable to me, we should not have to ban the use of facades in ServiceProviders (and thus all plugin classes) and we couldn't even if we wanted to. Instead, what we should do is automatically invalidate the cache and attempt to run package:discover again if the loading process fails for any providers / aliases registered by the cache files. Once we've done that, Laravel can choose to take the change or not, but either way since we're making the changes in this PR already we might as well work in the fix for this issue that's been annoying me for a while. Also I'm sure @LarryBarker would much appreciate it 😂

bennothommo:
@LukeTowers I'll see what I can do - I have a feeling this might be a tall order given that providers can be deferred, so there's no easy way to tell (and perhaps try/catch) when it might fail.

LukeTowers:
@bennothommo even if it's deferred, doesn't the framework have a centralized place for loading the packages when they're requested? If you could hook into that and then just check if the requested package existed in those cache files (if there isn't a better way of doing it) then I feel like you could wrap and refresh it there.

bennothommo:
@LukeTowers packages, yes. Facades are a different story, from memory.

LukeTowers:
@bennothommo can you change the target to Laravel 6 branch and the default value of the config to false? And then add a note to the L6 upgrade guide about this change, why we always recommend manual registration of providers and facades in plugins, and how to get the old behavior (namely, set app.loadDiscoveredPackages to true)

bennothommo:
@LukeTowers will do.

PR #493, opened 2020-05-17
improve path helper methods

mjauvin:
- fix uploads_path() helper method
- add media_path() helper method


LukeTowers:
@mjauvin please post a description of how the existing uploads_path() helper is broken. Additionally, shouldn't both those paths support being configurable just like the themes & plugin paths are?

mjauvin:
The uploads_path() helper was referencing  app('path.uploads') which didn't exist...

As for configurability, since their value is taken from config, I though this would be redundant to also make it changeable in the Application class... what do you think @LukeTowers?

bennothommo:
@mjauvin Unfortunately, this change doesn't seem to take into account the `storage.uploads.path` and `storage.media.path` configuration settings.

mjauvin:
@bennothommo you mean from the Application class? I may have read that wrong, but it looks like the Application class holds the default values for those paths, which can be overriden in the cms config files... am I getting this wrong?

bennothommo:
@mjauvin Ah sorry, you're right. Just as long as the helpers are referencing and allowing the correct configuration, which it seems to be, that's all good.

mjauvin:
need to update the docs as well for `media_path()`

mjauvin:
https://github.com/octobercms/docs/pull/456

LukeTowers:
With `$this->basePath` being added, does this still work for remote storage drivers? Could you add some unit tests for these helpers?

mjauvin:
Not sure what you mean @LukeTowers, I only added methods which didn't exist in Application class and fixed the one that was broken in the helpers.

mjauvin:
But yeah, I can look at adding unit tests for the Application class.

mjauvin:
gotta get back to this, please keep opened

mjauvin:
I need help with this. The tests are working in my local install and with github it says it cannot find Config.
I tried including the facade but no luck.

LukeTowers:
@mjauvin try accessing the config through `$this->app->config()`

mjauvin:
It's really odd, why are the tests running fine locally and not on the github environment?

bennothommo:
@mjauvin You cannot use facades in the unit tests because the framework is not technically loaded during them - it instead populates the `$this->app` property with a mock of the application.

You instead need to instantiate the class that the Facade provides and use it directly.

mjauvin:
@LukeTowers I'm about to throw something through the window...

seanthepottingshed:
@mjauvin 

Feel free to throw me out of the window, it is Friday after all!

src/Support/helpers.php
```diff
@@ -198,7 +198,22 @@ function plugins_path($path = '')
...
+if (!function_exists('media_path')) {
+    /**
+     * Get the path to the media folder.
+     *
+     * @param  string  $path
+     * @return string
+     */
+    function media_path($path = '')
+    {
+        $media_path = rtrim(Config::get('cms.storage.media.path', app('path.media')), '/');
```
mjauvin: @LukeTowers Should we be fetching `Config::get('cms.storage.media.path')` here or would it make more sense in Foundation\Application::mediaPath() ...

LukeTowers: Probably whatever the pattern is for the other helpers

mjauvin:
Whoohooooo!!!

mjauvin:
@LukeTowers tests done!

src/Support/helpers.php
```diff
@@ -172,7 +172,7 @@ function traceSql()
      */
     function config_path($path = '')
     {
-        return app('path.config').($path ? '/'.$path : $path);
+        return app('path.config').($path ? DIRECTORY_SEPARATOR.$path : $path);
```
bennothommo: @mjauvin I wonder if we have to, perhaps, make sure the correct directory separator is also used on the `$path` value provided for this method and the other path helpers as well, perhaps using `str_replace`. While people using Windows might use the correct separator for their own code, there's the potential for plugins to be using this method and not doing the same handling.

bennothommo: Also, needs spaces around concatenations:
```suggestion
        return app('path.config') . ($path ? DIRECTORY_SEPARATOR . $path : $path);
```

src/Support/helpers.php
```diff
@@ -185,7 +185,7 @@ function config_path($path = '')
      */
     function plugins_path($path = '')
     {
-        return app('path.plugins').($path ? '/'.$path : $path);
+        return app('path.plugins').($path ? DIRECTORY_SEPARATOR.$path : $path);
```
bennothommo:
```suggestion
        return app('path.plugins') . ($path ? DIRECTORY_SEPARATOR . $path : $path);
```

src/Support/helpers.php
```diff
@@ -198,7 +198,22 @@ function plugins_path($path = '')
      */
     function uploads_path($path = '')
     {
-        return app('path.uploads').($path ? '/'.$path : $path);
+        $uploads_path = rtrim(Config::get('cms.storage.uploads.path', app('path.uploads')), DIRECTORY_SEPARATOR);
+        return $uploads_path.($path ? DIRECTORY_SEPARATOR.ltrim($path, DIRECTORY_SEPARATOR) : $path);
```
bennothommo:
```suggestion
        return $uploads_path . ($path ? DIRECTORY_SEPARATOR . ltrim($path, DIRECTORY_SEPARATOR) : $path);
```

bennothommo: @mjauvin Do you need to do this path trimming on the other helpers too?

src/Support/helpers.php
```diff
@@ -198,7 +198,22 @@ function plugins_path($path = '')
...
+    /**
+     * Get the path to the media folder.
+     *
+     * @param  string  $path
+     * @return string
+     */
+    function media_path($path = '')
+    {
+        $media_path = rtrim(Config::get('cms.storage.media.path', app('path.media')), DIRECTORY_SEPARATOR);
+        return $media_path.($path ? DIRECTORY_SEPARATOR.ltrim($path, DIRECTORY_SEPARATOR) : $path);
```
bennothommo:
```suggestion
        return $media_path . ($path ? DIRECTORY_SEPARATOR . ltrim($path, DIRECTORY_SEPARATOR) : $path);
```

src/Support/helpers.php
```diff
@@ -211,7 +226,7 @@ function uploads_path($path = '')
      */
     function themes_path($path = '')
     {
-        return app('path.themes').($path ? '/'.$path : $path);
+        return app('path.themes').($path ? DIRECTORY_SEPARATOR.$path : $path);
```
bennothommo:
```suggestion
        return app('path.themes') . ($path ? DIRECTORY_SEPARATOR . $path : $path);
```

src/Support/helpers.php
```diff
@@ -224,7 +239,7 @@ function themes_path($path = '')
      */
     function temp_path($path = '')
     {
-        return app('path.temp').($path ? '/'.$path : $path);
+        return app('path.temp').($path ? DIRECTORY_SEPARATOR.$path : $path);
```
bennothommo:
```suggestion
        return app('path.temp') . ($path ? DIRECTORY_SEPARATOR . $path : $path);
```

tests/Foundation/ApplicationTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use October\Rain\Foundation\Application;
+
+class ApplicationTest extends TestCase
+{
+    protected function setUp():void
```
bennothommo:
```suggestion
    protected function setUp(): void
```

tests/Support/HelpersTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->assertEquals(media_path(), Config::get('cms.storage.media.path', app('path.media')));
+    }
+
+    public function testPathSuffix()
+    {
+        $path = 'extra-path';
+        $types = ['config', 'temp', 'plugins', 'themes'];
+        foreach ($types as $type) {
+            $method = $type.'_path';
+            $this->assertEquals($method($path), app('path.'.$type).DIRECTORY_SEPARATOR.$path);
```
bennothommo:
```suggestion
            $method = $type . '_path';
            $this->assertEquals($method($path), app('path.' . $type) . DIRECTORY_SEPARATOR . $path);
```

tests/Support/HelpersTest.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+    public function testPathSuffixWithConfig()
+    {
+        $path = 'extra-path';
+        $types = ['uploads', 'media'];
+        foreach ($types as $type) {
+            $method = $type.'_path';
+            $config = 'cms.storage.'.$type.'.path';
+            $this->assertEquals($method($path), Config::get($config, app('path.'.$type)).DIRECTORY_SEPARATOR.$path);
```
bennothommo:
```suggestion
            $method = $type . '_path';
            $config = 'cms.storage.' . $type . '.path';
            $this->assertEquals($method($path), Config::get($config, app('path.' . $type)) . DIRECTORY_SEPARATOR . $path);
```

bennothommo (changes requested):

src/Support/helpers.php
mjauvin: Good idea, I'll submit code to do this, thanks!

tests/Foundation/ApplicationTest.php
```diff
@@ -0,0 +1,32 @@
...
+        $this->app = new Application($this->basePath);
+    }
+
+    public function testPathMethods()
+    {
+        $this->assertEquals($this->app->pluginsPath(), $this->app->buildPath('/plugins'));
+        $this->assertEquals($this->app->themesPath(), $this->app->buildPath('/themes'));
+        $this->assertEquals($this->app->tempPath(), $this->app->buildPath('/storage/temp'));
+        $this->assertEquals($this->app->uploadsPath(), $this->app->buildPath('/storage/app/uploads'));
+        $this->assertEquals($this->app->mediaPath(), $this->app->buildPath('/storage/app/media'));
```
LukeTowers: I'd like to see testing for the buildPath method, since the other methods are basically small wrappers around it, it's more likely that it would break without breaking these tests.

src/Support/helpers.php
mjauvin: @bennothommo Yes, all gonna be handled by the next commit...

src/Support/helpers.php
```diff
@@ -926,6 +964,7 @@ function title_case($value)
      */
     function resolve_path($path)
     {
+        $path = str_replace('/', DIRECTORY_SEPARATOR, $path);
```
mjauvin: @bennothommo I replaced slashes here as well, is it ok?

mjauvin:
I believe this is complete.

bennothommo:
@mjauvin one minor change - sorry - would you mind changing the `join_path` helper to a `join` method in the `October\Rain\Filesystem\PathResolver` class? I'm hesitant to add too many helpers because they're easily overwritten and pollute the global namespace.

mjauvin:
Done @bennothommo

src/Filesystem/PathResolver.php
```diff
@@ -114,6 +114,25 @@ public static function within($path, $directory)
         return starts_with($path, $directory);
     }
 
+    /**
+     * join two paths, making sure they use DIRECTORY_SEPARATOR
```
LukeTowers:
```suggestion
     * Join two paths, making sure they use DIRECTORY_SEPARATOR
```

src/Foundation/Application.php
```diff
@@ -456,6 +540,24 @@ public function getCachedPackagesPath()
...
+     */
+    public function buildPath($path, $prefix = null)
+    {
+        if (!isset($prefix)) {
+            $prefix = $this->basePath;
+        }
+        $prefix = str_replace('/', DIRECTORY_SEPARATOR, $prefix);
+        $path = str_replace('/', DIRECTORY_SEPARATOR, $path);
+
+        return rtrim($prefix, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . ltrim($path, DIRECTORY_SEPARATOR);
```
LukeTowers: Why not use the new ::join() helper method here instead of duplicating the normalization work?

bennothommo:
I'm going to clean this up a little before merging this - I think some of the functionality of the `join` method could be put into the `normalizePath` method, and that method could be adjusted a little as well. Should hopefully have it done today.

mjauvin:
@bennothommo you might want to add unit tests for the new join() method as well while you're at it.

bennothommo:
@mjauvin will do!

src/Foundation/Application.php
```diff
@@ -182,7 +208,66 @@ public function setThemesPath($path)
...
+    }
+
+    /**
+     * Set the temp path for the application.
+     *
+     * @return string
+     */
+    public function setTempPath($path)
+    {
+        $path = str_replace('/', DIRECTORY_SEPARATOR, $path);
```
LukeTowers: Is there something in the PathResolver @bennothommo @mjauvin that could replace this? It's a bit of a code smell to have a one-liner path cleaner piece of code copy-pasted across every method instead of using a standard method to handle the common goal of making sure the input will generate clean output or fail appropriately regardless of what that input is.

src/Support/helpers.php
bennothommo: @mjauvin The PathResolver already normalises the slashes, so it doesn't need this line. I'll change it up with my fixes.

src/Foundation/Application.php
mjauvin: @LukeTowers I've replaced with @bennothommo 's PathResolver::standardize() method calls.

bennothommo:
@mjauvin So it appears the best way to mock facades is to use the Mockery library which Laravel uses. I've ended up just fixing the tests fully this way. If you want to see how it's done, have a look starting from here: https://github.com/octobercms/library/pull/493/files#diff-46ffd9a10d01a153dc357c4ac0c5d599e13c4d95a90a6d2c4e81bdce551b4318R11

src/Support/helpers.php
```diff
@@ -165,14 +165,15 @@ function traceSql()
...
+     * Get the path to the config folder.
      *
      * @param  string  $path
      * @return string
      */
     function config_path($path = '')
     {
-        return app('path.config').($path ? '/'.$path : $path);
+        $config_path = app('path.config');
+        return PathResolver::join($config_path, $path);
```
LukeTowers: Why can't these functions be one-liners?

bennothommo (approved):

mjauvin:
@LukeTowers @bennothommo I don't see the point in converting the path helpers to one-liners, at least not for `uploads_path()` and `media_path()`

mjauvin:
Otherwise, I'd say it's looking pretty good now, what do you guys think?

LukeTowers:
@mjauvin yup, looking pretty good. What docs changes are required, and how would you describe these changes in the changelog?

mjauvin:
@LukeTowers https://github.com/octobercms/docs/pull/490

Changelog should only be:
- fixes uploads_path() broken helper method
- add media_path() helper method

mjauvin:
For reference, the plugins_path() helper method was doing this:
```
return app('path.plugins').($path ? '/'.$path : $path);
```
But `app('path.plugins')` did not exist.

LukeTowers:
How was it working previously then? Because I used plugins_path extensively in my own projects and I don't think I've ever had a problem with it?

mjauvin:
@LukeTowers Sorry, I meant to write "uploads_path()" method was broken... my bad.

mjauvin:
@LukeTowers @bennothommo Aren't the provided tests enough for "testing needed" ?

PR #495, opened 2020-05-24
Add data-request-validate to the controller scaffolding

LukeTowers:
Related: https://github.com/octobercms/october/pull/5088

### DIFF
--- a/src/Scaffold/Console/controller/create.stub
+++ b/src/Scaffold/Console/controller/create.stub
@@ -18,6 +18,7 @@
                 <button
                     type="submit"
                     data-request="onSave"
+                    data-request-validate
                     data-hotkey="ctrl+s, cmd+s"
                     data-load-indicator="Creating {{singular_name}}..."
                     class="btn btn-primary">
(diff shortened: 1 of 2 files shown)


bennothommo:
@LukeTowers I'd probably suggest leaving this out of controller scaffolding and making it opt-in.

LukeTowers:
@bennothommo why do you say that? In my mind having it run on every "submit" AJAX request is desirable and a better experience.

bennothommo:
@LukeTowers My opinion is that browser based validation is too intrusive, in that it steals focus. It also sometimes is too unclear on the issue at hand (when dealing with regex or number validation for example). I also imagine that people might get frustrated when they finally resolve all browser validation issues only to get another validation error from within October, for things that the browser validation cannot check (like duplicate slugs).

LukeTowers:
@bennothommo I see what you're saying. I would be more inclined to remove it from the scaffolding if the inline form validation feature was complete and enabled by default, but as it stands right now I think the browser-based client validation is more useful than harmful for the bulk of users given that the only other validation performed is one error at a time server side validation, which in my opinion is a worse user experience than browser-based.

PR #496, opened 2020-06-01
Add missing comments in controller scaffolding

Samuell1:
Added missing comments in controller scaffolding

### DIFF
--- a/src/Scaffold/Console/controller/controller.stub
+++ b/src/Scaffold/Console/controller/controller.stub
@@ -8,14 +8,32 @@ use Backend\Classes\Controller;
  */
 class {{studly_name}} extends Controller
 {
+    /**
+     * @var array Extensions implemented by this controller.
+     */
     public $implement = [
         'Backend.Behaviors.FormController',
         'Backend.Behaviors.ListController'
     ];
 
+    /**
+     * @var array `FormController` configuration.
+     */
     public $formConfig = 'config_form.yaml';
+
+    /**
+     * @var array `ListController` configuration.
+     */
     public $listConfig = 'config_list.yaml';
 
+    /**
+     * @var array Permissions required to view this page.
+     */
+    public $requiredPermissions = ['{{lower_author}}.{{lower_plugin}}.{{lower_name}}'];
+
+    /**
+     * Constructor.
+     */
     public function __construct()
     {
         parent::__construct();
--- a/src/Scaffold/Console/controller/controller.stub
+++ b/src/Scaffold/Console/controller/controller.stub
@@ -22,7 +22,7 @@ class {{studly_name}} extends Controller
     public $formConfig = 'config_form.yaml';
 
     /**
-     * @var array `ListController` configuration.
+     * @var string Configuration file for the `ListController` behavior.
      */
     public $listConfig = 'config_list.yaml';
 
--- a/src/Scaffold/Console/controller/controller.stub
+++ b/src/Scaffold/Console/controller/controller.stub
@@ -31,9 +31,6 @@ class {{studly_name}} extends Controller
      */
     public $requiredPermissions = ['{{lower_author}}.{{lower_plugin}}.{{lower_name}}'];
 
-    /**
-     * Constructor.
-     */
     public function __construct()
     {
         parent::__construct();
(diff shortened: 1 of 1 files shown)


src/Scaffold/Console/controller/controller.stub
```diff
@@ -8,14 +8,29 @@ use Backend\Classes\Controller;
  */
 class {{studly_name}} extends Controller
 {
+    /**
+     * @var array Extensions implemented by this controller.
```
bennothommo, line 12:
```suggestion
     * @var array Behaviors that are implemented by this controller.
```

src/Scaffold/Console/controller/controller.stub
```diff
@@ -8,14 +8,29 @@ use Backend\Classes\Controller;
...
+     */
     public $implement = [
         'Backend.Behaviors.FormController',
         'Backend.Behaviors.ListController'
     ];
 
+    /**
+     * @var array `FormController` configuration.
```
bennothommo, line 20:
```suggestion
     * @var string Configuration file for the `FormController` behavior.
```

src/Scaffold/Console/controller/controller.stub
```diff
@@ -8,14 +8,29 @@ use Backend\Classes\Controller;
...
+     */
     public $formConfig = 'config_form.yaml';
+
+    /**
+     * @var array `ListController` configuration.
```
bennothommo, line 25:
```suggestion
     * @var string Configuration file for the `ListController` behavior.
```

src/Scaffold/Console/controller/controller.stub
```diff
@@ -8,14 +8,29 @@ use Backend\Classes\Controller;
...
+     */
     public $listConfig = 'config_list.yaml';
 
+    /**
+     * @var array Permissions required to view this page.
```
bennothommo, line 30:
```suggestion
     * @var array Permissions required to view this controller's pages.
```

bennothommo (dismissed):

src/Scaffold/Console/controller/controller.stub
Samuell1: Not sure if this is good if core uses same descriptions. For example like here: https://github.com/octobercms/october/blob/439ed2cc33a1c4ec37a862d640ab7736fd193325/modules/backend/controllers/Users.php

LukeTowers: "Behaviors" is more accurate, we'll use it instead.

src/Scaffold/Console/controller/controller.stub
LukeTowers, line 30:
```suggestion
     * @var array Permissions required to access this controller's actions.
```

src/Scaffold/Console/controller/controller.stub
```diff
@@ -8,14 +8,29 @@ use Backend\Classes\Controller;
...
+
+    /**
+     * @var string Configuration file for the `ListController` behavior.
+     */
     public $listConfig = 'config_list.yaml';
 
+    /**
+     * @var array Permissions required to access this controller's actions.
+     */
+    public $requiredPermissions = ['{{lower_author}}.{{lower_plugin}}.{{lower_name}}'];
```
LukeTowers, line 32: This shouldn't be included since that would require them to register a permission in the form of author.plugin.controller which isn't obvious to newcomers and would block access to the controller to anyone not a superadmin.

Samuell1: But without permission anybody can access this and new people dont know that and that makes it as security issue no? And maybe if we add this we should add default permission for plugin that is like `plugin.name.*`.

LukeTowers: Not anybody, only authenticated backend users. I wouldn't be opposed to a wildcard necessarily, but it still presents a problem because default scaffolded plugins don't provide any permissions.  

If we also make a change to the plugin scaffolding to have the plugins register a default permission of `author.plugin.manage_plugin` and then check for that here that should be fine.

LukeTowers: Check for a wildcard that is

Samuell1: @LukeTowers i think for now i deleted(or you can) this to merge this PR and then later i make another PR with changes to permissions.

Samuell1: @LukeTowers removed, later i make separate PR for it, PR should be free to merge.

PR #497, opened 2020-06-03
Fix transChoice broken with October build 466

aurelien-roy:
October 466 introduced a regression in our install.
The transChoice filter stopped working (the ":count" placeholder is no longer replaced on strings).

Not sure what caused the regression, but copying this line from Laravel's Translator implementation fixes the issue.


LukeTowers:
@aurelien-roy my bad on the copy job that's what caused the regression, didn't catch that line. Thanks for the catch!

PR #498, opened 2020-06-14
Make File paths more unique

RomainMazB:
I've faced a problem where the temp path was shared between many files at each runtime! It couldn't be a coincidence after 5/6 refresh.

To reproduce (at least one way, the one I've followed to ran into this):
Just loop calling the getThumb function:
```twig
    {% for file in files %}
        getPath for {{ file.getFilename() }}: {{ file.getPath() }}<br>
        getThumb for {{ file.getFilename() }}: {{ file.getThumb(100, 100) }}
    {% endfor %}
```
returns:
> getPath for capture-decran-de-2020-06-09-**00-12-20**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/5ee6590a85a37857751894.png
> getThumb for capture-decran-de-2020-06-09-**00-12-20**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/thumb__100_100_0_0_auto.png
___
> getPath for capture-decran-de-2020-06-09-**13-29-22**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/5ee6590a85c5a110138871.png
> getThumb for capture-decran-de-2020-06-09-**13-29-22**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/thumb__100_100_0_0_auto.png
___
> getPath for capture-decran-de-2020-06-09-**13-35-48**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/5ee6590a85d8c323265258.png
> getThumb for capture-decran-de-2020-06-09-**13-35-48**.png: //localhost:3000/storage/app/uploads/public/**5ee/659/0a8**/thumb__100_100_0_0_auto.png

The problem is that uniqid is based on the current time in microseconds, it's mentioned that it may return a non-unique id but if you try to run many uniqid() "simultaneously", the first characters would be the same. The end of the generated string is more "unique" than the start.

Look a this little test, which in the function would have generate 20 similar paths!

maz@Maz:~/websites/me$ php artisan tinker
Psy Shell v0.9.12 (PHP 7.4.6 — cli) by Justin Hileman
> for($i = 0; $i < 20; $i++) { echo uniqid(null, true). "\n"; }
**5ee665636***679d2.80259532
**5ee665636**683c3.60840028
**5ee665636**68548.61747877
**5ee665636**68729.93979041
**5ee665636**68885.10867562
**5ee665636**689a8.15563143
**5ee665636**68b38.44254708
**5ee665636**68ca8.77327403
**5ee665636**68e07.04888608
**5ee665636**68f53.45817785
**5ee665636**690a9.49324510
**5ee665636**69205.34624093
**5ee665636**69375.77893116
**5ee665636**695c1.55307522
**5ee665636**69777.13772748
**5ee665636**698f8.77848691
**5ee665636**69a61.45040098
**5ee665636**69c76.80682655
**5ee665636**69e32.28284276
**5ee665636**69f65.47962678

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -708,7 +708,7 @@ class File extends Model
             $ext = $this->data->guessExtension();
         }
         
-        $name = Str::random();
+        $name = strtolower(Str::random(22));
 
         return $this->disk_name = !empty($ext) ? $name.'.'.$ext : $name;
     }
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -3,7 +3,6 @@
 use Cache;
 use Storage;
 use File as FileHelper;
-use October\Rain\Support\Str;
 use October\Rain\Network\Http;
 use October\Rain\Database\Model;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
@@ -551,7 +550,32 @@ class File extends Model
     public function getThumbFilename($width, $height, $options)
     {
         $options = $this->getDefaultThumbOptions($options);
-        return 'thumb_' . $this->id . '_' . $width . '_' . $height . '_' . $options['offset'][0] . '_' . $options['offset'][1] . '_' . $options['mode'] . '.' . $options['extension'];
+
+
+
+        $optionsAsString = array_map(function ($option) use ($options) {
+        // Remove falses (such as the default interlace) and extension
+            if ((is_bool($options[$option]) && ! $options[$option]) || $option === 'extension') {
+                return false;
+            }
+
+            // Handle options which need an array (offset)
+            if (is_array($options[$option])) {
+                return implode('_', $options[$option]);
+            }
+
+            return $options[$option];
+        }, array_keys($options));
+
+        $optionsInFilename = implode('_', array_filter($optionsAsString, function ($option) {
+            return ($option);
+        }));
+
+        // Remove double underscore when using
+        $optionsInFilename = str_replace('__', '_', $optionsInFilename);
+
+        $diskNameWithoutExtension = str_ireplace('.'.$this->getExtension(), '', $this->getDiskName());
+        return $diskNameWithoutExtension .'_thumb_' . $optionsInFilename . '_' . $width . '_' . $height . '.' . $options['extension'];
     }
 
     /**
@@ -664,13 +688,18 @@ class File extends Model
      */
     public function deleteThumbs()
     {
-        $pattern = 'thumb_'.$this->id.'_';
+        $diskNameWithoutExtension = str_ireplace('.'.$this->getExtension(), '', $this->disk_name);
+
+        $patterns = [
+            'thumb_'.$this->id.'_',
+            $diskNameWithoutExtension.'_thumb_'
+        ];
 
         $directory = $this->getStorageDirectory() . $this->getPartitionDirectory();
         $allFiles = $this->storageCmd('files', $directory);
         $collection = [];
         foreach ($allFiles as $file) {
-            if (starts_with(basename($file), $pattern)) {
+            if (starts_with(basename($file), $patterns)) {
                 $collection[] = $file;
             }
         }
@@ -702,13 +731,13 @@ class File extends Model
         }
 
         $ext = strtolower($this->getExtension());
-        
+
         // If file was uploaded without extension, attempt to guess it
         if (!$ext && $this->data instanceof UploadedFile) {
             $ext = $this->data->guessExtension();
         }
-        
-        $name = strtolower(Str::random(22));
+
+        $name = str_replace('.', '', uniqid(null, true));
 
         return $this->disk_name = !empty($ext) ? $name.'.'.$ext : $name;
     }
@@ -966,12 +995,12 @@ class File extends Model
     }
 
     /**
-    * Generates a partition for the file.
-    * return /ABC/DE1/234 for an name of ABCDE1234.
-    * @param Attachment $attachment
-    * @param string $styleName
-    * @return mixed
-    */
+     * Generates a partition for the file.
+     * return /ABC/DE1/234 for an name of ABCDE1234.
+     * @param Attachment $attachment
+     * @param string $styleName
+     * @return mixed
+     */
     protected function getPartitionDirectory()
     {
         return implode('/', array_slice(str_split($this->disk_name, 3), 0, 3)) . '/';
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -973,7 +973,7 @@ class File extends Model
     */
     protected function getPartitionDirectory()
     {
-        return implode('/', array_slice(str_split($this->disk_name, 3), 0, 3)) . '/';
+        return implode('/', array_slice(str_split(substr($this->disk_name, 13), 3), 0, 3)) . '/';
     }
 
     /**
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -3,6 +3,7 @@
 use Cache;
 use Storage;
 use File as FileHelper;
+use October\Rain\Support\Str;
 use October\Rain\Network\Http;
 use October\Rain\Database\Model;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
@@ -707,7 +708,7 @@ class File extends Model
             $ext = $this->data->guessExtension();
         }
         
-        $name = str_replace('.', '', uniqid(null, true));
+        $name = Str::random();
 
         return $this->disk_name = !empty($ext) ? $name.'.'.$ext : $name;
     }
@@ -973,7 +974,7 @@ class File extends Model
     */
     protected function getPartitionDirectory()
     {
-        return implode('/', array_slice(str_split(substr($this->disk_name, 13), 3), 0, 3)) . '/';
+        return implode('/', array_slice(str_split($this->disk_name, 3), 0, 3)) . '/';
     }
 
     /**


LukeTowers:
I'm pretty sure this is a breaking change.

RomainMazB:
I don't think so:
- The filename and the path are generated when the File model is created. So all the file already created and saved will not be affected by this change. Only the new ones.

- Since it's a random filename and path, no one could have created an eventual custom function which guess them.

- This function is only used to generate temporary short lived file, AFAIK and after some searches it's never used and stored anywhere to save long lived file?


And finally even if it results to a breaking change, even if my PR is finally closed, the idea behind this should not be "just abandoned".

The problem I've faced is real, and I'm surprised nobody noticed this before? All I wanted to do is to display thumbs in a loop and the way the paths are generated just stuck me at this point.

LukeTowers:
@RomainMazB I don't disagree with you that the generation of the disk_name should be fixed if it's not unique enough, but this is definitely a breaking change. The disk_name property is where the file is stored in the filesystem and is persisted to the DB after file creation, it remains set. `getPartitionDirectory()` is used by everything from thumb generation, to getPath() to getPublicPath().

The correct place to put such a change would be in `getDiskName()`, perhaps by switching to something powered by str_random() (filtering for only filesystem safe characters of course). That way the change will only affect new files, which is what is required to avoid breaking changes.

RomainMazB:
Yeah, a more in-depth observation made me understand your thoughts.
I will provide a new way to generate a more random filename (which will cascade to a more random file path)

RomainMazB:
Using Str::random() seemed to me to be an accurate way to achieve this: even if we don't need cryptographically security, we keep it simple using a library supported by the project.

LukeTowers:
@RomainMazB you can pass how long you'd like the string to be to (so make it `Str::random(22)`). Please also convert to lowercase to ensure there aren't any issues with case sensitive file systems.

RomainMazB:
@luketowers I knew about the possibility to choose the length of the string but I said to myself it wasn't necessary

LukeTowers:
@bennothommo any thoughts on this one?

LukeTowers:
@RomainMazB Actually, after looking at this further, it looks like the problem is actually that "$this->id" is not being inserted into the file path for the thumb (https://github.com/octobercms/library/blob/develop/src/Database/Attach/File.php#L553 & https://github.com/octobercms/library/blob/develop/src/Database/Attach/File.php#L666).

The uniqid() works well enough for unique filenames by having the full entropy for the actual filename, and the first 9 characters are solely for sharding purposes; so it's not necessarily a problem that multiple files would show up under the same third level directory. The real problem is that all the thumbs for the files in a given directory will be named the same because they don't have their unique ID included in their thumb filename. 

If we can't get the ID working in that file path (which is going to be tricky to, because if we fix that all existing thumbs that were missing the ID will no longer be detected as existing triggering regeneration of thumb files for every site that uses thumbnails and updates to whatever version we fix it in).

Another thing to consider is that we'd like the API for File thumbs and thumbs generated for non-File managed files (think media library) to have similar naming conventions, so in the case of resized Media images their filename might have to be something like _thumb_md5($path)_width_height_other_options.

Also, while we're at it we should probably change the filename generation to include all option values, not just the whitelisted ones.

RomainMazB:
I'll take a deeper look into these changes probably tomorrow (I'll fix the rich editor tonight and will not have the time to do more).

I've thank about another easy way to fix: just reverse the generated string, but this won't give us the "enhancements" you want.

LukeTowers:
Unfortunately I think reversing it would actually make the problem worse as you would now have collisions across requests as the end of the string is basically the output of microtime, which is a much smaller space for collisions to come into play.

bennothommo:
@LukeTowers I'm happy to see this fixed - the inconvenience of having to regenerate the thumbnails trumps the possibility of thumbnail collisions, whichever way we go.

Samuell1:
Best to make the folder names unique will be hashing ID with some salt that is per application.

LukeTowers:
str_random is unique enough @Samuell1, although ideally we don't even need to do that. The root of the problem is colliding thumb filenames, we just need to fix that and then even just uniqid() is fine for our purposes.

bennothommo:
@Samuell1 

> Best to make the folder names unique will be hashing ID with some salt that is per application.

In fairness, this is the `storage/public` folder we're talking about here, we don't need to go overboard with randomisation ;)

LukeTowers:
Just to add to what @bennothommo said, the only reason we have random folder names in the first place is to have a somewhat random / even sharding of the file system to support a large number of files. So it doesn't really matter how random per se the filenames are, just that there's a relatively equal chance for any given sequence to be generated to distribute the files evenly amongst the shards (directory) to improve performance on some systems that don't handle large numbers of files under single directories very well.

RomainMazB:
> _thumb_md5($path)_width_height_other_options.

Just to be sure:
- I rollback the name to use uniqid
- I make sure the original (uniqid) filename is inserted into the thumb filename.
- I add all options in the filename.

> Another thing to consider is that we'd like the API for File thumbs and thumbs generated for non-File managed files (think media library) to have similar naming conventions

And to add an (maybe useless) information: my actual case is exactly this one. I'm creating a plugin to use the Media Manager as a frontend gallery, without the need to create database model. The way I'm using the `File` class as an API is like: 
```php
$file = (new File)->fromFile($mediaManagerFilePath);
```
Doing this way I can then use the `getThumb` method in my partial.

Don't know if it's the best way to do but this make me ask this question:
Do the `md5($filename)` is needed since it already exists a way to do without creating any tricky method to handle media manager's files and adopt a different filename convention (it differs from the uniqid)? :thinking:

LukeTowers:
@RomainMazB if all you need is a thumbnail of a media item why not use the `| resize()` filter provided by the image resizing plugins on the marketplace? (note that we're planning on integrating it into core at some point).

> Do the md5($filename) is needed since it already exists a way to do without creating any tricky method to handle media manager's files and adopt a different filename convention (it differs from the uniqid)? 🤔

Not sure exactly what you're asking, but I suggested that rather than `disk_name` because if it's md5 then both File models and all other resized images not attached to File models could have similar naming schemes for thumbnails.

RomainMazB:
@LukeTowers ~~the question was...~~ I've erased my text because while I was writing it, I've understood: the filename of resized images not attached to File models are using md5 and you want me to use md5 instead of uniqid to generated thumbs from File models. Let me know if I've misunderstood.

And for your first question about the `| resize()` filter is because I dislike using third party plugin, external libraries or other dependency to do something I can achieve "easily" without them. But that's another story.

LukeTowers:
Fair enough @RomainMazB, although in this case it's merely a temporary dependency as the | resize filter will be added into core soon with the same API.

LukeTowers:
@RomainMazB ah, I see further the problem you're having. This is actually a result of the way you're trying to [ab]use the File models. Because the thumb method assume there to be an ID (and this is a problem with the resize filters provided by third party plugins too), there is a conflict because there is nothing unique in the thumb file itself since your File models do not actually have an ID because they were never added to the system_files table.

RomainMazB:
> @RomainMazB ah, I see further the problem you're having. This is actually a result of the way you're trying to [ab]use the File models. Because the thumb method assume there to be an ID (and this is a problem with the resize filters provided by third party plugins too), there is a conflict because there is nothing unique in the thumb file itself since your File models do not actually have an ID because they were never added to the system_files table.

Wow! So this issue is not really an issue? Do I still need to "fix" this edge case though?

LukeTowers:
@RomainMazB probably would be a good idea to switch to using the disk name instead of the ID in the thumb generation.

LukeTowers:
maybe switch it to `$this->getDiskName() . '_thumb_' . etc`

RomainMazB:
> maybe switch it to `$this->getDiskName() . '_thumb_' . etc`

Using this, it works, but what about breaking change? This will crash the `deleteThumb` method:
```php
public function deleteThumbs()
    {
        $pattern = 'thumb_'.$this->id.'_';

        $directory = $this->getStorageDirectory() . $this->getPartitionDirectory();
        $allFiles = $this->storageCmd('files', $directory);
        $collection = [];
        foreach ($allFiles as $file) {
            if (starts_with(basename($file), $pattern)) {
                $collection[] = $file;
            }
        }
...
```
I could provide a fix for that, basically (starts_with allows us to pass an array of needles):
```php
$patterns = [
    'thumb_'.$this->id.'_',
    'thumb_'.$this->getDiskName()
];
```

LukeTowers:
Sure, although note that my suggested new filename starts with the disk_name, it doesn't replace ID with disk name

RomainMazB:
Didn't see the details ;)

Ok, I will commit in a minute.

RomainMazB:
It's perfectly working but I'm facing a weird behavior I still don't understand.
To display my thumb, I'm generating actually two thumbs, to make a lazy load:
```twig
{% set img = file.getThumb(width, height, {mode: 'crop'}) %}
{% set lazyImg = file.getThumb(width/10|round, height/10|round, {mode: 'crop'}) %}
{% set path = file.getPath() %}
<a href="{{ path }}">
                <img src="{{ lazyImg }}" srcset="{{ img }}">
</a>
```
But the paths the two thumbs are stored in different place o.0
```html
<a href="http://me.test/storage/app/uploads/public/5ee/be7/264/5eebe7264e5b5672230156.png" >
    <img src="http://me.test/storage/temp/public/1ac/89a/3b6/5eebe7264e5b5672230156_thumb_crop_0_0_90_26_26__26.png" srcset="http://me.test/storage/app/uploads/public/5ee/be7/264/5eebe7264e5b5672230156_thumb_crop_0_0_90_268_268.png">
</a>
```
It don't create any issue but I wanted to store them in the same directory but can't achieve that...

RomainMazB:
@LukeTowers any idea to solve that?

I'll try to have a deeper (at least fresher) look on this. I didn't understand how the generated path (which should be generated from the file name) can be different from the file's name?

LukeTowers:
@RomainMazB what's the current state of this, especially when taking into account the recent addition of core support for the `| resize` filter?

PR #499, opened 2020-06-18
All non-null values should be encrypted

msimkunas:
The `Encryptable` trait does not encrypt empty-like values, including `0`. However, zero values may be legitimate data that should be encrypted (money amounts, for example). This PR restricts the non-encryptable values to `null` only.

Additionally, this PR fixes a bug where trying to get an encryptable attribute with a value of `0` would throw `The payload is invalid` error.


bennothommo:
@msimkunas Would you mind providing a test case for the change?

msimkunas:
I've added the test case.

I also had to create a fake implementation of the `Crypt` class in it in order to be able to proceed with the tests. Unfortunately, it is not possible to mock the `Crypt` facade when testing this library because the Laravel app is not bootstrapped properly.

LukeTowers:
@msimkunas can you not use `\App::make('encryptor');`?

msimkunas:
@LukeTowers Perhaps. Although I feel that using a fake implementation of the encryptor is sufficient here because we don't need the encryption overhead in such a simple test case.

msimkunas:
The main issue here is that the `Encryptable` trait uses the `Crypt` facade directly therefore we need to somehow mock it in order for it to be resolved. I found that it is easier to go around this issue by simply creating a fake `Crypt` implementation.

By the way, running `\App::make('encryptor')` in the test case class throws an error `Class 'App' not found`.

bennothommo:
@msimkunas I've adjusted your test case to test against a true `Encrypter` instance - basically, the same thing that the `Crypt` facade would use. This way, we can be sure it is correctly working with the different values.

@LukeTowers I had to bring in the `illuminate/encryption` library as a dev dependency for this, however, it should be fully backwards compatible. Let me know if there's any issues.

PR #501, opened 2020-06-23
Feature/relation count fix

Klaasie:
This PR fixes the bug report that I wrote here: https://github.com/octobercms/october/issues/5134.

This PR includes the fix written for laravel 7.x which can be found here: https://github.com/laravel/framework/pull/29925/files.

In addition I've copied all of the tests concerning the with count statement and modified them to be relevant for Laravel 5.5 / OctoberCMS.

The fix makes sure that any of the order by statements that may have been added through global scopes are removed from the query before it is being executed. This while maintaining anything else that might be added through global scopes (E.g. stuff like softdelete).
Since the count method always returns one result sorting makes no sense anyway therefor this PR should be fully backwards compatible while fixing an issue where MySQL would give an error if it is set in strict mode.


bennothommo:
Thanks heaps @Klaasie. Would you mind just fixing the code formatting issues reported here for us? https://github.com/octobercms/library/pull/501/checks?check_run_id=798335748

Klaasie:
@bennothommo Done!

src/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -0,0 +1,69 @@
+<?php namespace October\Rain\Database\Eloquent\Concerns;
+
+use Illuminate\Database\Eloquent\Concerns\QueriesRelationships as IlluminateQueriesRelationships;
+use Illuminate\Support\Str;
+
+trait QueriesRelationships
+{
+    use IlluminateQueriesRelationships {
+        IlluminateQueriesRelationships::withCount as IlluminateWithCount;
```
bennothommo: @Klaasie is there any need for this alias, given that you don't use it within your new `withCount` method?

src/Database/Builder.php
```diff
@@ -14,6 +14,8 @@
  */
 class Builder extends BuilderModel
 {
+    use Eloquent\Concerns\QueriesRelationships;
```
bennothommo: @Klaasie while this is technically correct, our current convention is to use the fully namespaced trait name, just to make it clear whether it's an October or Laravel class.

src/Database/Eloquent/Concerns/QueriesRelationships.php
Klaasie: I guess given the circumstances the alias doesn't make a lot of sense. I thought adding the alias would provide developers with the opportunity to use the Illuminate version of the method if the need may be.

But since there's no difference in results and this actually fixes a potential error I guess it could be left out. Would you want me to make that modification?

bennothommo: Yep, I reckon just remove it. :)

src/Database/Builder.php
Klaasie: I've changed the statement :)

src/Database/Eloquent/Concerns/QueriesRelationships.php
Klaasie: Done!

src/Database/Builder.php
```diff
@@ -14,6 +14,8 @@
  */
 class Builder extends BuilderModel
 {
+    use \October\Rain\Database\Eloquent\Concerns\QueriesRelationships;
```
LukeTowers: This should be namespaced to `\October\Rain\Database\Concerns\QueriesRelations;` (remove the Eloquent)

src/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -0,0 +1,66 @@
+<?php namespace October\Rain\Database\Eloquent\Concerns;
+
+use Illuminate\Support\Str;
```
LukeTowers: We should use our Str helper, not Laravel's.

src/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -0,0 +1,66 @@
...
+
+trait QueriesRelationships
+{
+    use \Illuminate\Database\Eloquent\Concerns\QueriesRelationships;
```
LukeTowers: Should this be a use or an extend?

src/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -0,0 +1,66 @@
...
+                $relation->getRelated()->newQuery(),
+                $this
+            );
+
+            $query->callScope($constraints);
+
+            $query = $query->mergeConstraintsFrom($relation->getQuery())->toBase();
+
+            $query->orders = null;
+            $query->setBindings([], 'order');
```
LukeTowers: A comment added to explain these lines would be helpful for future review, perhaps even link to the issue / PR discussing it.

src/Database/Eloquent/Concerns/QueriesRelationships.php
Klaasie: Traits can't be extended so you have to use "use".

Actually, in this particular case you might argue that the "use" is not necessary at all since the trait I wrote overwrites the `withCount` method on the october model level where the eloquent trait is used on the eloquent model. The tests prove that it still works.

However, I'd opt to leave the "use" statement as it best clarifies what we're doing here (which is override one specific method of the QueriesRelationships trait.

src/Database/Builder.php
Klaasie: I've moved the file and modified the namespace

src/Database/Eloquent/Concerns/QueriesRelationships.php
Klaasie: Modified the Str helper used to the October one :)

src/Database/Eloquent/Concerns/QueriesRelationships.php
Klaasie: I've added comment that I feel explains what's happening with a link that will give context why. Let me know if this will suffice.

src/Database/Eloquent/Concerns/QueriesRelationships.php
LukeTowers: Fair enough, although does the "double use" cause any issues with performance or class resolving of any kind?

Klaasie: Right, so I've tested the performance with the most popular method I found on google which is simply creating a lot of instances through a for loop. and measure the begin and end time.

I saw no noticeable differences creating 100 000 model instances. All results varied between 0.13 and 0.16 seconds.
If there is a better way to go about this let me know :)

Traits can't be resolved so I can't think of any issues there. 

I walked through the method I implemented again and did notice that the method does use `getRelationWithoutConstraints` and `mergeConstraintsFrom`, both which are part of the "parent" trait.
So removing the use statement could theoratically make the code break in that case (e.g. the eloquent version not being used). Even if it's unlikely to happen.

Klaasie:
Any ETA on when this will be tested / accepted? :)

LukeTowers:
@bennothommo what was left to review on this?

bennothommo:
@LukeTowers I think this just needed to be tested - but since it's got unit tests, it's probably good to go.

@Klaasie just one more change requested - would you mind changing the unit tests to use `DbTestCase` (https://github.com/octobercms/library/blob/develop/tests/DbTestCase.php)? The unit test schema should be set up similar to the following: https://github.com/octobercms/library/blob/develop/tests/Database/Traits/EncryptableTest.php

Klaasie:
@bennothommo Done!

daftspunk:
Introducing this in 1.1 since it will stay on Laravel 6 

Thanks @Klaasie 😃

PR #502, opened 2020-07-08
Fix file attachments eager load with PostgreSQL

RomainMazB:
See #5200 https://github.com/octobercms/october/issues/5200

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -28,7 +28,7 @@ class File extends Model
      * Relations
      */
     public $morphTo = [
-        'attachment' => []
+        'attachment' => ['keyType' => 'string']
     ];
 
     /**
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -28,7 +28,9 @@ class File extends Model
      * Relations
      */
     public $morphTo = [
-        'attachment' => ['keyType' => 'string']
+        'attachment' => [
+            'keyType' => 'string'
+        ],
     ];
 
     /**
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -687,7 +687,7 @@ trait HasRelationships
 
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
+        return new AttachMany($instance->newQuery(), $this, $table . '.' . $type, $table . '.' . $id, $isPublic, $localKey, $relationName, $keyType);
     }
 
     /**
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -660,7 +660,7 @@ trait HasRelationships
 
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
+        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
     }
 
     /**
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -660,7 +660,7 @@ trait HasRelationships
 
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
+        return new AttachOne($instance->newQuery(), $this, $table . '.' . $type, $table . '.' . $id, $isPublic, $localKey, $relationName, $keyType);
     }
 
     /**
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -653,9 +653,14 @@ trait HasRelationships
 
         $table = $instance->getTable();
 
+        $keyType = data_get(
+            $instance->getRelationDefinitions(),
+            'morphTo.attachment.keyType'
+        );
+
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName);
+        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
     }
 
     /**
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -675,9 +675,14 @@ trait HasRelationships
 
         $table = $instance->getTable();
 
+        $keyType = data_get(
+            $instance->getRelationDefinitions(),
+            'morphTo.attachment.keyType'
+        );
+
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName);
+        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
     }
 
     /**
--- a/src/Database/Relations/AttachMany.php
+++ b/src/Database/Relations/AttachMany.php
@@ -27,6 +27,13 @@ class AttachMany extends MorphManyBase
 
         $this->public = $isPublic;
 
+        /**
+         *  Since Laravel 5.7, whereInMethod has been added
+         *  https://github.com/illuminate/database/blob/5.7/Eloquent/Relations/Relation.php#L317-L324
+         *  Because of that, the attachMany relationships, which relies on a string key 'attachment_id' must be set to the right type
+         *  Without that, it would call the whereIn method and fails on eagerLoad on some strict DBMS (occurred on PostgreSQL)
+         *  It has been set to string as default into the \Database\Attach\File class but could be overridden by the user
+         */
         if ($keyType !== null) {
             $parent->setKeyType($keyType);
         }
--- a/src/Database/Relations/AttachMany.php
+++ b/src/Database/Relations/AttachMany.php
@@ -12,7 +12,14 @@ class AttachMany extends MorphManyBase
 
     /**
      * Create a new has many relationship instance.
-     * @return void
+     * @param Builder $query
+     * @param Model $parent
+     * @param $type
+     * @param $id
+     * @param $isPublic
+     * @param $localKey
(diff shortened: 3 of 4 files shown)


LukeTowers:
The problem should not be solved here, this is a hack. The real problem is that deeper in the internals it's treating a foreign key as if it's inherently an integer when it isn't. If you can find that, then that's what needs to be fixed.

LukeTowers:
Really what needs to happen is that the Database\Attach\File Model overrides whereInMethod() to always return whereIn instead of whereIntegerInRaw since the foreign key on the system_files table is a VARCHAR, not an integer: 

https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/src/Illuminate/Database/Eloquent/Relations/Relation.php#L317-L323
 
Alternatively, maybe we need to add support for a configuration option in the relationship definitions that would call this method that would allow us to tell the relation trait that the given foreign key is actually not an integer, instead of the Laravel internals trying to automagically detect the type of foreign key by looking at the model it's referencing to see what key type it's using.

RomainMazB:
> Really what needs to happen is that the Database\Attach\File Model overrides whereInMethod() to always return whereIn instead of whereIntegerInRaw since the foreign key on the system_files table is a VARCHAR, not an integer:
> 
> https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/src/Illuminate/Database/Eloquent/Relations/Relation.php#L317-L323

I've seen this part of the code, but this is not a varchar: `getType` doesn't look into the database to see which type the `attachment_id` field is:
- it [looks into the model to see which type the key is](https://github.com/laravel/framework/blob/64d80c095785c0b65250797848f25f9ae65fb70f/src/Illuminate/Database/Eloquent/Model.php#L1437-L1453).
- [default is set to integer](https://github.com/laravel/framework/blob/64d80c095785c0b65250797848f25f9ae65fb70f/src/Illuminate/Database/Eloquent/Model.php#L60).
- and a global search in the project didn't found any override of this property.

Knowing that:
Another way to do would be to `setTypeKey` to string on the `RainLab\Database\Model` constructor to ensure it's not treated as an integer (just tested it and it works) but doing so I can't tell which bug could cascades from this modification.

*I've tried Laravel type casting, didn't work*

LukeTowers:
> This could be done also but the fact it only appears on PostgreSQL which is probably used by less than < 5% of the users, I feel like it should not be implemented as a "global parameter" which could be modified when unneeded. 

I don't feel like this is something that's specific to PostgreSQL, the problem is fundamentally that Eloquent is trying to explicitly use an integer comparison on a key field that may be a string. We should be able to override the Laravel magic through configuration of relationships so I think that option is fine. Something like

```php
$morphTo = ['attachment' => ['key' => 'attachment_id', 'keyType' => 'string']];
```

RomainMazB:
@LukeTowers 

I will test on a MySQL server and let you know the results.

I've just finished a site using MySQL and a ton of images using eager load, this is why I feel like it's only related to PostgreSQL. But I'm not totally sure if I was on the Laravel 6 branch on this project.

Another bad side to this implementation (using a new config key) is the fact we'll need to use `setKeyType` somewhere in the relation code which will be then applied to the model and reflect on all other relationship.

We could set the key type to string, apply the relation then revert the key type to the one previously used. But I feel like it's not less hacky than the way I've done.

LukeTowers:
@RomainMazB there's nothing saying that we have to call setKeyType, we can override the entire method to look at the relation config itself rather than the automagic that's currently happening in Laravel. @bennothommo what are your thoughts on this?

bennothommo:
@LukeTowers @RomainMazB hmm, I'm not sure I'm quite for a `keyType` definition in the relation config, if only because it means we'd have to define the same key type for every model that uses that related model in question. It would be more cleaner if it could be defined on the related model itself.

RomainMazB:
@bennothommo I will investigate in this way to see what is possible if @LukeTowers is ok with that: define the `keyType` in the related model (File in this case).

But still I guess I'll need to override the `getType` function to handle this modification:
- If the related model explicitely define a type, use it
- Else call `parent::getType()`

LukeTowers:
@bennothommo You wouldn't have to though. Laravel already allows the model to specify what key type is used by the related model by defining it on the model itself. The problem however is that the the system_files uses a string key `attachment_id` to be able to support both integer & string keys on related models. So regardless of what the Model referenced by the `attachment` relationship is using as it's key type, the local relationship column itself actually uses a different key type.

None of the models using the file model would have to define a key type, it's only the file model itself that would have to explicitly tell the framework "Hey, so I know that the Attachment relationship model says it's using an integer as it's key type, but listen, attachment_id is actually a varchar column, not an integer"

bennothommo:
@LukeTowers that was what I was meaning though - I'd rather not have to, for example, set every model that wants to use `File` as an attachment to have a `keyType` definition in its relation config - it would be better if it uses the `$keyType` definition in File itself to determine the correct query used. It looks like we're both on the same page on that, anyway.

LukeTowers:
@bennothommo yes, sounds like we are. It's confusing because a model that wants to add a relationship to the file model should use attachOne or attachMany relationships, but the relationship in the other direction is called "attachment" (File -> Source model, means that the Source Model is the File's "attachment").

See https://github.com/octobercms/library/blob/develop/src/Database/Attach/File.php#L31. That is where we would define keyType => string.

RomainMazB:
@LukeTowers Currently investigating on this:
I've solved the issue with the behavior you wanted: define the keyType from the File morphTo definition.

I'll review myself my code before committing this weekend. I wanted to let you know to not duplicate the work (because it's related to L6, you may want it to be solved before 468 and do it yourself).

RomainMazB:
Ok, here is the speech.

After adding the `keyType` property [inside the morphTo property](https://github.com/RomainMazB/library/blob/patch-3/src/Database/Attach/File.php#L31) in the parent model (File), we need [retrieve the `keyType` value](https://github.com/RomainMazB/library/blob/patch-3/src/Database/Concerns/HasRelationships.php#L683-L690) from the `attachOne` or `atttachMany` creation in the `hasRelationships` trait and pass it to the `attachMany` and `attachOne` classes which finally use this keyType to [bind the proper `keyType` on the parent model](https://github.com/RomainMazB/library/blob/patch-3/src/Database/Relations/AttachMany.php#L23-L25).

Using an optional keyType parameter on relationship, it should not break anything.

src/Database/Concerns/HasRelationships.php
```diff
@@ -653,9 +653,14 @@ public function attachOne($related, $isPublic = true, $localKey = null, $relatio
...
 
+        $keyType = data_get(
+            $instance->getRelationDefinitions(),
+            'morphTo.attachment.keyType'
+        );
+
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName);
+        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
```
LukeTowers: @RomainMazB typo?

RomainMazB:
Kind of: I was testing on frontend with the post `featured_images` relationship. So my I focused on attachMany, then I copy pasted the logic the same logic to attachOne but no attachOne was queried.

I saw the problem when I tried to go to the backend and the brand logo load was crashing all the backend.

Fixed and tested.

src/Database/Relations/AttachMany.php
```diff
@@ -14,12 +14,16 @@ class AttachMany extends MorphManyBase
      * Create a new has many relationship instance.
```
bennothommo, line 14: @RomainMazB Would you mind doing us a favour and adding the params to the docblock?

src/Database/Relations/AttachOne.php
```diff
@@ -14,12 +14,16 @@ class AttachOne extends MorphOneBase
      * Create a new has many relationship instance.
```
bennothommo, line 14: @RomainMazB Same here too :)

bennothommo:
Looks fine to me otherwise - not tested, though.

src/Database/Relations/AttachMany.php
RomainMazB: Yep no problem. It should have be done without the need to ask :grimacing:

LukeTowers: These need type data and ideally an explanation of their purpose @RomainMazB

RomainMazB: The fact no concise docblock was implemented anywhere disappointed me about how I was supposed to do.

LukeTowers: @RomainMazB ?

RomainMazB: I've added an explicit comment on both `AttachMany` and `AttachOne` file.

RomainMazB:
FYI. This still occurs on build 469, I've just added a more explicit comment. Hope it will be sufficient to close this PR.

vincentdesiree:
I had the same issue so i tested this. It worked for me.

src/Database/Concerns/HasRelationships.php
```diff
@@ -675,9 +680,14 @@ public function attachMany($related, $isPublic = null, $localKey = null, $relati
...
 
+        $keyType = data_get(
+            $instance->getRelationDefinitions(),
+            'morphTo.attachment.keyType'
+        );
+
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName);
+        return new AttachMany($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
```
bennothommo:
```suggestion
        return new AttachMany($instance->newQuery(), $this, $table . '.' . $type, $table . '.' . $id, $isPublic, $localKey, $relationName, $keyType);
```

src/Database/Concerns/HasRelationships.php
```diff
@@ -653,9 +653,14 @@ public function attachOne($related, $isPublic = true, $localKey = null, $relatio
...
 
+        $keyType = data_get(
+            $instance->getRelationDefinitions(),
+            'morphTo.attachment.keyType'
+        );
+
         $localKey = $localKey ?: $this->getKeyName();
 
-        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName);
+        return new AttachOne($instance->newQuery(), $this, $table.'.'.$type, $table.'.'.$id, $isPublic, $localKey, $relationName, $keyType);
```
bennothommo:
```suggestion
        return new AttachOne($instance->newQuery(), $this, $table . '.' . $type, $table . '.' . $id, $isPublic, $localKey, $relationName, $keyType);
```

src/Database/Attach/File.php
```diff
@@ -28,7 +28,7 @@ class File extends Model
      * Relations
      */
     public $morphTo = [
-        'attachment' => []
+        'attachment' => ['keyType' => 'string']
```
bennothommo:
```suggestion
        'attachment' => [
            'keyType' => 'string'
        ],
```

PR #503, opened 2020-07-14
String value should not be trimmed in the core.

RomainMazB:
Hi,
Discovered when experienced an [issue on RainLab\Forum](https://github.com/rainlab/forum-plugin/issues/145), I (and @alxy ) don't see why this rule is applied.

Why aren't we allowed to store string in database with some extra space? This should not be implemented here: if some system's or plugin's model want to trim extra space, it should be done locally in the end-model. Not in the core.


LukeTowers:
Original code was added https://github.com/octobercms/library/commit/684c796afcf763e6589722f077888a4ca728a25f

LukeTowers:
@RomainMazB after talking to @daftspunk & @bennothommo about this the ideal way forward is to provide a flag on the model properties to disable this behavior:

```php
   /**
     * @var bool Trim string attributes
     */
    public $trimStrings = true;
```

Default is true but models extending the base model can override the property and set it to false and handle attribute trimming their own way.

RomainMazB:
Reopen, I'm back
VJ8n

LukeTowers:
@RomainMazB could you add a unit test for both states? Enabled vs disabled

RomainMazB:
Sure!

bennothommo:
@RomainMazB just need to unit tests to be added before we consider this further.

RomainMazB:
@bennothommo Hope this is what you expected.

I don't practice tests so much.

bennothommo:
Looks good to me @RomainMazB - nice one!

@LukeTowers any feedback on this?

LukeTowers:
Seems fine to me, although I'm not 100% convinced that we should use it for fixing the related forum issue.

RomainMazB:
@LukeTowers I discovered it using the Forum plugin, but it's accurate everywhere in the CMS.

All the strings are trimmed in any attributes of any models (I guess I don't teach you anything here).
For some it's logic: I want my blog's posts title to be trimmed.

For others it's not: If I create a website like october-tricks, I don't want the tricks which just hold a piece of code to be trimmed (using markdown syntax with one line before and after the block, my code block would never work).

LukeTowers:
For the markdown examples I'm leaning more towards that should be a separate fix (perhaps prefixing with a comment before processing it) given that I don't necessarily want to be allowing untrimmed markdown content, but I'm open to feedback.

bennothommo:
Markdown should be "right-trimmed", in that any extra whitespace after the content should be stripped, but when Markdown does process whitespace at the beginning of a string in a particular way, that should be kept.

mjauvin:
@bennothommo two spaces at the end of a line is a line break in markdown

bennothommo:
@mjauvin indeed, you are correct. My mistake. ~~No trimming should be done on Markdown then.~~

That being said, when I mean right trimming, I mean any whitespace *after all content*. Spacing in between content will be kept.

RomainMazB:
@alxy any though on that?

_I'm tagging him because the starting point of this PR is a discussion with him_

alxy:
@daftspunk May I ask why this cannot be merged? It doesnt break anything, and in fact its really weird that trim is applied by default on all attributes.

daftspunk:
@alxy Sorry forgot to make reference to the commit 85ab729f3d3c97f2cc473cfa95d484b36fc419c1

Been meaning to add this for a while

RomainMazB:
@daftspunk any reason for not just having merged this PR if this is something you wanted to add for a while? 

Not a big deal but not being credited seems to be the rule here now 🤔

Peace ❄️

daftspunk:
@RomainMazB you may notice I was consulted on this piece of work and gave the instruction to create this variable, including the decision on its name. The PR had it in the wrong position and the test is not necessary for something as simple as this

Sorry that your PR went to waste but that happens sometimes

daftspunk:
This issue created some drama. This wasn't merged because it would just have been rolled back by my next commit. Can't merge it now anyway. Credit to @RomainMazB ans @alxy for helping and spending time on this. Thank you for your contributions to the project

PR #504, opened 2020-07-18
Http docblock asks for an array, when really what it wants is a callable

jjfmarket:
My editor barfed when I tried to pass an anonymous function here, and… the phpdoc expected array. While it the callable $options _could_ be an array, if it were using the callable literal style [$object, 'method'], it's more likely to be passed an anonymous function (like described in the example)


LukeTowers:
@jjfmarket thanks!

PR #508, opened 2020-07-28
Improvements to cookie value validation

LukeTowers:


LukeTowers:
I tested it and it invalidates all existing cookies except for the `october_session`, which then gets upgraded in the response to the request. This will ensure that users will remain logged in after they complete the update, although they will lose their persistence cookie if they have one and have to login again at some point.

PR #509, opened 2020-08-04
Add "resolve_path" helper method

bennothommo:
Adds a `resolve_path()` method that operates very similar to `realpath()` except in that it will allow missing files and folders. It will resolve relative directory markers and expand any symlinks encountered.

This can then be used in https://github.com/octobercms/october/pull/5229 to allow for the creation of missing subdirectories for assets whilst simultaneously ensuring that any files and directories remain within the theme directory.


bennothommo:
Code Quality test is just complaining about the file with a space in the filename - since we don't use that anywhere else, we can ignore it. It won't come up again after this is merged.

LukeTowers:
I'd feel more comfortable if we could fix that issue before we merge this @bennothommo, otherwise this LGTM!

PR #511, opened 2020-08-11
Fix one character discrepancy between limited HTML strings and CRLF/LF line breaks

bennothommo:
Linked to https://github.com/octobercms/library/pull/510 - this came up in the Windows unit tests. This PR should be merged in with that PR before merge.

Windows CRLF line breaks are considered to be two characters, whereas Unix LF line breaks are one. This causes a discrepancy between Windows and Linux for strings that use `HtmlBuilder::limit()` method, where HTML strings that are limited in Windows will be shorter than on Linux.

This adds a new method to `HtmlBuilder` which will get the "readable" length of a string and consider both line break types to one character, as well as any extra spacing in between characters which is rendered as a single space in HTML.

### DIFF
--- a/src/Html/HtmlBuilder.php
+++ b/src/Html/HtmlBuilder.php
@@ -484,7 +484,6 @@ class HtmlBuilder
      */
     protected static function getReadableLength($str)
     {
-        $str = str_replace(["\n", "\r"], "\n", $str);
         return strlen(preg_replace('/\s+/', ' ', $str));
     }
 
--- a/src/Html/HtmlBuilder.php
+++ b/src/Html/HtmlBuilder.php
@@ -428,14 +428,14 @@ class HtmlBuilder
             list($tag, $tagPosition) = $match[0];
 
             $str = substr($html, $position, $tagPosition - $position);
-            if ($printedLength + strlen($str) > $maxLength) {
+            if ($printedLength + self::getReadableLength($str) > $maxLength) {
                 $result .= substr($str, 0, $maxLength - $printedLength) . $end;
                 $printedLength = $maxLength;
                 break;
             }
 
             $result .= $str;
-            $printedLength += strlen($str);
+            $printedLength += self::getReadableLength($str);
             if ($printedLength >= $maxLength) {
                 $result .= $end;
                 break;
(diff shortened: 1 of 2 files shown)


src/Html/HtmlBuilder.php
```diff
@@ -474,6 +474,20 @@ public static function limit($html, $maxLength = 100, $end = '...')
...
+     *
+     * Considers multiple spaces and line breaks to be 1 character, regardless of OS.
+     *
+     * @param string $str
+     * @return int
+     */
+    protected static function getReadableLength($str)
+    {
+        $str = str_replace(["\n", "\r"], "\n", $str);
+        return strlen(preg_replace('/\s+/', ' ', $str));
```
LukeTowers, line 487: @bennothommo could you expand on why this is necessary? Also, doesn't this turn Windows line breaks into two unix linebreaks?

bennothommo: @LukeTowers as far as I can tell from the intent of the `HtmlBuilder::limit()` method, it's trying to limit the characters of a HTML string whilst allowing tags to be correctly used in the limited string.

I made this change to calculate the "readable" length of the string when determining where to crop the string. Take the following unit test:
```
<p>The quick brown fox jumped          over the lazy dog</p>\r\n<p>The quick brown fox jumped over the lazy dog</p>
```
This previously would've resulted in the string being trimmed a lot further, as each space character was considered to be one character, even though when it's actually rendered in the browser, it's just one space character as browsers ignore multiple whitespaces. Simultaneously with the CRLF line break (`\r\n`) being two characters, when it's really just one "character".

My change effectively makes it so that both these instances are considered as one character, effectively including more "readable" characters in the limit.

As for your second point, it seems to be passing the tests, but you're right - I will adjust that.

bennothommo: I've removed the line break change line, as the next line handles it anyway (preg `\s` matches newlines too)

src/Html/HtmlBuilder.php
```diff
@@ -428,14 +428,14 @@ public static function limit($html, $maxLength = 100, $end = '...')
             list($tag, $tagPosition) = $match[0];
 
             $str = substr($html, $position, $tagPosition - $position);
-            if ($printedLength + strlen($str) > $maxLength) {
+            if ($printedLength + self::getReadableLength($str) > $maxLength) {
```
LukeTowers:
```suggestion
            if ($printedLength + static::getReadableLength($str) > $maxLength) {
```

src/Html/HtmlBuilder.php
```diff
@@ -428,14 +428,14 @@ public static function limit($html, $maxLength = 100, $end = '...')
...
                 $result .= substr($str, 0, $maxLength - $printedLength) . $end;
                 $printedLength = $maxLength;
                 break;
             }
 
             $result .= $str;
-            $printedLength += strlen($str);
+            $printedLength += self::getReadableLength($str);
```
LukeTowers:
```suggestion
            $printedLength += static::getReadableLength($str);
```

LukeTowers (changes requested):
@bennothommo I always recommend using `static` over `self` in framework code because it's more supportive of being extended.

PR #512, opened 2020-08-11
Fix Windows path resolution in "resolve_path" helper.

bennothommo:
Linked to #510 - this came up in the Windows unit tests. This PR should be merged in with that PR before merge.

Fixes https://github.com/octobercms/october/issues/5236.


src/Support/helpers.php
```diff
@@ -918,26 +918,31 @@ function title_case($value)
...
+     * Returns canonical path if it can be resolved, otherwise `false`.
      *
      * @param  string  $path
      * @return string|bool
      */
     function resolve_path($path)
     {
         // Normalise directory separators
-        $path = str_replace('\\', '/', $path);
+        $path = str_replace(['\\', '/'], '/', $path);
```
LukeTowers: What's the logic behind this change?

src/Support/helpers.php
```diff
@@ -918,26 +918,31 @@ function title_case($value)
...
+        $drive = (preg_match('/^([A-Z]:)/', $path, $matches) === 1)
+            ? $matches[1]
+            : null;
 
         // Check for a relative or absolute path, and prepend the working directory if relative, then split up the
         // path into segments
```
LukeTowers:
```suggestion
        // Check if this path is absolute or relative then split it into segments
```

src/Support/helpers.php
```diff
@@ -918,26 +918,31 @@ function title_case($value)
...
-        if (substr($path, 0, 1) === '/') {
+        if (substr($path, 0, 1) === '/' || !is_null($drive)) {
             $pathSegments = explode('/', $path);
         } else {
-            $path = getcwd() . '/' . $path;
+            $path = str_replace(['\\', '/'], '/', getcwd()) . '/' . $path;
```
LukeTowers:
```suggestion
            // Prepend the working directory to relative paths
            $path = str_replace(['\\', '/'], '/', getcwd()) . '/' . $path;
```

src/Support/helpers.php
LukeTowers: We might need to move this into a helper class that gets called from this function instead so that we can take advantage of methods to reduce / clarify some of the duplicated cleaning functionality

LukeTowers (changes requested):

src/Support/helpers.php
bennothommo: @LukeTowers to make it easier to track and parse, I opted to change all directory separators while the path is being resolved to Unix `/` separators, regardless of OS. It's only when the final resolved path is generated and returned will the OS' directory separator be used.

bennothommo: @LukeTowers I've converted the helper method to a Helper class (`October\Rain\Filesystem\PathResolver`) which might make it easier to track what is happening.

bennothommo:
After discussing https://github.com/octobercms/october/issues/5236 with @Samuell1, we were able to determine that his fault also lies with the `starts_with` method call here: https://github.com/octobercms/october/blob/develop/modules/cms/classes/CmsObject.php#L235.

The path created in the lines before are using forward slashes as though they are a Unix path, but this won't be the case for Windows, so the `starts_with` method in conjunction with this `resolve_path` helper is failing on Windows. I have added a `within` chainable method to this new helper class which can normalise the given directory and determine if a path is within that directory.

src/Filesystem/PathResolver.php
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * A utility to resolve paths to their canonical location and handle path queries.
+ *
+ * @package october\filesystem
+ * @author Ben Thomson
+ */
+class PathResolver
+{
+    /** @var string The path to resolve or compare. */
+    protected $path;
```
LukeTowers: I don't think this is worth being a class that needs to be instantiated, each method could very easily just be a static method, no need for state here.

LukeTowers (changes requested):

src/Filesystem/PathResolver.php
bennothommo: Yeah good point, I've simplified the class a bit.

PR #513, opened 2020-08-12
Fix method doc return type

rajakhoury:


src/Parse/Bracket.php
```diff
@@ -35,7 +35,7 @@ public function setOptions($options = [])
      * @param  string $template
      * @param  array $vars
      * @param  array $options
-     * @return self
+     * @return bool|string|string[]|null
```
LukeTowers: This would only return bool or string from what I can tell.

rajakhoury: Was auto generated by the editor. But i guess you are right. 
The `string` or `array`  is referenced from [str_replace](https://www.php.net/manual/en/function.str-replace.php)  and `null` from [preg_replace ](https://www.php.net/manual/en/function.preg-replace.php) - See `parseKeyBooleans` - we can cast the returned value to string or just update the comment to bool/string/array instead of self.

LukeTowers: Array is not going to be returned because str_replace is not being provided an array as input. Null is unlikely to be returned as it only gets returned by preg_replace if an error occurs, which is unlikely. If you want to include it, then what should actually happen is that the code should properly handle that return value so it doesn't get passed up the call stack and we should add a unit test specifically for whatever input would have to be provided to return that value.

rajakhoury: Seems unnecessary. Just replacing self with boolean and string.

rajakhoury (commented):
Test errors

ini-test

src/Parse/Ini.php
```diff
@@ -121,7 +121,7 @@ public function expandProperty(&$array, $key, $value)
             return $array = $value;
         }
 
-        $keys = explode('|', $key);
+        $keys = explode('|', (string) $key);
```
LukeTowers: Why was this change required?

rajakhoury: I'm sorry I wasn't clear enough. I don't think changes are required as I mentioned earlier the test failed on my end because I had strict type declaration (rain fork). 
In this example the &key passed from the test fixtures is not always a string and test failed. Maybe this could be a future consideration.

LukeTowers: I would rather deal with why it isn't a string sometimes instead of explicitly converting and potentially covering up issues.

LukeTowers:
@rajakhoury FYI, this is why PRs should be submitted from a feature branch rather than made directly on your fork's develop branch.

rajakhoury:
Yes, i messed up.

LukeTowers:
No worries, just something to know for the future.

PR #514, opened 2020-08-13
Allow morphTo relations to specify custom keys

jjfmarket:
Remove a check that asserts that a morphTo relationship should not have any additional arguments. if the morphMany relationship defines custom id or type keys, then the morphTo relationship must define them as well, or the relationship will not be found.

In my case, the database already has some unwieldly column names that i'm using and I don't want to use the magic column names in HasRelationships->getMorphs

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -269,14 +269,6 @@ trait HasRelationships
             ));
         }
 
-        if (isset($relation[0]) && $relationType == 'morphTo') {
-            throw new InvalidArgumentException(sprintf(
-                "Relation '%s' on model '%s' is a morphTo relation and should not contain additional arguments.",
-                $relationName,
-                get_called_class()
-            ));
-        }
-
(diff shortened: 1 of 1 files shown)


bennothommo:
@jjfmarket While this does remove the check, I don't believe it's going to work for you. The key, as far as I can tell, is not passed to the `morphTo` method which means that it won't be using your custom key (which I believe is the `$ownerKey` property).

Have you tested this and confirmed it working with a `morphTo` relation with a custom key?

jjfmarket:
Yea, tested and confirmed. You can see in the 'morphTo' case in the switch statement passes 'id' and 'type' which are the column names for id and type of the related model

https://github.com/octobercms/library/blob/cd143d9973538084a66307e0c3dfd82a8fda1922/src/Database/Concerns/HasRelationships.php#L299

jjfmarket:
Hey actually, you're right. The issue was that I had specified the classname. Removing that solved my problem. Thanks!

PR #515, opened 2020-08-19
check if imagecreatefromwebp function exists

mabehiry:
Before, if imagecreatefromwebp function doesn't exist, it will break.
Now, it will not process the image.

### DIFF
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -521,8 +521,14 @@ class Resizer
                 $this->retainImageTransparency($img);
                 break;
             case 'image/webp':
-                $img = @imagecreatefromwebp($filePath);
-                $this->retainImageTransparency($img);
+                if (function_exists('imagecreatefromwebp'))
+				{
+					$img = @imagecreatefromwebp($filePath);
+					$this->retainImageTransparency($img);
+				}
+				else {
+					$img = '';
+				}
                 break;
             default:
                 throw new Exception(sprintf('Invalid mime type: %s. Accepted types: image/jpeg, image/gif, image/png, image/webp.', $this->mime));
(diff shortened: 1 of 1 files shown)


src/Database/Attach/Resizer.php
```diff
@@ -521,8 +521,14 @@ protected function openImage($file)
                 $this->retainImageTransparency($img);
                 break;
             case 'image/webp':
-                $img = @imagecreatefromwebp($filePath);
-                $this->retainImageTransparency($img);
+			    if (function_exists('imagecreatefromwebp'))
```
LukeTowers, line 524: Spaces not tabs please.

LukeTowers:
Under what conditions does that function not exist? And what will happen with your code now if that function doesn't exist?

mabehiry:
Sometimes, it is not included in the installed GD library, so Webp images fail to be resized. So I get this error:
Call to undefined function imagecreatefromwebp()
After fixing, if the function is not found, it will not fail, it will just not resize the image.

LukeTowers:
@mabehiry that's not acceptable. If the function isn't present then it should throw an exception telling the developer that the function isn't present and ideally how they can fix the issue.

bennothommo:
I agree with @LukeTowers - this shouldn't be failing silently. If you want to use WEBP images but haven't configured the GD plugin correctly or installed the necessary libraries, you (or the client) should be informed of this.

mabehiry:
You're right, but for me for example, I failed to install the library. Although GD library is installed, I couldn't configure it to for webp images. The same is for anyone who is using a shared hosting. So, the entire app fails when user uploads an image which is in jpg format, but it -for some reason- requires this function.
I think that it is better to give a warning instead of failing.

LukeTowers:
@mabehiry the warning is the exception. If you're saying this code path is triggered by someone uploading a JPEG file, that seems suspicious to me, since this should only be running on webp files. Is the file actually a webp image using a .jpg extension?

mabehiry:
@LukeTowers Yes, I think it is webp images with jpg extension.

LukeTowers:
@mabehiry then that's an invalid file to be uploading in the first place so an exception should definitely be thrown.

PR #517, opened 2020-08-25
Prevent block termination from ending parent buffer

### LINKED ISSUE #5108: Additional Block::placeholder() causing backend views to render incorrectly
- OctoberCMS Build: ### 
- PHP Version: 7.4
- Plugins Installed: October\Test

### Description:
When trying to add additional `Block` placeholders to the `default.htm` layout, the page is rendered incorrectly. 

### Steps To Reproduce:
1. Install the Test plugin if not done so already
2. Update `backend/layouts/default.htm` to add the following after the breadcrumb block placeholder on lines 45-50:
```
<?php if ($testBlock = Block::placeholder('test-block')): ?>
<!-- Test block -->
<div class="control-breadcrumb">
<?= $testBlock ?>
</div>
<?php endif ?>
```
3. Add the following to `plugins/october/test/controllers/people/create.htm`:
```
<?php Block::put('test-block') ?>
    <ul>
        <li>Test</li>
        <li>Block</li>
    </ul>
<?php Block::endPut() ?>
```
4. Notice that the form and/or body is rendered outside of the page, and that the page source shows the <head> element is rendered in the middle of the page:
image
image

As an example, the expected output would be something like this:
image
image

bennothommo:
Fixes https://github.com/octobercms/october/issues/5108

These 3 lines in the BlockBuilder are basically breaking out of a parent output buffer in the scenario mentioned in the linked issue:
https://github.com/octobercms/library/blob/develop/src/Html/BlockBuilder.php#L47-L49

Essentially, if there's any output between the two blocks (such as some normal content, or even just whitespace), those lines will do another `ob_end_clean`, which actually breaks out of the parent buffer (in your case, the `body` buffer, which is delivered here: https://github.com/octobercms/october/blob/develop/modules/system/traits/ViewMaker.php#L126-L127). This is what causes the content to end up at the top of the page if you use multiple blocks.

This should allow multiple blocks to be used. If any content is within the blocks, it should be picked up and appended to the body via the `ViewMaker` code linked above.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
--- a/src/Html/BlockBuilder.php
+++ b/src/Html/BlockBuilder.php
@@ -43,10 +43,6 @@ class BlockBuilder
     public function endPut($append = false)
     {
         $this->endBlock($append);
-
-        if (!count($this->blockStack) && (ob_get_length() > 0)) {
-            ob_end_clean();
-        }
     }
 
     /**
@@ -66,8 +62,7 @@ class BlockBuilder
 
         if ($append) {
             $this->append($name, $contents);
-        }
-        else {
+        } else {
             $this->blocks[$name] = $contents;
         }
     }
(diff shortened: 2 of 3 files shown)


LukeTowers:
@bennothommo are we able to add some unit testing based on the current functionality? Not sure if this could cause unintended differences in behaviour.

bennothommo:
@LukeTowers no problem. It is a change of behaviour - content in between blocks is now kept instead of discarded, but I don't see an issue with that.

bennothommo:
Tests now added.

bennothommo:
@LukeTowers I'm happy to merge this if you have no objections. I believe it's adequately tested.

LukeTowers:
Seems fine to me @bennothommo, just make sure the change is clearly documented in meta

PR #518, opened 2020-08-28
Allow to install latest jenssegers/date

kylekatarnls:
jenssegers/date now relies on Carbon 2 for the translatons which support more languages.

All the features of jenssegers are now provided out of the box by Carbon, so the new jenssegers/date is way lighter (https://github.com/jenssegers/date/blob/master/src/Date.php) and directly use internal methods of Carbon to just add the translation support right in `parse()` and `createFromFormat()` method.

De facto, it fixed a lot of remaining bugs of jenssegers/date 3: https://github.com/jenssegers/date/pull/317

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "symfony/yaml": "^3.4",
         "twig/twig": "~2.0",
         "league/csv": "~9.1",
-        "jenssegers/date": "~3.5",
+        "jenssegers/date": "~3.5 || ~4.0",
         "laravel/framework": "~6.0",
         "laravel/tinker": "~2.0"
     },


bennothommo:
@kylekatarnls Thanks for submitting this - everything looks good, we just need to test this against October CMS itself.

LukeTowers:
@bennothommo when you test can you test just removing this dependency altogether and putting the code in the Argon class instead?

kylekatarnls:
I see the `Argon` class is empty. If you're about removing extra-stuff, you should consider removing this one first IMHO.

LukeTowers:
@kylekatarnls the Argon class is empty for this exact situation. Because it existed as a wrapper over Jenssegers and we referenced Argon instead of Jensegers elsewhere in our code we're able to completely remove Jenssegers and not have to change anything else because everything is referencing Argon instead.

bennothommo:
@LukeTowers I'd probably keep the Jenssegers library as it looks like it's somewhat maintained, although @kylekatarnls might be able to shed some light on whether the library is going to be maintained or if @jenssegers [comment here](https://github.com/jenssegers/date/pull/317#issuecomment-596989284) is indicating that the library will be dropped now that Carbon is covering its use case a little better.

kylekatarnls:
The adding of JenssegersDate were very relevant as it was in advance for the I18N support for years. But I bootstrapped a fresh October app and I didn't find any relevant usages of the Argon/JenssegersDate. It sounds like October only use `diffForHumans()` and Carbon supports I18N in this method since Carbon 1.18 (March 2015), and the place I can see where `Argon` is called are actually duplicated code of stuffs in Laravel adapted to get `Argon` instances instead of `Carbon` instance.

Unless I missed something, you could drop Argon/JenssegersDate since the 1.18 release of Carbon, it would not make any difference.

The main feature of JenssegersDate is to convert date strings in non-English languages to date instances. Is there any module or third-party that could use it and still require?

Note than since Laravel 7, you also can use `Date::use(Argon::class)` instead of overriding `asDateTime`, `freshTimestamp` and such methods.

LukeTowers:
@kylekatarnls we use it to handle localized dates, can't recall where specifically right now though. October supports a wide variety of backend translations out of the box though. As far as using `Date::use()` goes, October is pinned to LTS releases of Laravel, so Laravel 7 is not usable for us right now.

LukeTowers:
This has been replaced by https://github.com/octobercms/library/commit/7c6ef31a1fb6b9aab507527e9af950012a2dec59. Thanks for your initial PR @kylekatarnls!

PR #520, opened 2020-09-07
Ini parser with strict types

rajakhoury:
related #513

### DIFF
--- a/src/Parse/Ini.php
+++ b/src/Parse/Ini.php
@@ -202,7 +202,7 @@ class Ini
                     }
                 }
             }
-            elseif (strlen($value)) {
+            elseif ($value . '' !== '') {
                 $content .= $key.' = '.$this->evalValue($value).PHP_EOL;
             }
         }
--- a/src/Parse/Ini.php
+++ b/src/Parse/Ini.php
@@ -121,7 +121,7 @@ class Ini
             return $array = $value;
         }
 
-        $keys = explode('|', $key);
+        $keys = explode('|', (string) $key);
 
         while (count($keys) > 1) {
             $key = array_shift($keys);


src/Parse/Ini.php
```diff
@@ -121,7 +121,7 @@ public function expandProperty(&$array, $key, $value)
             return $array = $value;
         }
 
-        $keys = explode('|', $key);
+        $keys = explode('|', (string) $key);
```
LukeTowers, line 124: When would this not be a string?

src/Parse/Ini.php
```diff
@@ -202,7 +202,7 @@ protected function renderProperties($vars = [])
                     }
                 }
             }
-            elseif (strlen($value)) {
+            elseif ($value . '' !== '') {
```
LukeTowers, line 205: Why was this change made?

rajakhoury: This fails the `testBasic` see [here](https://github.com/octobercms/library/blob/develop/tests/Parse/IniTest.php#L19) - The [render ](https://github.com/octobercms/library/blob/develop/tests/Parse/IniTest.php#L40) method is passed values that aren't string. 
Again this error only happens if you have `declare(strict_types=1);` declared in your `October\Rain\Parse\Ini.php` file which I understand is not a coding standard for this library but thought this may be useful as a side note.

As for the work around maybe add `(strlen((string) $value))` or whatever you think is appropriate.

Which also takes to question the `evalValue` method - It is supposed to return a string and why?

        // Numeric - This doesn't return a string as the value is returned as it is.
        if (is_numeric($value)) {
            return $value;   
        }

test screenshot

src/Parse/Ini.php
rajakhoury: Not sure what's going on here but it fails the `testArray` - the array keys?
again see comment below this is just a related to strict types.

screenshot

rajakhoury:
I am using this library in private projects not related to octobercms - but I think these parsers aren't tied to OC that's why I open this PR. 

Let me know if I should close it or this needs further investigation.

LukeTowers:
We don't use strict types in the October core and we're not going to merge random type casting changes without a good reason and understanding of why they're required. Thanks for your interest though!

PR #523, opened 2020-09-23
save/retrieve locale when serializing mail data in the queue

mjauvin:
When a mail is queued, we need to save the locale that was used to generate the mail in order to be able to localize the mail view when the mail is dequeued.

Upon dequeueing the mail, we retrieve the locale and set the App locale with it.

### DIFF
--- a/src/Mail/Mailable.php
+++ b/src/Mail/Mailable.php
@@ -48,7 +48,7 @@ class Mailable extends MailableBase
      */
     public function withSerializedData($data)
     {
-        // save current locale to be serialized
+        // Ensure that the current locale is stored with the rest of the data for proper translation of queued messages
         $defaultData = [
             '_current_locale' => App::getLocale(),
         ];
--- a/src/Mail/Mailable.php
+++ b/src/Mail/Mailable.php
@@ -55,8 +55,7 @@ class Mailable extends MailableBase
     public function withSerializedData($data)
     {
         // save current locale in $data to be serialized
-        list($locale) = explode('-', App::getLocale());
-        $data['_saved_locale'] = $locale;
+        $data['_saved_locale'] = App::getLocale());
 
         foreach ($data as $param => $value) {
             $this->viewData[$param] = $this->getSerializedPropertyValue($value);
--- a/src/Mail/Mailable.php
+++ b/src/Mail/Mailable.php
@@ -33,12 +33,6 @@ class Mailable extends MailableBase
     {
         $data = $this->viewData;
 
-        // retrieve saved locale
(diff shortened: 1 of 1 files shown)


src/Mail/Mailable.php
```diff
@@ -47,6 +54,9 @@ public function buildViewData()
      */
     public function withSerializedData($data)
     {
+        // save current locale in $data to be serialized
+        $data['_saved_locale'] = App::getLocale());
```
LukeTowers: Key should probably be `_current_locale` instead, and I'd recommend something like the following:
```php
$defaultData = [
    '_current_locale' => App::getLocale(),
];

$data = array_merge($defaultData, $data);
```
to allow developers to override that value if desired.

src/Mail/Mailable.php
```diff
@@ -32,6 +33,12 @@ public function buildViewData()
     {
         $data = $this->viewData;
 
+        // retrieved saved locale
+        if (isset($data['_saved_locale'])) {
+            App::setLocale($data['_saved_locale']);
+            unset($data['_saved_locale']);
+        }
```
LukeTowers: I'd recommend also storing the current locale before making the change and restoring it after the mail has been processed just in case.

src/Mail/Mailable.php
mjauvin: good idea.

src/Mail/Mailable.php
mjauvin: in case of what? can you give more details of why/where you'd want this?

LukeTowers: As a rough example, lets say that someone was logging messages after the email was sent, if the locale was switched but not changed back then the messages could start appearing in the log in a different locale. Additionally, if multiple mails are processed in a single request (not sure if this applies to queued emails, but this method might be called on regular emails too), a single email could then change the entire application locale for the rest of the request without it being properly reset back to the original locale immediately after sending the email.

mjauvin: ok, I'll leave the _current_locale variable in there, it can be used by whoever needs it and shouldn't interfere with other requests.

mjauvin:
What do we need to move this forward, it's blocking https://github.com/rainlab/translate-plugin/pull/605

src/Mail/Mailable.php
```diff
@@ -47,6 +48,13 @@ public function buildViewData()
      */
     public function withSerializedData($data)
     {
+        // save current locale to be serialized
```
LukeTowers:
```suggestion
        // Ensure that the current locale is stored with the rest of the data for proper translation of queued messages
```

PR #524, opened 2020-09-23
add methods to peek the protected $casts array

mjauvin:
There is currently no way to know if a model attribute is a cast or which cast type it's using.

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -469,9 +469,31 @@ class Model extends EloquentModel
     }
 
     /**
-     * Checks if an attribute is jsonable or not.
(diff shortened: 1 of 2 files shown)


mjauvin:
I don't understand how those tests passed before... unless they weren't ?

Samuell1:
@mjauvin Because that methods exists in laravel https://github.com/laravel/framework/commit/f59f80385f50d284d5276b531b9ec61131c41b21

bennothommo:
@Samuell1 is correct - these methods (or at least `getCasts`) already exist for October database models.

PR #525, opened 2020-09-25
Added custom fallback image feature

### LINKED ISSUE #5150: Allow custom fallback image for missing images
The rain library currently uses [a base64 string](https://github.com/octobercms/library/blob/30e610d871622df754f028bc566a412a3727f14a/src/Database/Attach/BrokenImage.php#L26) that translates to this image. It would be good to allow upload of custom image/base64 encoded string like say a company logo so that when BrokenImage appears on my website it shows the company logo instead. Is there any other way to do it? I thought maybe it would be a good idea to make a plugin for this but I cannot see how I could manipulate the october library code from plugin to allow such custom image.

mohsin:
This completes PR octobercms/october#5284 and closes octobercms/october#5150


bennothommo:
Closing as per the comment here: https://github.com/octobercms/october/pull/5284#issuecomment-699762096

PR #529, opened 2020-10-16
Add most of test classes namespace

peter279k:
According to the `composer.json` file, it contains the `October\\Rain\\Tests\\` namespace for all classes.

It should be good to add above namespace for test classes.

And some classes don't add this namespace because they contain two or more example classes on the same PHP files or include some classes manually on `fixtures` folder during the `PHPUnit::setUp` method calling.


peter279k:
Closed because I use the wrong branch.

PR #530, opened 2020-10-16
Add most of test classes namespace

peter279k:
# Changed log

According to the `composer.json` file, it contains the `October\\Rain\\Tests\\` namespace for all classes.

It should be good to add above namespace for test classes.

And some classes don't add this namespace because they contain two or more example classes on the same PHP files or include some classes manually on `fixtures` folder during the `PHPUnit::setUp` method calling.

### DIFF
--- a/tests/Assetic/MockAsset.php
+++ b/tests/Assetic/MockAsset.php
@@ -1,5 +1,7 @@
 <?php
 
+namespace October\Rain\Tests\Assetic;
+
 use October\Rain\Assetic\Asset\AssetInterface;
 use October\Rain\Assetic\Filter\FilterInterface;
 
--- a/tests/Assetic/StylesheetMinifyTest.php
+++ b/tests/Assetic/StylesheetMinifyTest.php
@@ -1,6 +1,9 @@
 <?php
 
+namespace October\Rain\Tests\Assetic;
+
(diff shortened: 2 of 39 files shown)


LukeTowers:
What's the benefit of doing this? Also, october's convention is to have the namespace statement on the same line as the opening php tag, so `<?php namepsace X\Y\Z;`

peter279k:
The benefit is about loading testing classes automatically and it can refer this [setting](https://getcomposer.org/doc/04-schema.md#autoload-dev) explanation.

And the namespace will change into `<?php namepsace X\Y\Z;`.

LukeTowers:
@bennothommo is there any benefit to this PR?

bennothommo:
@peter279k Thanks for your effort on this, however, we have previously added namespaces to the tests and then subsequently removed them due to the extra complexities and backwards incompatibilities that introducing namespaces creates (https://github.com/octobercms/library/pull/456/commits/547fe9b7efb4a9f2647cc18ddd128ffa97990342). Introducing namespaces effectively forces all plugins with unit tests to have to change their test cases to use the namespaced class as well.

We already autoload all necessary tests through https://github.com/octobercms/october/blob/develop/phpunit.xml - while this is in a different repo, this is set to run the library unit tests as well as the CMS unit tests, and it's generally better to test them in combination to catch any issues.

PR #536, opened 2020-10-26
Apply conditions/scopes to related model query

### LINKED ISSUE #4952: Relation conditions aren't being applied when saving models
- OctoberCMS Build: 464
- PHP Version: 7.4.2

### Description:
If you have two belongsToMany relations on the same table with a `conditions` 
value set, record retrieval (`$record->myRelation`) works correctly and applies 
the specified conditions, however saving `$record` wiil result in the first 
relation being wiped and only the second relation will save.

`newPivotQuery()` isn't applying constraints as it should.

### Steps To Reproduce:
* `git clone https://github.com/Flynsarmy/oc-relationtest-plugin.git` to */plugins/flynsarmy/relationtest*
* `php artisan plugin:refresh Flynsarmy.RelationTest`
* Go to */backend/flynsarmy/relationtest/posts/update/1*
* Notice we have two categories in our *category* relation and two tags in our *tags* relation. Both relations connect to the same table but use a `conditions` option in our model.
* Hit Save
* Refresh
* *Categories* has been wiped but *Tags* saved.

### Issue in Depth

In our `Post` model we have two relations defined:
```php
    public $belongsToMany = [
        'tags' => [
            'Flynsarmy\RelationTest\Models\Term',
            'table' => 'flynsarmy_relationtest_posts_terms',
            'key'        => 'post_id',
            'otherKey'   => 'term_id',
            'conditions' => 'type = "tag"'
        ],
        'categories' => [
            'Flynsarmy\RelationTest\Models\Term',
            'table' => 'flynsarmy_relationtest_posts_terms',
            'key'        => 'post_id',
            'otherKey'   => 'term_id',
            'conditions' => 'type = "category"'
        ],
    ];
```
These link to the same table but retrieval and display on our form works as 
expected thanks to our `conditions` statement. This can be further be confirmed 
with
```bash
php artisan tinker
Flynsarmy\RelationTest\Models\Post::first()->tags
Flynsarmy\RelationTest\Models\Post::first()->categories
```

When saving the model, `October\Rain\Database\Relations\BelongsToMany::setSimpleValue()` 
calls `$this->sync($value);`.
The `sync` method defined in */vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php*
has the lines
```php
    $current = $this->newPivotQuery()->pluck(
        $this->relatedPivotKey
    )->all();
```
to determine which records need to be saved for the current relation. The problem is
`newPivotQuery()` doesn't apply our `conditions`, so the query 
```sql
select * from `flynsarmy_relationtest_posts_terms` where `post_id` = ?
```
is executed and ALL records in this table are returned instead of just the ones
relevent to the current relation.
The next lines in this class:
```php
$detach = array_diff($current, array_keys(
    $records = $this->formatRecordsList($this->parseIds($ids))
));
```
detect which records weren't selected on our form and lists them for removal.

Because our form has the *Categories* field first, it marks the tag records for removal.
Removal occurs, and we move on to the *Tags* field. The same thing happens and our
categories are marked for removal.

The end result is only the tags are saved and all categories are removed.

mjauvin:
Fixes: https://github.com/octobercms/october/issues/4952

This overrides the following method in Laravel:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L83

Specifically replaces the following lines to fetch the records that *actually* belongs to the relation (using the conditions/scopes of the relation):

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L92-L93


mjauvin:
@Flynsarmy can you confirm this fixes your problem?

Flynsarmy:
Can confirm, this fixes my problem. Thanks!

Flynsarmy:
This PR helps but there still appears to be a bug which you can replicate with my test plugin in octobercms/october#4952. Uncheck both Tag checkboxes but leave one or more Category checkboxes checked. Hit save and refresh. You'll notice the categories did not save correctly. A look in the DB confirms no categories were saved to the `flynsarmy_relationtest_posts_terms` table.

mjauvin:
@Flynsarmy try this, I think I fixed the remaining bug.

Flynsarmy:
@mjauvin Looks good! Fixed.

src/Database/Relations/BelongsToMany.php
```diff
@@ -409,4 +409,48 @@ public function getRelatedIds($sessionKey = null)
         traceLog('Method BelongsToMany::allRelatedIds has been deprecated, use BelongsToMany::allRelatedIds instead.');
```
LukeTowers:
```suggestion
        traceLog('Method BelongsToMany::getRelatedIds has been deprecated, use BelongsToMany::allRelatedIds instead.');
```

mjauvin:
I and @Flynsarmy have tested this as a working fix.

src/Database/Relations/BelongsToMany.php
```diff
@@ -188,7 +188,7 @@ public function detach($ids = null, $touch = true)
         /*
          * See Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithPivotTable
```
LukeTowers, line 189: Can you update all the relation before* events to not be halting events and instead require exceptions to be thrown to halt them?

LukeTowers: might just be the one beforeAttach one above

LukeTowers: This comment is meant for #537 btw

src/Database/Relations/BelongsToMany.php
```diff
@@ -406,7 +406,51 @@ public function getOtherKey()
...
+     * @return array
+     */
+    public function sync($ids, $detaching = true)
+    {
+        $changes = [
+            'attached' => [], 'detached' => [], 'updated' => [],
+        ];
+
+        // Get all the IDs for the related model
+        $current = $this->allRelatedIds()->all();
```
LukeTowers: Wouldn't a better fix be to adjust the getCurrentlyAttachedPivots() method to return the correct query builder instead? That way we don't have to override this method at all?

mjauvin: You may be right, although I am not sure of all the implications that would have on the other classes using the `InteractsWithPivotTable` trait.

LukeTowers: Presumably it would fix similar issues with those method consumers at the same time. @bennothommo any thoughts? (btw, Ben's got spotty internet right now, might take a couple days)

LukeTowers:
A unit test for the original issue would be helpful here @mjauvin

src/Database/Relations/BelongsToMany.php
mjauvin: It's definitely worth investigating if I have some help from Ben or yourself...

mjauvin: The new commit addresses the issue with the same result, but overrides getCurrentlyAttachedPivots() method instead of the sync() method.

Note that it currently only affects `BelongToMany` relation but not `MorphMany`.

mjauvin:
@LukeTowers there are tests in october/october for testing relations, would that make sense to improve on those tests instead of rewriting these in october/library?

LukeTowers:
Might be, any thoughts @bennothommo?

bennothommo:
@mjauvin @LukeTowers the test should really exist in here if the changed functionality is here - while it's likely that 99.9% of the time, this will be used in October CMS, it's still proper testing procedure for this sort of functionality to be tested in isolation.

I would simply port a couple of those relation tests back into this library and adjust as necessary.

LukeTowers:
@bennothommo I think we might need your skills to setup database tests for this library then, I was having a heck of time trying to make it work for the guarded tests

mjauvin:
@LukeTowers I found this:
https://github.com/octobercms/library/blob/1.1/tests/Database/Traits/EncryptableTest.php
```
    public function setUp(): void
    {
        $capsule = new Illuminate\Database\Capsule\Manager;
        $capsule->addConnection([
            'driver'   => 'sqlite',
            'database' => ':memory:',
            'prefix'   => ''
        ]);

        $capsule->setAsGlobal();
        $capsule->bootEloquent();

        $capsule->schema()->create('secrets', function ($table) {
            $table->increments('id');
            $table->string('secret');
            $table->timestamps();
        });
}
```
That should help

bennothommo:
@LukeTowers what @mjauvin posted is the only way of testing database functions that I'm aware of with the library. Give that a try and let me know if it doesn't work - I'll have to port some of the functionality from October CMS back into the library.

I swear I did this originally with the L6 development, but had it reversed afterwards - c'est la vie :(.

tests/Database/RelationsTest.php
```diff
@@ -0,0 +1,161 @@
+<?php
+
+class RelationsTest extends DbTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->create_tables();
+        $this->seed_tables();
```
bennothommo: @mjauvin Convention is to use camelcase for method names
```suggestion
        $this->createTables();
        $this->seedTables();
```

mjauvin: yeah, don't know where that came from... 8)

mjauvin:
Should be good now!

tests/DbTestCase.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+use October\Rain\Database\Model as ActiveRecord;
```
LukeTowers: Why name this ActiveRecord? Why not just use it as it is?

mjauvin: don't know, I used what was in PluginTestCase...

LukeTowers: I would just remove it, it's not used like that anywhere in the actual code base.

mjauvin: done!

tests/Database/RelationsTest.php
```diff
@@ -0,0 +1,161 @@
...
+    public function seedTables()
+    {
+        $post = Post::create([
+            'title' => 'A Post',
+        ]);
+
+        $post->tags()->create(['type'=>'tag', 'name'=>'A Tag']);
+        $post->tags()->create(['type'=>'tag', 'name'=>'Second Tag']);
+        $post->categories()->create(['type'=>'category', 'name'=>'A Category']);
+        $post->categories()->create(['type'=>'category', 'name'=>'Second Category']);
```
bennothommo:
```suggestion
        $post->tags()->create(['type' => 'tag', 'name' => 'A Tag']);
        $post->tags()->create(['type' => 'tag', 'name' => 'Second Tag']);
        $post->categories()->create(['type' => 'category', 'name' => 'A Category']);
        $post->categories()->create(['type' => 'category', 'name' => 'Second Category']);
```

bennothommo (changes requested):

bennothommo:
@mjauvin Just to eliminate any doubt, would you mind adding a couple more scenarios in the tests, namely:

- Syncing multiple records
- Detaching one record when multiple are attached

If those are covered, I'm happy to merge.

bennothommo:
Also, I'm going to merge your DbTest changes manually into `develop`, as I need it for Luke's PR. When you do the above, you'll need to merge `develop` back into your branch - hopefully, it won't conflict but if it does, just accept any changes from `develop`.

mjauvin:
The tests have been added @bennothommo

tests/Database/RelationsTest.php
```diff
@@ -0,0 +1,184 @@
...
+
+        $post->categories()->sync($category_ids);
+        $this->assertEquals(4, $post->categories()->count());
+    }
+
+    public function testBelongsToManyDetachOneCategory()
+    {
+        $post = Post::first();
+
+        $post->categories()->detach([6]);
```
bennothommo: @mjauvin I would recommend not using a static ID for tests - it makes them too susceptible to failure should any further changes to this particular test suite be done. Also, do you know if the tables are reset after each test? An ID of `6` seems to be a bit weird if this were a fresh DB.

mjauvin: Hey @bennothommo, you're right but until I find how to flush the cache between queries, it's hard to do otherwise.

The number 6 is because I add 6 records to the Term model in the seedTables() method.

Thanks!

mjauvin: I used this in the mean time, at least it's not using a static id.

mjauvin: @LukeTowers do you know how to flush the query cache after a db query?

LukeTowers: @mjauvin https://octobercms.com/docs/database/query#in-memory-caching

mjauvin: Thanks @LukeTowers, exactly what I was looking for!

mjauvin:
Anything else I need to do to here?

PR #537, opened 2020-10-27
Add more relation events

### LINKED ISSUE #4561: Add other model's relation events
By now there are only attach/detach events for [BelongsToMany relation](https://octobercms.com/docs/api/model/relation/afterattach). Adding other events will make using OctoberCMS more comfortable. Good example there: https://github.com/chelout/laravel-relationship-events. 
Please add this feature request [to bounty](https://www.bountysource.com/teams/october/issues).

mjauvin:
Fixes: https://github.com/octobercms/october/issues/4561


src/Database/Relations/MorphTo.php
```diff
@@ -70,4 +70,85 @@ public function getSimpleValue()
...
+         *
+         * Example usage:
+         *
+         *     $model->bindEvent('model.relation.beforeAssociate', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+         *         TODO: add example code
+         *     });
+         *
+         */
+        if ($this->parent->fireEvent('model.relation.beforeAssociate', [$this->relationName, $this->related], true) === false) {
+            return;
```
LukeTowers: I don't think these events should be halting, firstly that breaks the contract of what this method says it returns, and secondly if someone really absolutely needs to prevent a relationship from happening at this layer of the application then they can throw an exception instead.

src/Database/Relations/BelongsTo.php
```diff
@@ -28,25 +28,77 @@ public function __construct(Builder $query, Model $child, $foreignKey, $ownerKey
...
+         *     });
+         *
+         */
+        $this->parent->fireEvent('model.relation.beforeAssociate', [$this->relationName, $this->related]);
+
         if ($sessionKey === null) {
             $this->associate($model);
         }
         else {
             $this->child->bindDeferred($this->relationName, $model, $sessionKey);
```
LukeTowers: This brings to mind a question, should the events be fired if the relationship has been deferred or should they wait until the deferred relationship has been completed? @bennothommo any thoughts?

src/Database/Relations/BelongsToMany.php
```diff
@@ -181,9 +179,7 @@ public function detach($ids = null, $touch = true)
          *     });
          *
          */
-        if ($this->parent->fireEvent('model.relation.beforeDetach', [$this->relationName, $attachedIdList], true) === false) {
-            return;
-        }
+        $this->parent->fireEvent('model.relation.beforeDetach', [$this->relationName, $attachedIdList]);
```
LukeTowers, line 192: This method and the attach method also need to return the result of the parent method like the docblock says they do.

src/Database/Relations/BelongsTo.php
```diff
@@ -28,25 +28,77 @@ public function __construct(Builder $query, Model $child, $foreignKey, $ownerKey
      */
     public function add(Model $model, $sessionKey = null)
```
Rike-cz: I think before/after Associate/Dissociate events binding would be better to place within dissociate()/associate() methods, not only to add()/remove().

mjauvin: @Rike-cz fire only in associate/dissociate methods ?

mjauvin: Never mind, I see the add/remove methods call the assosiate/dissociate anyway... so I'll override associate/dissociate and move the event firing there.

mjauvin: There you go.

Rike-cz:
One thing to attach/detach event. I thought that it [was written previously](https://octobercms.com/docs/api/model/relation/beforeattach) like you can halt whole procedure through false value returned on before*** event. It's not possible now?

And do you plan to implement HasOneOrMany, AttachOneOrMany, MorphOneOrMany?

Rike-cz (approved):

mjauvin:
> One thing to attach/detach event. I thought that it [was written previously](https://octobercms.com/docs/api/model/relation/beforeattach) like you can halt whole procedure through false value returned on before*** event. It's not possible now?

@LukeTowers requested [this to be changed](https://github.com/octobercms/library/pull/536#discussion_r514578576). The way to halt these is to throw an exception now. 

> And do you plan to implement HasOneOrMany, AttachOneOrMany, MorphOneOrMany?

For which events? I thought you said you were now thinking this was not required, did I read that wrong?

ref.
> And here is the question. To have inversional events for these relationships or to call inversional counterparts above when needed? When needed means add/remove methods where relationship arises.
If is it good idea to have them here too (I don't think so now)...

src/Database/Relations/BelongsTo.php
mjauvin: @LukeTowers I updated these to fire in the `associate/dissociate` methods instead, which resolves this question I think...

Rike-cz:
@mjauvin I was thinking, but last word has @LukeTowers. Can you tell us, what do you think about HasOneOrMany, AttachOneOrMany, MorphOneOrMany [events](https://github.com/octobercms/october/issues/4561#issuecomment-717374806), @LukeTowers? Useless?

Edit: I thought about two approaches to HasOneOrMany etc. and definitelly yes, add/remove-events is good and more clean idea.

src/Database/Relations/BelongsToMany.php
```diff
@@ -202,6 +200,8 @@ public function detach($ids = null, $touch = true)
          *
          */
         $this->parent->fireEvent('model.relation.afterDetach', [$this->relationName, $attachedIdList]);
```
LukeTowers: Should the after event params include the result of the method?

mjauvin: Yeah, was asking myself the same question afterwards... @Rike-cz?

Rike-cz: Agree.

src/Database/Relations/BelongsOrMorphTo.php
```diff
@@ -0,0 +1,87 @@
...
+         *
+         * Example usage:
+         *
+         *     $model->bindEvent('model.relation.afterAssociate', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+         *         $relatedClass = get_class($relatedModel);
+         *         traceLog("Relation {$relationName} was associated to model {$relatedClass}.");
+         *     });
+         *
+         */
+        $this->parent->fireEvent('model.relation.afterAssociate', [$this->relationName, $this->related]);
```
LukeTowers: All of the after events should include the result, you might have to have it as the third argument so that you can support the existing relation event.

mjauvin: @LukeTowers In this case, the parent returns $this->child which is the same model on which the event is fired, so I don't think this is very helpful, is it?

LukeTowers: I suppose not, $this->child is equal to $this->related?

mjauvin: @LukeTowers No, it's equal to `$this->parent` (for BelongsTo and MorphTo relations)

LukeTowers: Ah, so the return value of this method is the parent model that the event is bound to in the first place?

mjauvin: Yop.

src/Database/Relations/AttachOneOrMany.php
```diff
@@ -185,6 +200,20 @@ public function add(Model $model, $sessionKey = null)
...
+
+            /**
+             * @event model.relation.afterAdd
+             * Called after adding a relation to the model (only for AttachOne/AttachMany relations)
+             *
+             * Example usage:
+             *
+             *     $model->bindEvent('model.relation.afterAdd', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+             *         $relatedClass = get_class($relatedModel);
+             *         traceLog("Relation {$relationName} was added to model {$relatedClass}.");
```
LukeTowers: Isn't $this->related the FileModeel being attached to $this->parent? In which case it would be `$relatedClass was added as $relationName to get_class($model)`?

mjauvin: Indeed.

mjauvin: fixed

src/Database/Relations/BelongsToMany.php
```diff
@@ -121,14 +123,11 @@ public function attach($id, array $attributes = [], $touch = true)
...
          *         }
          *     });
          *
          */
-        if ($this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, $insertData], true) === false) {
-            return;
-        }
+        $this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, $insertData]);
 
         // Here we will insert the attachment records into the pivot table. Once we have
```
Rike-cz: **Question:** Detach method calls parent::detach(), great. But why attach method doesn't? Some [parent code](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L249-L251) is bypassed. I know that it wasn't changed by @mjauvin, so only asking...

src/Database/Relations/BelongsOrMorphTo.php
```diff
@@ -0,0 +1,89 @@
+<?php namespace October\Rain\Database\Relations;
+
+trait BelongsOrMorphTo
+{
+    /**
+     * Associate the model instance to the given parent.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
```
Rike-cz: It should be  `\Illuminate\Database\Eloquent\Model|int|string` for BelongsTo. `\Illuminate\Database\Eloquent\Model` is only for MorphTo.

src/Database/Relations/BelongsOrMorphTo.php
```diff
@@ -0,0 +1,89 @@
...
+         * Example usage:
+         *
+         *     $model->bindEvent('model.relation.beforeAssociate', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+         *         if ($relationName === 'dummyRelation') {
+         *             throw new \Exception("Invalid relation!");
+         *         }
+         *     });
+         *
+         */
+        $this->parent->fireEvent('model.relation.beforeAssociate', [$this->relationName, $this->related]);
```
Rike-cz: $model missing here in params. `$this->related` is only dummy instance. Method's `$model` is real instance **OR id** of associated model, which is good to know inside event handling.
Identically in _afterAssociate_.

src/Database/Relations/HasOneOrMany.php
```diff
@@ -56,6 +56,21 @@ public function create(array $attributes = [], $sessionKey = null)
...
+             * Example usage:
+             *
+             *     $model->bindEvent('model.relation.beforeAdd', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+             *         if ($relationName === 'dummyRelation') {
+             *             throw new \Exception("Invalid relation!");
+             *         }
+             *     });
+             *
+             */
+            $this->parent->fireEvent('model.relation.beforeAdd', [$this->relationName, $this->related]);
```
Rike-cz: Identically as above. Related model here is method's `$model`. Identically in _afterAdd_, _before/afterRemove_ and _MorphOneOrMany_ too.

Rike-cz (commented):
Ok, my feedback.

src/Database/Relations/BelongsToMany.php
mjauvin: You're definitely right here, `parent::attach()` has changed in Laravel 6 and these changes are lost because of this. We should definitely call `parent::attach()`

src/Database/Relations/BelongsOrMorphTo.php
mjauvin: @Rike-cz When you handle the event, you already have the model instance, just pass it to the handler using `use ($model)` as in the examples.

src/Database/Relations/BelongsOrMorphTo.php
mjauvin: right.

src/Database/Relations/BelongsOrMorphTo.php
Rike-cz: @mjauvin Yes, but it's still right side of this inversional relation (modelB belongsTo modelA). I also need modelB instance. `$this->related` is dummy instance of right model, but how to get instance of currently associated model?

Usecase, Category hasMany Products as well Product belongsTo Category. 
```
$product = Product::find(1);
$category = Category::find(1); // or simply int 1
$product->category()->associate($category);
```

This is Product model constructor:
```
public function __construct(array $attributes = [])
{
    parent::__construct($attributes);
    $this->bindEvent('model.relation.beforeAssociate', function ($relationName, $related) {
        // $this is Product model, good
        // $relationName is 'category', good, Product belongsTo Category via category relation
        // $related is dummy instance of Category model, ID === null
    });
}
```
Where I can find Category instance with ID 1, which was actually associated?

mjauvin: Yes, you are correct, I will fix this.
What about in `dissociate()` event handler ?

Rike-cz: I am not sure if anywhere is related model fetched when dissociate. I must look deeper.

Rike-cz: Hmm, only when remove method called. For me, dissociate without related model.

src/Database/Relations/AttachOneOrMany.php
```diff
@@ -166,6 +166,21 @@ public function add(Model $model, $sessionKey = null)
...
+             * Example usage:
+             *
+             *     $model->bindEvent('model.relation.beforeAdd', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+             *         if ($relationName === 'dummyRelation') {
+             *             throw new \Exception("Invalid relation!");
+             *         }
+             *     });
+             *
+             */
+            $this->parent->fireEvent('model.relation.beforeAdd', [$this->relationName, $this->related]);
```
Rike-cz: Last observation: AttachOneOrMany is special case of HasOneOrMany, so `$this->related` could be replaced with `$model` too. After that it will be done for me.

mjauvin: done @Rike-cz !

Rike-cz: @mjauvin I think not `$this->model`, but `$model` only.

mjauvin: oops, my bad... fixed.

src/Database/Relations/MorphTo.php
```diff
@@ -22,6 +22,90 @@ public function __construct(Builder $query, Model $parent, $foreignKey, $otherKe
...
+         * Example usage:
+         *
+         *     $model->bindEvent('model.relation.afterAssociate', function (string $relationName, \October\Rain\Database\Model $relatedModel) use (\October\Rain\Database\Model $model) {
+         *         $relatedClass = get_class($relatedModel);
+         *         $modelClass = get_class($model);
+         *         traceLog("{$relatedClass} was associated as {$relationName} to {$modelClass}.");
+         *     });
+         *
+         */
+        $this->parent->fireEvent('model.relation.afterAssociate', [$this->relationName, $this->related]);
```
Rike-cz: Missed last one `$this->related`, really last :-)

Rike-cz (changes requested):

src/Database/Relations/MorphTo.php
mjauvin: done!

Rike-cz (approved):

mjauvin:
This is ready guys.

mjauvin:
Can this be added to Milestone 1.1.2?

LukeTowers:
Will add it to 1.1.3 instead

src/Database/Relations/BelongsToMany.php
```diff
@@ -133,17 +135,14 @@ public function attach($id, array $attributes = [], $touch = true)
...
          */
-        if ($this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, $insertData], true) === false) {
-            return;
-        }
+        $this->parent->fireEvent('model.relation.beforeAttach', [$this->relationName, $attachedIdList, $insertData]);
 
-        /**
-         * @see Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithPivotTable
+        /*
+         * @See Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithPivotTable
```
bennothommo: @mjauvin Wasn't the original change correct?
```suggestion
        /**
         * @see \Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithPivotTable
```

daftspunk:
Still some work needed on this, in particular, adherence to [the developer guide](https://octobercms.com/help/guidelines/developer#event-naming) and before events should halt to meet continuity with the rest of the platform

Overall looks good though, will be available in next release. Thanks @mjauvin 🙏

PR #541, opened 2020-11-15
Change to Illuminate Database Manager

Denoder:
I haven't found any reference to October's variant of the database manager. If anyone can tell me then we can ignore this pull.

### DIFF
--- a/src/Foundation/Application.php
+++ b/src/Foundation/Application.php
@@ -448,7 +448,7 @@ class Application extends ApplicationBase
             'config'               => [\Illuminate\Config\Repository::class, \Illuminate\Contracts\Config\Repository::class],
             'cookie'               => [\Illuminate\Cookie\CookieJar::class, \Illuminate\Contracts\Cookie\Factory::class, \Illuminate\Contracts\Cookie\QueueingFactory::class],
             'encrypter'            => [\Illuminate\Encryption\Encrypter::class, \Illuminate\Contracts\Encryption\Encrypter::class],
-            'db'                   => [\October\Rain\Database\DatabaseManager::class],
+            'db'                   => [\Illuminate\Database\DatabaseManager::class, \Illuminate\Database\ConnectionResolverInterface::class],
             'db.connection'        => [\Illuminate\Database\Connection::class, \Illuminate\Database\ConnectionInterface::class],
             'events'               => [\Illuminate\Events\Dispatcher::class, \Illuminate\Contracts\Events\Dispatcher::class],
             'files'                => [\Illuminate\Filesystem\Filesystem::class],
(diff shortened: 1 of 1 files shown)


mjauvin:
Indeed, good catch!

mjauvin:
I only found one reference for this alias here:

https://github.com/octobercms/library/blob/1d366ab303819373e29d1ddde8a2c4d01585ad13/src/Support/Facades/Schema.php#L31

Did your PR originate from this @Teranode ?

Denoder:
It didn't. I maintain a auth bridge for laravel's default auth drivers and was just looking at where the application containers lead to and didn't see a file being referenced for that.

mjauvin:
Actually, it's referenced in those as well:
```
src/Support/Facades/Schema.php:        $builder = static::$app['db']->connection($name)->getSchemaBuilder();
src/Support/Facades/Schema.php:        $builder = static::$app['db']->connection()->getSchemaBuilder();
src/Foundation/Application.php:            'db'                   => [\October\Rain\Database\DatabaseManager::class],
src/Database/DatabaseServiceProvider.php:        Model::setConnectionResolver($this->app['db']);
src/Database/DatabaseServiceProvider.php:        $this->app->singleton('db', function ($app) {
src/Database/DatabaseServiceProvider.php:            return $app['db']->connection();
src/Database/DatabaseServiceProvider.php:            return new Dongle($this->getDefaultDatabaseDriver(), $app['db']);
src/Database/DatabaseServiceProvider.php:        $defaultConnection = $this->app['db']->getDefaultConnection();
src/Validation/ValidationServiceProvider.php:            if (isset($app['db'], $app['validation.presence'])) {
```

LukeTowers:
I don't see an "october version" of this file / class period, so I'm curious as to how this is even working as it is right now.

bennothommo:
@LukeTowers unless I'm mistaken, doesn't the alias get overwritten (or superceded) by the `db` singleton that's provided by the `October\Rain\Database\DatabaseServiceProvider` class, anyway?

LukeTowers:
@bennothommo it would appear so, which makes that entire line redundant in the first place. I'm assuming the application works fine if we just remove that line entirely?

Denoder:
> @bennothommo it would appear so, which makes that entire line redundant in the first place. I'm assuming the application works fine if we just remove that line entirely?

Removing the line from the Application file does indeed affect nothing

LukeTowers:
@Teranode actually seeing as all the other ones are included anyways we might be better to leave it with the illuminate versions from Laravel to serve a documentation purpose

bennothommo:
@LukeTowers I'd prefer that we either remove it, or extend it via calling the parent method first and adding our own aliases as required - I noticed the parent method has a few aliases that we don't specify, and we currently just overwrite the method with our own.

LukeTowers:
We don't necessarily include everything Laravel includes by default though, for example `auth`.

bennothommo:
Ah, fair point.

Denoder:
So reverse and put back Illuminate?

bennothommo:
@Teranode I reckon we change it to match Laravel (https://github.com/laravel/framework/blob/6.x/src/Illuminate/Foundation/Application.php#L1189), as we don't have any classes that replace the classes that Laravel alias for `db`. @LukeTowers is right - I can't find any evidence of us having a `DatabaseManager` class in at least the last 5 years.

PR #543, opened 2020-11-28
Fix belongsToMany relation

mjauvin:
Follows on from #536.

I also added the test that was failing from october/october to make sure this actually fixes the problem.

### DIFF
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -417,5 +417,4 @@ class BelongsToMany extends BelongsToManyBase
 
         return $query->join($this->related->getTable(), $this->relatedPivotKey, '=', $this->relatedKey);
     }
-
 }
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -130,14 +130,7 @@ class BelongsToMany extends BelongsToManyBase
             return;
         }
 
-        // Here we will insert the attachment records into the pivot table. Once we have
-        // inserted the records, we will touch the relationships if necessary and the
-        // function will return. We can parse the IDs before inserting the records.
-        $this->newPivotStatement()->insert($insertData);
-
-        if ($touch) {
-            $this->touchIfTouching();
-        }
+        parent::attach($id, $attributes, $touch);
 
         /**
          * @event model.relation.afterAttach
@@ -164,7 +157,7 @@ class BelongsToMany extends BelongsToManyBase
     {
         $attachedIdList = $this->parseIds($ids);
         if (empty($attachedIdList)) {
-            $attachedIdList = $this->allRelatedIds()->all();
+            $attachedIdList = $this->newPivotQuery()->lists($this->relatedPivotKey);
         }
 
         /**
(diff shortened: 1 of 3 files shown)


src/Database/Relations/BelongsToMany.php
```diff
@@ -411,12 +404,17 @@ public function getRelatedIds($sessionKey = null)
...
-    protected function getCurrentlyAttachedPivots()
+    public function newPivotQuery()
     {
-        return $this->getQuery()->get();
+        $query = parent::newPivotQuery();
+
+        // add relation's conditions and scopes to the query
+        $this->addDefinedConstraintsToQuery($query);
+
+        return $query->join($this->related->getTable(), $this->relatedPivotKey, '=', $this->relatedKey);
```
bennothommo: @mjauvin Is there a reason this join is included?

mjauvin: Yes, that's how the Constraints are made to work with the related table @bennothommo

bennothommo: @mjauvin Isn't that already done in the parent call though? I'm concerned this is adding an unnecessary constraint to *all* pivot queries.

mjauvin: newPivotQuery() only returns the pivot table query, in order to apply the constraints, the related table needs to be joined to the Pivot table.

mjauvin: In Laravel, they don't have "conditions" and "scopes" when defining the relations, so this is why their `newPivotQuery()` does not need to join with the related table.

bennothommo: @mjauvin Unless I'm mistaken, these constraints are already added through the RelationController at various points, so this is effectively adding them again and then adding a join.

Before these changes, there was no necessity to implement a join, or even to overwrite the `newPivotQuery` method - of course, this didn't cover the particular case where there's two relations with the same target model but with different conditions, but that implies it *was* working with the more likely scenario of one relation with the target model.

I'm just trying to make sure this doesn't introduce an unintended side effect.

mjauvin: @bennothommo you mean in `modules/backend/behaviors/RelationController` ? That code ?
```
            /*  
             * Apply defined constraints
             */
            if ($sqlConditions = $this->getConfig('view[conditions]')) {
                $widget->bindEvent('list.extendQueryBefore', function ($query) use ($sqlConditions) {
                    $query->whereRaw($sqlConditions);
                });
            }
            elseif ($scopeMethod = $this->getConfig('view[scope]')) {
                $widget->bindEvent('list.extendQueryBefore', function ($query) use ($scopeMethod) {
                    $query->$scopeMethod($this->model);
                });
            }
            else {
                $widget->bindEvent('list.extendQueryBefore', function ($query) use ($widget) {
                    $this->relationObject->addDefinedConstraintsToQuery($query);
                    if ($widget->getSortColumn()) {
                        $query->getQuery()->orders = [];
                    }
                });
            }
```

Denoder:
I hadn't updated my library in a bit (since my pull request). The when i updated to this release i ended up getting a lot of ambiguous errors: 


i had to remove 

```
    /**
     * Create a new query builder for the pivot table.
     *
     * This is an extension of Laravel's `newPivotQuery` method that allows `belongsToMany` and `morphToMany` relations
     * to have conditions.
     *
     * @return \Illuminate\Database\Query\Builder
     */
    public function newPivotQuery()
    {
        $query = parent::newPivotQuery();

        // add relation's conditions and scopes to the query
        $this->addDefinedConstraintsToQuery($query);

        return $query->join($this->related->getTable(), $this->relatedPivotKey, '=', $this->relatedKey);
    }

    ```

bennothommo:
@Teranode are you using either of those fields for conditions or scope filters when defining the relation?

Denoder:
@bennothommo None of them uses conditions or scopes.

mjauvin:
@Teranode Can you show the relations definitions for the models exhibiting this error?

Also, the migration table for the pivot table would be helpful.

mjauvin:
@Teranode maybe opening a new issue for this would be better... you can reference this PR for reference.

PR #545, opened 2020-11-30
Extend Laravel Events Dispatcher

mjauvin:
We can reuse pretty much everything, just a few additions.

Need to investigate if the `Macroable` trait could cause issues. @bennothommo ?


mjauvin:
ping!

PR #547, opened 2020-12-13
Fix mailer.beforeAddContent event handler params order in docblock

mjauvin:


mjauvin:
This is a trivial comment fix.

PR #549, opened 2020-12-18
Add trusted hosts support to library

bennothommo:
Adds support for a new configuration value `app.trustedHosts` that defines the allowed hosts for the application. This allows a developer to prevent [host header poisoning](https://portswigger.net/web-security/host-header).

Possible values:
- `true`: Trust the host specified in app.url, as well as all subdomains.
- `false`: Disable the trusted hosts feature.
- array: Defines the domains to be trusted hosts. Each item should be a string defining a domain, or a regex pattern.


src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,64 @@
...
+     * @return array
+     */
+    public function hosts()
+    {
+        $hosts = Config::get('app.trustedHosts', []);
+
+        if (is_array($hosts)) {
+            foreach ($hosts as &$host) {
+                // Allow for people including protocol in their configured hosts
+                if (starts_with($host, 'http')) {
```
LukeTowers:
```suggestion
                if (starts_with($host, 'http://') || starts_with($host, 'https://') {
```

LukeTowers:
@bennothommo could you look at adapting the tests from https://github.com/laravel/framework/pull/27206/files#diff-58962468d2999fa537014332347c02f1989dddd182b7645732ad350c4aa77bab?

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,64 @@
...
+                // Ensure that trusted hosts specified as just a plain hostname are
+                // properly converted into a strict hostname matching pattern,
+                // otherwise example.com allows sub.example.com
+                if (filter_var($host, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME)) {
+                    $host = '^' . preg_quote($host) . '$';
+                }
+            }
+        }
+
+        if ($hosts === true) {
```
bennothommo:
```suggestion
        } elseif ($hosts === true) {
```

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            // Do strict checks for IP address hosts
+            if (filter_var($host, FILTER_VALIDATE_IP)) {
+                return '^' . preg_quote($host) . '$';
+            }
+
+            // Allow domains to be matched against a www subdomain
+            if (filter_var($host, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME)) {
+                if (preg_match('/^www\.(.*?)$/i', $host, $matches)) {
```
LukeTowers: Could you explain what this regex is doing here? Also, I don't think that we should be automatically adding the www subdomain in any case except for trustedHosts = true, if people are explicitly defining hosts with an array then they shouldn't be having those explicit definitions automatically changed at all.

tests/Foundation/Http/Middleware/CheckForTrustedHostTest.php
```diff
@@ -0,0 +1,145 @@
...
+        $urlGenerator->to('/');
+    }
+
+    public function testThrowExceptionForUntrustedServerAddr()
+    {
+        $this->expectException(SuspiciousOperationException::class);
+
+        $trustedHosts = ['octobercms.com'];
+        $headers = [];
+        $servers = ['SERVER_NAME' => 'malicious.com'];
```
LukeTowers: This looks identical to the untrusted server name test

tests/Foundation/Http/Middleware/CheckForTrustedHostTest.php
```diff
@@ -0,0 +1,145 @@
...
+
+    public static function setUpBeforeClass(): void
+    {
+        self::$orignalTrustHosts = Request::getTrustedHosts();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        Request::setTrustedHosts(self::$orignalTrustHosts);
+    }
```
LukeTowers, line 25: Can we also get a test for a trusted host specified as a regex pattern?

src/Foundation/Http/Middleware/CheckForTrustedHost.php
bennothommo: Fair point - will change it.

bennothommo: The regex here was checking if the given domain starts with `www`, and if so, to allow the domain without the `www` subdomain, given that quite a few hosts (most notably cPanel) will automatically set both to point to the same site.

tests/Foundation/Http/Middleware/CheckForTrustedHostTest.php
bennothommo: Whoops, that's supposed to be SERVER_ADDR - fixed.

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,82 @@
+<?php namespace October\Rain\Foundation\Http\Middleware;
+
+use Config;
+use Illuminate\Http\Middleware\TrustHosts as BaseMiddleware;
```
LukeTowers: If we want to backport this to 5.5 I don't think this middleware exists in 5.5.

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,82 @@
...
+                return [];
+            }
+
+            // If app.url is used, allow both the domain and the `www` subdomain
+            $host = parse_url($url, PHP_URL_HOST);
+
+            if (preg_match('/^www\.(.*?)$/i', $host, $matches)) {
+                return '^(www\.)?' . preg_quote($matches[1]) . '$';
+            } else {
+                return '^(www\.)?' . preg_quote($host) . '$';
```
LukeTowers: Shouldn't these be `$host =` instead of `return` given that the return type is an array? Also, could you add an inline comment to explain why there's two different branches here? (if I'm not mistaken, something like `// Ensure that www.example.com and example.com app.url values are normalized to allow both www.example.com and example.com`

src/Foundation/Http/Middleware/CheckForTrustedHost.php
bennothommo: ~~@LukeTowers I'm happy to leave this as a L6 branch only change - I feel it's a new feature moreso than a security fix.~~

EDIT: Scratch that - OTOH, it was a properly disclosed security issue. We'll just have to port the middleware for 5.5 as well.

bennothommo: @LukeTowers once this has been merged in the L6 branch, I'll do a separate PR for the 5.5 branch.

src/Foundation/Http/Middleware/CheckForTrustedHost.php
bennothommo: @LukeTowers does the comment here suffice? https://github.com/octobercms/library/pull/549/files/2cc51c9f523a24c5e2aabb54ce86eb754c69a0e6#diff-25fa1eb570e4c381198dcffea34a2c648e9d3ad5d05afe7a657d156fd120d7b0R48

src/Foundation/Http/Middleware/CheckForTrustedHost.php
LukeTowers: If we have to have a separate middleware anyways we might as well just do that at this point so that we can have a clean merge back into 5.5

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            // If app.url is used, allow both the domain and the `www` subdomain
+            $host = parse_url($url, PHP_URL_HOST);
+
+            if (preg_match('/^www\.(.*?)$/i', $host, $matches)) {
+                $host = '^(www\.)?' . preg_quote($matches[1]) . '$';
+            } else {
+                $host = '^(www\.)?' . preg_quote($host) . '$';
+            }
```
LukeTowers:
```suggestion
            // Allow both the domain and the `www` subdomain for app.url
            // regardless of the presence of www in the app.url value
            $host = parse_url($url, PHP_URL_HOST);
            if (preg_match('/^www\.(.*?)$/i', $host, $matches)) {
                $host = '^(www\.)?' . preg_quote($matches[1]) . '$';
            } else {
                $host = '^(www\.)?' . preg_quote($host) . '$';
            }
```

src/Foundation/Http/Middleware/CheckForTrustedHost.php
bennothommo: Done :)

src/Foundation/Http/Middleware/CheckForTrustedHost.php
```diff
@@ -0,0 +1,82 @@
...
+            return [];
+        }
+
+        $hosts = array_map(function ($host) {
+            // If a URL is provided, extract the host
+            if (filter_var($host, FILTER_VALIDATE_URL)) {
+                $host = parse_url($host, PHP_URL_HOST);
+            }
+
+            // Do strict checks for IP address hosts
```
LukeTowers:
```suggestion
            // Prepare IP address & plain hostname values to be processed by the regex filter
```

PR #550, opened 2020-12-21
Multiple fixes related to PR #543

### LINKED ISSUE #5479: Calling sync() twice on many-to-many relationship fails in the dev branch, works in 1.0
- OctoberCMS Build: Develop branch; tested with laravel 6.18.43 or 6.20.15
- PHP Version: 7.4.8
- Database Engine: MySql (also fails in SqLite)
- Plugins Installed: https://github.com/cjkpl/oc-sync-plugin (contains the test case)

### Description:
Having two models, e.g. Student and Lesson, and many to many relationship between them, calling sync() second time fails:
```
        $s = Student::first();
        $l = Lesson::first();

        $s->lessons()->sync([$l->id]);
        $s->lessons()->sync([$l->id]);
```

### Steps To Reproduce:
1. Install a test plugin prepared to demonstrate the issue: https://github.com/cjkpl/oc-sync-plugin.git
2. Run unit test on the plugin.
3. Observed behavior: will work on the 1.0 branch (with Laravel 5.x); will throw and error on dev branch (with Laravel 6.x).

mjauvin:
Related to #543 
- $attachedIdList should be an array/collection not an object
- Add fully qualified key names in newPivotQuery() to prevent conflicts.

Fixes https://github.com/octobercms/october/issues/5479#issuecomment-772704701

### DIFF
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -89,6 +89,18 @@ class BelongsToMany extends BelongsToManyBase
     }
 
     /**
+     * Override sync() method of BelongToMany relation in order to flush the query cache.
+     * @param array $ids
+     * @param bool $detaching
+     * @return array
+     */
+    public function sync($ids, $detaching = true)
+    {
+        parent::sync($ids, $detaching);
+        $this->flushDuplicateCache();
+    }
+
+    /**
      * Create a new instance of this related model with deferred binding support.
      */
     public function create(array $attributes = [], array $pivotData = [], $sessionKey = null)
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -160,7 +160,7 @@ class BelongsToMany extends BelongsToManyBase
     {
         $attachedIdList = $this->parseIds($ids);
         if (empty($attachedIdList)) {
-            $attachedIdList = $this->allRelatedIds();
+            $attachedIdList = $this->allRelatedIds()->all();
         }
 
         /**
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -137,6 +137,7 @@ trait DefinedConstraints
         // add relation's conditions and scopes to the query
         $this->addDefinedConstraintsToQuery($query);
 
-        return $query->join($this->related->getTable(), $this->relatedPivotKey, '=', $this->relatedKey);
+        $related = $this->getRelated();
+        return $query->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey());
     }
 }
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -138,6 +138,9 @@ trait DefinedConstraints
         $this->addDefinedConstraintsToQuery($query);
 
         $related = $this->getRelated();
-        return $query->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey());
+
+        return $query
+            ->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey())
+            ->select($this->getTable().'.*');
     }
 }
--- a/tests/Database/RelationsTest.php
+++ b/tests/Database/RelationsTest.php
@@ -29,6 +29,8 @@ class RelationsTest extends DbTestCase
             $table->primary(['post_id', 'term_id']);
             $table->unsignedInteger('post_id');
             $table->unsignedInteger('term_id');
+            $table->string('data')->nullable();
+            $table->timestamps();
         });
     }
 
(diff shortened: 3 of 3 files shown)


mjauvin:
I introduced that bug in my last PR (https://github.com/octobercms/library/pull/543)

That should REALLY be added to the 1.1.2 release, otherwise we'll be in trouble...

mjauvin:
@Teranode can you check if this PR resolves [your issues](https://github.com/octobercms/library/pull/543#issuecomment-748677293) ?

Thanks.

LukeTowers:
@mjauvin can you add some tests for the issue?

Denoder:
Does not fix the issue it's still the same, I even tried just doing:

```php
    /**
     * Create a new query builder for the pivot table.
     *
     * This is an extension of Laravel's `newPivotQuery` method that allows `belongsToMany` and `morphToMany` relations
     * to have conditions.
     *
     * @return \Illuminate\Database\Query\Builder
     */
    public function newPivotQuery()
    {
        parent::newPivotQuery();
    }
}
```
and it gives this error:


-----------

this is the method that get's the error:

```php
        $collection = $user->collection()->updateExistingPivot($animeId, [
            'episodes' => $request->episodes,
            'status' => $request->status
        ]);
```

and the relation:
```php
    /**
     * @var array Relations
     */
    public $belongsToMany = [
        'collection' => [
            Anime::class, 
            'table' => 'anime_queue',
            'key' => 'user_id',
            'otherKey' => 'anime_id',
            'pivot' => ['status', 'episodes', 'score', 'anilist_id'],
            'timestamps' => true
        ],
        'watchlist' => [
            Episode::class,
            'table' => 'anime_episodes_watchlist',
            'key' => 'user_id',
            'otherKey' => 'episode_id',
            'pivot' => ['progress'],
            'timestamps' => true
        ]
    ];
```

anything utilizing a pivot would get affected.
Whether im using updateExistingPivot or syncWithoutDetaching, if i touch the pivot it will error out.

mjauvin:
@Teranode in the code you modified, you need to RETURN the query or the parent's method response...

mjauvin:
What is the exact error when using this PR as is?

Denoder:
> @Teranode in the code you modified, you need to RETURN the query or the parent's method response...

you're right, that fixed it.


> What is the exact error when using this PR as is?

mjauvin:
I don't understand, do you have a `status` field in the pivot table AND in the related table?

Denoder:
@mjauvin yes, and that could easily be solved by just specifying the table name within `updateExistingPivot` however this error did not occur at all before, it's also saying that the update field is ambiguous because both the tables have an updated_at field

mjauvin:
Yes, we're going to have to make more changes, this is now joining two tables which was not the case before.

mjauvin:
@LukeTowers ~how can we select specific fields on the joined table (not the pivot) for the query?~

-- never mind

mjauvin:
@Teranode could you try adding a select to the newPivotQuery() to select only the fields of the pivot table to restore previous behavior? 

Something like this:
```
    public function newPivotQuery()
    {   
        $query = parent::newPivotQuery();

        // add relation's conditions and scopes to the query
        $this->addDefinedConstraintsToQuery($query);

        $related = $this->getRelated();

        return $query
            ->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey())
            ->select($this->getTable().'.*');
    }
```

Denoder:
the same thing happens unfortunately :(

mjauvin:
@Teranode I tried to reproduce this and I failed...

Are you sure you are using the 1.1.x branch?

mjauvin:
@Teranode do you happen to use cast/json fields in your model? Can you how the full model class?

Denoder:
here: https://gist.github.com/Teranode/36d2c4328d1704fdfe2ab3d5dbfc3d50

mjauvin:
@Teranode can you try different things to pinpoint why you are getting the error and I am not?

Can you temporarily remove the TranslatableModel behavior just to make sure this is not causing the problem?

Thanks for your help with this man!

mjauvin:
I added a pivot data test that calls `updateExistingPivot()` to try to reproduce your issue. Can you take a look at the test and see what's different with your usage?

Denoder:
Would this work?

```
    /**
     * Create a new query builder for the pivot table.
     *
     * This is an extension of Laravel's `newPivotQuery` method that allows `belongsToMany` and `morphToMany` relations
     * to have conditions.
     *
     * @return \Illuminate\Database\Query\Builder
     */
    public function newPivotQuery()
    {
        $query = parent::newPivotQuery();

        // add relation's conditions and scopes to the query
        $this->addDefinedConstraintsToQuery($query);

        $related = $this->getRelated();

        return $query;
    }
```

adding the join breaks it, even just leaving the select works, but when you add the join it breaks.

Denoder:
also here's another gist of a section using the updateExistingPivot/syncWithoutDetaching:
https://gist.github.com/Teranode/b7fd284bbddbbf987e822085329d5095

any instance of that being used will just not work, and i have several files that uses updateExistingPivot.

mjauvin:
@Teranode any way to create a plugin on github with your models and controller that would help me to diagnose this? That'd be awesome!

Denoder:
I don't think I can do that, this seems to only be a problem with me I suppose. I don't know why, so I guess ill just put this in my bridge. No one else seems to have reported an issue.

LukeTowers:
@Teranode it would really help if you could, if it's a problem with you then it's quite possible it will affect others too. You seem to be the only one reporting it at the moment because you actually use the latest code.

mjauvin:
@LukeTowers I added the tests but still cannot reproduce this myself.

mjauvin:
@cjkpl if you want to test this, I've merged the change I showed you earlier.

Denoder:
I'm still recieving the same error if i keep
```
    ->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey())
    ->select($this->getTable().'.*')
```

mjauvin:
@Teranode please provide us with a plugin to test this, or a PR to the October Test plugin so that we can replicate the issue.

PR #551, opened 2020-12-23
Converting bound events to serializable closures to allow caching

interworks-morr:
Solves "Exception: Serialization of 'Closure' is not allowed" error when caching models to a file store that have traits added.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "league/csv": "~9.1",
         "nesbot/carbon": "^2.0",
         "laravel/framework": "~6.0",
-        "laravel/tinker": "~2.0"
+        "laravel/tinker": "~2.0",
+        "opis/closure": "^3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.4|^9.3.3",
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "nesbot/carbon": "^2.0",
         "laravel/framework": "~6.0",
         "laravel/tinker": "~2.0",
-        "opis/closure": "^3.1"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -45,7 +45,8 @@
         "league/csv": "~9.1",
         "nesbot/carbon": "^2.0",
         "laravel/framework": "~6.0",
-        "laravel/tinker": "~2.0"
+        "laravel/tinker": "~2.0",
+        "opis/closure": "^3.1"
```
LukeTowers: Needs to be 3.6 at a minimum because that's what Laravel 6 requires: https://github.com/laravel/framework/blob/6.x/composer.json#L29.

mjauvin: if I'm not mistaken, Laravel 6 already includes this, so this should not be required on 1.1.x

interworks-morr: The intention is to patch this for the next release of OctoberCMS.  I'm not aware of your release time frames for 1.1, Laravel 6, etc.  Will there be more versions between now and when those are made available via october:update, therefore a better branch to merge this against, or shall I just remove the requirement from composer since Laravel 6 already provides it?

LukeTowers: @interworks-morr v1.1 with Laravel 6 is already out via composer, it's just the marketplace that's taking a while to get upgraded. I'm fine with leaving it in since we're now using it directly, we just need to bump up the requirement to 3.6

LukeTowers:
Would it be possible for you to add some unit tests for this?

daftspunk:
There isn't a clear reason or use case why this is needed and I suspect it occurs too low in the stack. Feel free to elaborate and we can revisit

PR #553, opened 2021-01-05
Add auto-detach option into belongsToMany relationship

Rike-cz:
Discussed here: https://github.com/octobercms/october/discussions/5445

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -928,7 +928,9 @@ class Model extends EloquentModel
              */
             if ($type == 'belongsToMany') {
                 foreach ($relations as $name => $options) {
-                    $this->{$name}()->detach();
+                    if (Arr::get($options, 'detach', true)) {
+                        $this->{$name}()->detach();
+                    }
                 }
             }
         }


daftspunk:
Hi @Rike-cz, this would almost always lead to corrupted data in the join table. Can you describe in more detail why you want this included?

Rike-cz:
@daftspunk I had nice examples in discussion. Where is it? Link above fails with 404. 
When you use belongsToMany to get data by non-primary keys, you are able to create relation, that looks little like "belongsToManyThrough", but with better and cheaper handling. For example you have 3 models: Category, ProductSet and Product. Product belongsTo ProductSet, ProductSet belongsToMany Categories.
I can make with variation of 'key', 'otherKey', 'parentKey', and 'relatedKey' Product belongsToMany Categories shorcut, it's simple and legal, but when deleting this Product, auto-detach kills ProductSet belongsToMany Categories relation, but it's not required in this case. ProductSet stayed in database and there is no reason to destroy relation between ProductSet and Categories.

For my point of view is the best to have behavior of Eloquent close to Laravel natural behavior. Without hidden automations. And enable them by flag, like "delete" in hasMany case. So I added "detach", but inverted due to breaking changes. Developer can switch it off manually.

daftspunk:
That makes sense. I can see how the forceful automation would be a problem. Ok, I've added this for the next release and documented it

https://github.com/octobercms/docs/commit/4f0af67826e3af6bc7940ab6ec779e96f00ad53a

Thanks :pray:

PR #556, opened 2021-01-22
Fix getRelationType method

mjauvin:
morphTo relations have an empty definition, so we need to test against null instead.

The change that lead to this problem was below:

https://github.com/octobercms/library/pull/460/files#diff-b0185ec47eb87315ad4b78f740db185dd938ca525888d249d820c1a7d5be9754R200-R206


mjauvin:
@LukeTowers if you can merge this ASAP so that we can resume PR submission on main october repo, that would be nice.

currently, new PRs and all commits to existing PRs are failing the tests.

PR #560, opened 2021-02-06
Set application key in .env file (if exists) or fallback config file.

### LINKED ISSUE #5496: key:generate command not updating .env file
- OctoberCMS Build: v1.1.1
- PHP Version: 7.4
- Database Engine: MySQL

### Description:
When using a `.env` file for OctoberCMS: `php artisan october:env` the `php artisan key:generate` should generate the key and place it in the `.env` file instead of writing it into the app.php file.

### Steps To Reproduce:
1. Run the command `php artisan october:env`
2. Then `php artisan key:generate`
3. Notice how the key gets generated and placed inside of the app.php file instead of the desired .env file.

If you would like, I can send a PR for this since I think this should be included since that is most likely what the developer wants when using the .env file option (at least I know I do).

ericp-mrel:
Fixes https://github.com/octobercms/october/issues/5496

### DIFF
--- a/src/Foundation/Console/KeyGenerateCommand.php
+++ b/src/Foundation/Console/KeyGenerateCommand.php
@@ -45,14 +45,17 @@ class KeyGenerateCommand extends KeyGenerateCommandBase
             return $this->line('<comment>'.$key.'</comment>');
         }
 
+        $currentKey = $this->laravel['config']['app.key'];
+
+        if (strlen($currentKey) !== 0 && (! $this->confirmToProceed())) {
+            return;
+        }
+
         // Next, we will replace the application key in the env file (if it exists) or config file
         // so it is automatically setup for this developer. This key gets generated using a
         // secure random byte generator and is later base64 encoded for storage.
-        if (
-            ($this->files->exists($this->laravel->environmentFilePath()) && !$this->setKeyInEnvironmentFile($key)) ||
-            !$this->setKeyInConfigFile($key)
-        ) {
-            return;
+        if (!$this->setKeyInEnvironmentFile($key)) {
+            $this->setKeyInConfigFile($key);
         }
 
         $this->laravel['config']['app.key'] = $key;
@@ -61,22 +64,43 @@ class KeyGenerateCommand extends KeyGenerateCommandBase
     }
 
     /**
+     * Set the application key in the environment file.
+     *
+     * @param string $key
+     * @return bool
+     */
+    protected function setKeyInEnvironmentFile($key): bool
+    {
+        if (!$this->files->exists($this->laravel->environmentFilePath())) {
+            return false;
+        }
+
+        $contents = $this->files->get($this->laravel->environmentFilePath());
+
+        $matches = [];
+        preg_match($this->keyReplacementPattern(), $contents, $matches);
+
+        // Abort if the app key is not able to be written to the .env file.
+        if (empty($matches)) {
+            return false;
+        }
+
+        $this->writeNewEnvironmentFileWith($key);
+
+        return true;
+    }
+
+    /**
      * Set the application key in the config file.
      *
      * @param  string  $key
      * @return bool
      */
-    protected function setKeyInConfigFile($key)
+    protected function setKeyInConfigFile($key): bool
     {
-        if (!$this->confirmToProceed()) {
-            return false;
-        }
+        [$path, $contents] = $this->getKeyFile();
 
-        $currentKey = $this->laravel['config']['app.key'];
-
-        list($path, $contents) = $this->getKeyFile();
-
-        $contents = str_replace($currentKey, $key, $contents);
+        $contents = preg_replace($this->configKeyReplacementPattern(), "$1{$key}$3", $contents);
 
         $this->files->put($path, $contents);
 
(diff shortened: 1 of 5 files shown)


src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -45,10 +45,13 @@ public function handle()
...
 
-        // Next, we will replace the application key in the config file so it is
-        // automatically setup for this developer. This key gets generated using a
+        // Next, we will replace the application key in the env file (if it exists) or config file
+        // so it is automatically setup for this developer. This key gets generated using a
         // secure random byte generator and is later base64 encoded for storage.
-        if (!$this->setKeyInConfigFile($key)) {
+        if (
+            ($this->files->exists($this->laravel->environmentFilePath()) && !$this->setKeyInEnvironmentFile($key))
+            || !$this->setKeyInConfigFile($key)
```
LukeTowers:
```suggestion
            ($this->files->exists($this->laravel->environmentFilePath()) && !$this->setKeyInEnvironmentFile($key)) ||
            !$this->setKeyInConfigFile($key)
```

bennothommo:
@ericp-mrel Tested and it works fine, but there's one more scenario we need to cover here. A not-so-common but still supported way of configuration in October is to use a `.env` file just to define the current environment (dev, staging, prod, etc.) and still use the environment configuration directories under `config/dev`, `config/staging`, etc.

Your code however will only try the environment file if it exists, but won't write anything as there's no `APP_KEY` item defined. We'll need to do a check in `.env` first to see if there's an `APP_KEY` item that is defined - if not, it needs to fall back to the config file scenario, which I believe covers the multiple environment setup.

ericp-mrel:
@bennothommo I've added functionality so it will check if the environment specific app config file exists and write the key there instead of only falling back to the main app config file.

Do you think it's worth me adding checks to actually validate the app key line exists in that file as well (e.g. /config/dev/app.php)?
`'key' => 'CHANGE_ME!!!!!!!!!!!!!!!!!!!!!!!',`

bennothommo:
@ericp-mrel That would be good. Ideally, the flow would be as follows:

- **If .env exists**
  - If `APP_KEY` is specified, write the key to .env
  - If `APP_KEY` is not specified, but `APP_ENV` is specified, go to **If environment config exists** and try to apply key to specified environment.
  - Else, go to **Else**.
- **If environment config exists**
  - If `config/<environment>/app.php` exists for environment, and `key` attribute is available, write key to this config value.
  - Else, go to **Else**.
- **Else**
  - If `config/app.php` exists, and `key` attribute is available, write key to this config value.
  - Else, do nothing.

ericp-mrel:
@bennothommo I've added the necessary logic to properly check the environment specific app config file after the .env file check and then the fallback to the main app config file.

I've also switched to using a regex for the key replacement in order to fix the issue where the app key wasn't being updated if it was originally empty since it was trying to perform a string replacement for an empty string on the entire file, which obviously doesn't result in any change.

Let me know if there's any code you think I could clean up.

Thanks

src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -45,35 +45,62 @@ public function handle()
             return $this->line('<comment>'.$key.'</comment>');
         }
 
-        // Next, we will replace the application key in the config file so it is
-        // automatically setup for this developer. This key gets generated using a
-        // secure random byte generator and is later base64 encoded for storage.
-        if (!$this->setKeyInConfigFile($key)) {
+        $currentKey = $this->laravel['config']['app.key'];
+
+        if (strlen($currentKey) !== 0 && (! $this->confirmToProceed())) {
```
LukeTowers:
```suggestion
        if (!empty($currentKey) && (!$this->confirmToProceed())) {
```

src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -45,35 +45,62 @@ public function handle()
...
             return;
         }
 
+        // Next, we will replace the application key in the env file (if it exists) or config file
+        // so it is automatically setup for this developer. This key gets generated using a
+        // secure random byte generator and is later base64 encoded for storage.
+        if (! $this->setKeyInEnvironmentFile($key)) {
```
LukeTowers:
```suggestion
        if (!$this->setKeyInEnvironmentFile($key)) {
```

src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -85,12 +112,36 @@ protected function setKeyInConfigFile($key)
...
 
         return [$path, $contents];
     }
+
+    /**
+     * Get a regex pattern that will match config APP_KEY with any random key
+     *
+     * @return string
+     */
+    protected function configKeyReplacementPattern(): string
```
LukeTowers: Isn't there a ConfigWriter class built into October? Why are we reinventing the wheel here?

ericp-mrel: I was not aware of the config writer class, but I tried using it just now and it looks like it will need to be modified in order to support updating values which contain an env call `'key' => env('APP_KEY', 'CHANGE_ME!'),` since it only works with regular key => value lines. `'key' => 'CHANGE_ME',`

I'll try and see if there's anything from this PR that can be used: https://github.com/daftspunk/laravel-config-writer/pull/22/files

bennothommo: @ericp-mrel I wouldn't worry too much about that. If it's using an `env` call, they've set up the site to use the `.env` file.

LukeTowers:
Would it be possible to get some unit tests for each of the cases @bennothommo described?

ericp-mrel:
> Would it be possible to get some unit tests for each of the cases @bennothommo described?

Sorry, I'm new to testing. How exactly would I write a test for this command without affecting the rest of the test since it requires creating/updating existing files (which could affect tests which ran after it).

I tried looking at Laravel itself as a starting point to see how they do it, and they don't have a any tests for the key generate command either.

bennothommo:
@ericp-mrel You could check the tests I wrote for the `october:env` command - they may be able to assist you:
https://github.com/octobercms/october/blob/develop/tests/unit/system/console/OctoberEnvTest.php

bennothommo:
I'd in fact suggest putting the tests in the main October repo as a separate PR - the `key:generate` command functionality makes changes to core code, not framework files.

ericp-mrel:
Alright, I'll send a PR to the main repo, but is there anything special I need to do in order for the tests to actually work with the CI tests since these tests require another PR on the framework repo which isn't merge?

bennothommo:
@ericp-mrel don't stress too much about the CI on the separate repo - the tests will fail, but I'll manually test it with everything merged together when it's ready.

daftspunk:
See comment in https://github.com/octobercms/october/issues/5496

PR #561, opened 2021-02-08
revert breaking change with newPivotQuery

### LINKED ISSUE #5499: belongsToMany detach() fails when scope is defined on the relationship
This is a problem that was introduced in octobercms/library#543. https://github.com/octobercms/library/pull/543/files#diff-56f77bab4b0aa050a15e11466ccccbe82ce391f7adeed778ddc5c66c41152127R138

The issue stems from the fact that the query object here is not a Eloquent query builder associated with a model but a base query builder that doesn't have any access to the relevant scope methods.

You can replicate the issue by using RainLab.Blog, the WP importer plugin with my PR applied https://github.com/klyp/october-wp-importer/pull/7, and then attempting to import posts that have categories associated with them from a WordPress export file. You'll get the error "method scopeIsPublished does not exist"

```php
$category->posts()->detach($post->id);
```

mjauvin:
I throw the towel, I can't find a proper way to fix this for all cases.

Fixes https://github.com/octobercms/october/issues/5499

### DIFF
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -121,5 +121,4 @@ trait DefinedConstraints
             $query->$scope($this->parent);
         }
     }
-
 }
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -122,25 +122,4 @@ trait DefinedConstraints
         }
     }
 
-    /**
-     * Create a new query builder for the pivot table.
-     *
-     * This is an extension of Laravel's `newPivotQuery` method that allows `belongsToMany` and `morphToMany` relations
-     * to have conditions.
-     *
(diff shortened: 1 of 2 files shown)


LukeTowers:
Does this rebreak @Flynsarmy's original issue?

mjauvin:
Yes it does.

bennothommo:
@mjauvin I'm sure we can fix this in some way over reverting it - while @Flynsarmy's original issue was a fringe case, I can imagine it's probably in use out there. I'll try and have another look at it first before suggesting whether to revert or not.

Flynsarmy:
A pretty important fringe considering the functionality is required to create a wordpress-style taxonomy/term system!

bennothommo:
@Flynsarmy I only say fringe because there's several other ways it could be done. ;)

mjauvin:
You're welcome to try @bennothommo

mjauvin:
Replaced by #562

PR #562, opened 2021-02-11
Apply the scope on the related model

### LINKED ISSUE #5499: belongsToMany detach() fails when scope is defined on the relationship

mjauvin:
Fixes https://github.com/octobercms/october/issues/5499

### DIFF
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -118,7 +118,7 @@ trait DefinedConstraints
          * Scope
          */
         if ($scope = array_get($args, 'scope')) {
-            $this->getRelated()->$scope($query);
+            $this->getRelated()->scopes([$scope => [$this->parent]]);
         }
     }
 
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -118,7 +118,7 @@ trait DefinedConstraints
          * Scope
          */
         if ($scope = array_get($args, 'scope')) {
-            $query->$scope($this->parent);
+            $this->getRelated()->$scope($query);
         }
     }
 
--- a/tests/Database/RelationsTest.php
+++ b/tests/Database/RelationsTest.php
@@ -279,7 +279,7 @@ class Post extends \October\Rain\Database\Model
             'otherKey'  => 'term_id',
             'pivot'     => ['data'],
             'timestamps' => true,
-            'conditions' => 'type = "category"',
+            'scope'     => 'isCategory',
         ],
         'terms' => [
             Term::class,
@@ -314,6 +314,16 @@ class Term extends \October\Rain\Database\Model
             'conditions' => 'type = "post"',
         ],
     ];
+
+    public function scopeIsTag($query)
+    {
+        return $query->where('type', 'tag');
+    }
+
+    public function scopeIsCategory($query)
+    {
+        return $query->where('type', 'category');
+    }
 }
 
 class Morphs extends \October\Rain\Database\Model


bennothommo:
I can confirm that this fix resolves the test case that @mjauvin [alluded to](https://github.com/octobercms/october/issues/5499#issuecomment-775175367).

@LukeTowers @Teranode when both of you have a moment, could you test this out with your scenarios and let us know if this resolves them or not?

Denoder:
Tested and does not work. 

I used telescope to try to see what is conflicing. This is what a updateExistingPivot looks like (in regards to my query).

```
update
  `anime_queue`
set
  `episodes` = 64,
  `status` = 'Completed',
  `updated_at` = '2021-02-11 05:13:41'
where
  `user_id` = 1
  and `anime_id` in ('1')
```

and the newPivotQuery uses:

```
$query
            ->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey())
            ->select($this->getTable().'.*');
```

the `anime` table (which it's trying to join) has a `status` field and an `updated_at` field wouldnt this be conflicing since you're joining tables that "might" have the same fields and updating leads to the ambiguous error since it doesn't know which table it wants to update (in this case updated_at is always present if you're using timestamps)

bennothommo:
Thanks @Teranode for letting us know.

Would you be able to provide us with a relation and model(s) config that exhibits this issue so we can replicate and look into it further?

Denoder:
@bennothommo
Here: https://gist.github.com/Teranode/363ab56d34ecfd8688d6123e4a829ecb

src/Database/Relations/DefinedConstraints.php
```diff
@@ -118,7 +118,7 @@ public function addDefinedConstraintsToQuery($query, $args = null)
          * Scope
          */
         if ($scope = array_get($args, 'scope')) {
-            $query->$scope($this->parent);
+            $this->getRelated()->$scope($query);
```
LukeTowers: I'm not understanding how this would work. This would now call the scope method on a model object instead of a eloquent query builder to be passed to a model object, and when you call `$model->{$scope}` the internal magic will instantiate a new base query builder, and pass that query builder instance to the actual method `$model->scope{$scope}($query);`.

If I'm following correctly, I believe that this line would actually have to be

```php
$scopeMethod = 'scope' . ucfirst($scope);
$this->getRelated()->{$scopeMethod}($query, $this->parent);
```

Additionally, if we're to handle this in an even more laravel-y way, I think the correct way of manually calling a model scope without an eloquent builder instance is `$this->getRelated()->scopes([$scope => [$this->parent]]);` (https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Eloquent/Builder.php#L906) (which also does a bunch of other things to improve stability).

LukeTowers: @bennothommo @mjauvin any thoughts on the above?

mjauvin:
@LukeTowers @bennothommo if I modify the relation definition in the test case to use a scope instead of  a condition, this actually does not work either way (with previous or current method of calling the scope). It won't generate an error, but the resulting query is not correct.

I really don't know how to handle this anymore.

bennothommo:
I'll need to adjust @mjauvin's tests so that this particular scenario can be replicated - I'm struggling to follow the workings of where this fix is up to and how it affects Laravel's inner workings.

bennothommo:
I've put in an alternate fix for this (#563) and updated the unit tests to try both conditions and scopes in the relation config. As far as I can tell, this replicates the Blog post bug and shows that @LukeTowers suggested fix seems to resolve it. I haven't yet got my head around @Teranode's config, but once I do, I'll see if anything different is done there and update the tests to suit if needed.

mjauvin:
Replaced by #563

PR #563, opened 2021-02-23
Alternative fix for october#5499

bennothommo:
Applies the fix suggested by @LukeTowers here: https://github.com/octobercms/library/pull/562#discussion_r580083985

Also adjusts the unit tests to try conditions and scopes in the relation config.

Replaces https://github.com/octobercms/library/pull/562

### DIFF
--- a/src/Database/Relations/DefinedConstraints.php
+++ b/src/Database/Relations/DefinedConstraints.php
@@ -118,7 +118,8 @@ trait DefinedConstraints
          * Scope
          */
         if ($scope = array_get($args, 'scope')) {
-            $query->$scope($this->parent);
+            $scopeMethod = 'scope' . ucfirst($scope);
+            $this->getRelated()->{$scopeMethod}($query, $this->parent);
         }
     }
 
@@ -141,6 +142,6 @@ trait DefinedConstraints
 
         return $query
             ->join($related->getTable(), $related->getQualifiedKeyName(), '=', $this->getOtherKey())
-            ->select($this->getTable().'.*');
+            ->select($this->getTable() . '.*');
     }
 }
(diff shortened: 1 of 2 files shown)


bennothommo:
@Teranode could you try this fix instead with your setup (just the changes for [`DefinedConstraints`](https://github.com/octobercms/library/pull/563/files#diff-56f77bab4b0aa050a15e11466ccccbe82ce391f7adeed778ddc5c66c41152127)) and let me know if that has any better luck?

mjauvin:
@bennothommo I confirm this way of calling the scope fixes the problem either using conditions or scope for the relation definition.

I tried Luke's second suggested method and didn't try the first one.

Nice work guys!

Denoder:
@bennothommo Still the same thing

LukeTowers:
@Teranode were you able to provide reproduction instructions with the test plugin? That would help us out a lot with solving your issue.

bennothommo:
@LukeTowers As far as I can tell, @Teranode's scenario is that they're using a field name that is ambiguous in the conditions. Is this correct, @Teranode?

I would say a potential fix is to prefix the field with the relation name which would remove the ambiguity, but I haven't tried that out yet.

Denoder:
> @LukeTowers As far as I can tell, @Teranode's scenario is that they're using a field name that is ambiguous in the conditions. Is this correct, @Teranode?
> 
> I would say a potential fix is to prefix the field with the relation name which would remove the ambiguity, but I haven't tried that out yet.

yes and no, yes in the sense that the `status` field is ambiguous and i could change it. No in the sense that when I do change it. the `updated_at` field gets called ambiguous because the pivot method is ambiguously trying to update the `status` and `updated_at` field which both tables have.

bennothommo:
@Teranode and this is happening when using `updateExistingPivot`?

Denoder:
> @Teranode and this is happening when using `updateExistingPivot`?

yes and also syncWithoutDetaching

bennothommo:
Cool - I'll try add it to the tests.

PR #564, opened 2021-03-11
Fix & improvement to Sortable behavior

mjauvin:
- Fix broken method to check for SORT_ORDER const existance on Class;
  The previous code to check if the SORT_ORDER constant was defined on the model class didn't work.
- Add a more dynamic method to get the sort_order from the model's sort_order_column property;
  You can now dynamically extended a model to change the sort_order_column using addDynamicProperty (since a constant cannot be defined when extending a model).


daftspunk:
Thanks Marc

PR #565, opened 2021-03-25
Use translation string than pluralized placeholder

mohsin:
The controller appears to automatically detect and fill the appropriate model name into the above translation string without having to specify it. So it seems like a good idea to have this set here itself.

Benefits:
1. As a developer I don't need to replace this with the translation string each time for each controller's form config.
2. As a newbie developer who doesn't know that there is a system translation string that automatically fills in the model name, it would be redundant to be creating in-plugin strings for the same.


daftspunk:
Fair points. You can also just remove the `title` property altogether and it will default to these values.

We do want to keep the stub files from getting too messy with locale strings. Since config_list appears to have some generic keys in it already, this should be fine

Thanks!

PR #566, opened 2021-04-07
Make the stylesheet minifier aware of CSS vars

tobias-kuendig:
**Edit:** Found the Blog entry that describes the issue and added the link.

We are using Tailwind in October. Unfortunately, October's stylesheet minifier makes some optimizations that break the Tailwind output.

Tailwind makes heavy use of CSS variables. It also includes some workarounds like the following code:

```css
    --tw-ring-inset: var(--tw-empty, /*!*/ /*!*/);
```

The space after the `,` in the `var` function is actually relevant. The minifier must not remove it. Tailwind inserts these special comments to tell a minifier to leave the space untouched.

Read more about it here: https://adamwathan.me/composing-the-uncomposable-with-css-variables/

The stylesheet minifier currently removes all comments and unnecessary whitespace regardless of the context. It also removes any units of zero values (`0px` becomes `0`).

This PR changes a few of the regexes in the Stylesheet minifier to leave comments inside `()` untouched. It also makes sure spaces between two comments are not removed. Finally, it removes the conversion of zero values with units to values without a unit.

These are the changes to the regexes:

#### Remove spaces before and after comment

`[^\/]` was added to the end so spaces after a comment are only removed, if there is no `/` after the space (makes sure that the space between `/*!*/ /*!*/` is not removed)

#### Remove comment blocks, everything between /* and */

Added a Negative Lookahead at the end, that makes sure that comments inside `()` are ignored. This includes comments inside any `var()` declaration.

#### Remove space after , : ; { } */ >

This regex has been modified so it matches each of the above character followed by a space OR `*/` followed by a space but NOT followed by a space AND a `/`. This is again to prevent the space between comments from being removed.

### DIFF
--- a/src/Assetic/Filter/StylesheetMinify.php
+++ b/src/Assetic/Filter/StylesheetMinify.php
@@ -53,9 +53,6 @@ class StylesheetMinify implements FilterInterface
         // -6.0100em to -6.01em, .0100 to .01, 1.200px to 1.2px
         $css = preg_replace('/((?<!\\\\)\:|\s)(\-?)(\d?\.\d+?)0+([^\d])/S', '$1$2$3$4', $css);
 
-        // Strips units if value is 0 (converts 0px to 0), ignores values inside ()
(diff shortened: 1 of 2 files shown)


daftspunk:
Looks great, thanks @tobias-kuendig !

PR #567, opened 2021-04-15
Fixed StylesheetMinify regex

tobias-kuendig:
This PR fixes the problem introduced in #566.

It is a lot closer to the original minifier. 

1. Added exceptions for `/*!` comments. The comments are left untouched (not removed, spaces before and after remain)
2. I also fixed the `Remove space before , ; { } >` regex as it did not seem to do anything. This means a rule like `body {}` is now properly minified to `body{}` because the space before `{` is now correctly removed.


daftspunk:
Fixed in 43d65ae0a0a82cff15985e9099a17df895249f21

PR #569, opened 2021-06-11
Translated controller scaffolds

prhost:


daftspunk:
We're going to refactor these again soon, but this PR is a good stop-gap until that time. Thanks!

PR #571, opened 2021-08-24
Add option to parse hidden field

zoomyboy:
I added this as an option, because I would like to define the twig markup separately in my template. So you can declare all your fields upfront and render them in which way you want.


zoomyboy:
Any upates on this?

daftspunk:
Hey @zoomyboy 

This looks like a candidate for the `{variable}` field type that can support passing a `type` value

```
<h1>{variable type="text" name="websiteName" label="Website Name"}Our wonderful website{/variable}</h1>
```

Identical test added here:

e292ab27084ae47b15304ec9b44c81bd9659d49a

I hope this helps!

PR #574, opened 2021-11-11
Rehash password when hashing driver has changed

jacobdekeizer:
Resolves octobercms/october-private#95.

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -1,6 +1,7 @@
 <?php namespace October\Rain\Auth;
 
 use Cookie;
+use Illuminate\Support\Facades\Hash;
 use Session;
 use Request;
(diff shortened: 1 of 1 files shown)


prhost:
it seems to me venerable to update the password in an authentication method. Also shouldn't I pass the password hash to needsRehash and not the raw password?

jacobdekeizer:
@prhost yes you're right, gone wrong with refactor. I will update it to use the hashed password. I couldn't find a better placement with the current Auth setup, maybe @daftspunk has some ideas.

daftspunk:
Thanks @jacobdekeizer. This change has been reviewed and implemented across the following commits:

- ab3deb0bcc1c8ad9b0fc824f65425fa608564180 Public API defined
- c05f65a21cdb25c18bdd28af8ca580313f067bce Policy implemented

Best regards

PR #575, opened 2021-11-13
Change order of overrides

alxy:
I think the order of overrides is wrong here.

Attribute names that are explicitly defined on the model instance via the `$attributeNames` property should override any default names. As `array_merge()` uses the last value found if duplicate keys exist, we need to switch the order here to get the described behavior. Otherwise, setting this property has no effect in most cases.


daftspunk:
Good catch; except this PR reduces priority for explicit references, which should be at the top.

Fixed in 20d5d908c6a9e1aefd9d6117010451c2bea42658

PR #577, opened 2022-01-10
Ignore minify if the assets has .min.js

jackyhuynh0407:
Today i catch some error in js when disable debug mode.
And I found the problem is because of the .min.js (which already minified), so I ignore all the minified js, and it work well now.


daftspunk:
This is great, thanks!

PR #578, opened 2022-01-12
generate scss source map when debug mode enable

jackyhuynh0407:
Scss sourcemap is very necessary for debugging. So this PR will generate soucemap when debug mode enable. With this we can online debugging with the editor perfectly


daftspunk:
Thanks for this @jackyhuynh0407 

Reviewed in f1e8f9ad74867d6fae18a0218a58d908d51eefb2

Note the following changes:
- config `cms.enable_asset_source_maps` is needed to be set TRUE 
- the output location differs as `/storage/app/resources/sourcemap`

Best regards

jackyhuynh0407:
Awesome! Thanks very much for your hard working. Your code is clearer and better than me.

PR #583, opened 2022-04-25
Fix Mailer error.

FlusherDock1:
dispatchSentEvent() waits for SentMessage instance, not Message.


daftspunk:
Good catch, thanks

PR #590, opened 2022-07-25
Fix SQL error when using useRelationCount + Sortable trait

vevers:
As discussed by e-mail with @samgeorges :

This fixes an SQL error when using a listview with column with a `relationCount` for a child model, which has the Sortable trait. This issue applies only to v2.x and doesn't seem to occur in v3.

**How to reproduce issue:**
- OctoberCMS 2.2.32
- Use a list with a column displaying the count of a relation using relationCount: true
- Make sure the relation model has been made sortable using the Sortable trait.

Example:
- Let’s say we have a Slideshow model and a Slide model. 
- Each Slideshow can have multiple Slides using a hasMany relation.
- The Slideshow listview displays a column with the Slides count for each slideshow
- The Slide model is Sortable. 

**This fails because:**
- relationCount: true adds a ->withCount() to the querybuilder (Backend\Widgets\Lists->prepareQuery()) method.
- This causes Laravel to generate a subquery counting the number of results
- But, the Sortable trait modifies this subquery and adds an ORDER BY clause.
- This causes MySQL to fail when strict mode is enabled. 

**Solution**
- This pull-request updates the SortableScope trait with a check to verify the scope isn't used in conjunction with a count(*) select statement.
-  Added corresponding test to SortableTest

### DIFF
--- a/src/Database/Scopes/SortableScope.php
+++ b/src/Database/Scopes/SortableScope.php
@@ -17,7 +17,30 @@ class SortableScope implements ScopeInterface
      */
     public function apply(BuilderBase $builder, ModelBase $model)
     {
+        if ($this->calledByAggregator($builder)) {
+            return;
+        }
+
         $builder->getQuery()->orderBy($model->getQualifiedSortOrderColumn());
+    }
+
+    /**
+     * check if the current query is called by an aggregator, i.e. when used with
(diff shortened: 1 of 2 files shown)


samgeorges:
Looks great! Thanks for the detailed fix and tests.

PR #591, opened 2022-08-30
Apply white background to jpg image when its convert from png

Samuell1:
When image is converted from PNG it applies black background color, but mostly JPG images are needed with white color.


daftspunk:
Rolled back: `$width, $height` were not defined

Samuell1:
@daftspunk Oh i forgot to copy it, i added it, should i make new PR?

PR #592, opened 2022-09-05
Apply white background to jpg image when its converted from png

Samuell1:
When image is converted from PNG it applies black background color, but mostly JPG images are needed with white color.


Samuell1:
@daftspunk Any news on this, it should be working correctly now.

PR #595, opened 2022-10-11
Add alias for 'flash' so it is resolvable using the class name (3.x)

adrenth:
This allows developers to resolve the FlashBag using Dependency Injection:

```php
public function __construct(
    private FlashBag $flash
) {
}

public function doSomething(): void
{
    $this->flash->info('I did something...');
}
```

Refs: https://github.com/octobercms/october-private/issues/264


daftspunk:
Thanks!

PR #596, opened 2022-10-16
Fix custom system paths

acasar:
By default ``cache_path`` and ``storage_path`` are the same, so it would make sense to support that for custom paths too. 

Right now when we set the same custom value for these two paths in the ``system.php`` config file, subfolders are only generated for the ``cache_path``, while the ``storage_path`` subfolders are skipped since the root folder already exists: https://github.com/octobercms/library/blob/3.x/src/Foundation/Bootstrap/RegisterOctober.php#L86

This PR fixes that by merging $cachePaths and $storagePaths when both paths are the same.


daftspunk:
Makes sense, thanks!

PR #604, opened 2023-01-19
Import classes on models without backslash on start

Samuell1:
This makes adding new traits more compatible and style friendly for editors when doing refactors.


daftspunk:
Our coding style suggests that if a class name is used only once, it does not need to be imported, especially for traits which are in close proximity to the imports. This PR is a good example of why (less code).

PR #606, opened 2023-03-05
Validation trait: Allow validation of related models from parent

alxy:
This allows us to do validations on related models as well. Consider the case where `Service` belongsTo `Client`, you could create a validation rule like `'client.size' => 'required'` in the `Service` model.


daftspunk:
Thanks @alxy

PR #611, opened 2023-04-27
Update Sluggable.php

ZhiweiWu0425:
Transliterate a string to its closest ASCII representation. Prevent situations where special languages cannot generate slug properly


daftspunk:
Merged in 963ee0eb6b3f4b4e74252ec9c059a46df5e673ba (3.4)

PR #618, opened 2023-07-22
Remove env usage

Samuell1:


daftspunk:
This is fixed in faa45a733996b428837c72934f0f14b8e4826db9 without breaking the composer URL

PR #626, opened 2024-04-11
Update ExtendableTrait.php get_parent_class function to remove deprec…

theshado40:
Php 8.3 introduced an deprecation error on the get_parent_class function.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -435,7 +435,7 @@ trait ExtendableTrait
             }
         }
 
-        $parent = get_parent_class();
+        $parent = get_parent_class($this);
         if ($parent !== false && method_exists($parent, '__get')) {
             return parent::__get($name);
         }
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

daftspunk:
This change looked safe as houses, but since `$this` referred to `static::class` when PHP internally used `self::class`. This produced an error since the static class had a parent, but the class implementing the trait had none.

PHP should add this to their upgrade guide or something:

- `get_parent_class()` is equivalent to `get_parent_class(self::class)`

daftspunk:
Retry fix here: https://github.com/octobercms/library/commit/a9566ca160354be28a871cc9bfb230827b1391af

Tests fine over here, need another PHP 8.3 user to confirm .

Eoler:
> Retry fix here: [a9566ca](https://github.com/octobercms/library/commit/a9566ca160354be28a871cc9bfb230827b1391af)
> Tests fine over here, need another PHP 8.3 user to confirm .

Yes, that one works for me in PHPv8.2+

PR #627, opened 2024-04-19
Correctly pass connection to Grammar class

Samuell1:
This fixes issue with running ddRawSql methods on query

` RuntimeException  The database driver's grammar implementation does not support escaping values.`

Not sure if this needs to be done for every connection driver just tested MYSQL and it works.


daftspunk:
Thanks!

PR #628, opened 2024-05-06
Add filesystem.localToPublic event

acasar:
Would you consider adding this event to the core? We have a special case where we are using symlinks to include core OctoberCMS files into the project to isolate each project under its own cPanel account while still beeing able to update all projects with a single deploy. Currently such a special situation is sometimes not handeled correctly by the ``localToPublic`` helper (null is returned).

image

By adding this event we could allow developers to implement custom logic for converting local to public paths and solve special cases such as this.


daftspunk:
Looks great, thanks!

PR #630, opened 2025-02-19
Restore Illuminate Redirector alias for backward compatibility

jacobdekeizer:
As described in https://github.com/octobercms/october-private/issues/682, calls to the Illuminate Redirector via dependency injection fail to set the session because the 'redirect' alias no longer exists. This change restores that functionality.


daftspunk:
Thanks!

PR #631, opened 2025-03-21
Fix saving the model

mrmaarek:
After updating the save signature, the calls have not been changed:

https://github.com/octobercms/library/commit/acdfa403a4b082f391a93b728ff1ba737a5030d8

Now i can't save a model anymore


daftspunk:
I'm not sure I follow, it allows null (casts to empty array)

```php
> (array) null
= []
```

Why can't you save the model anymore?

PR #632, opened 2025-03-27
Why is the VerifyCsrfToken middleware commented out in this code?

i3bepb:
Why is the VerifyCsrfToken middleware commented out in this code?
Is this a typo?


daftspunk:
Hi @i3bepb 

It's not a typo, but it breaks AJAX, which has native browser protection for CSRF, and it isn't required. 

I should mention: October has its own checks for CSRF, we could potentially switch over to this middleware in v4 as a breaking change now that we have a `{% meta %}` tag.

PR #637, opened 2025-07-01
rtrim() on AssetFactory.php

pvullioud:
rtrim(): Passing null to parameter #1 ($string) of type string is deprecated in PHP 8.3


daftspunk:
Thanks!

PR #639, opened 2025-07-22
Use October\...\HandleExceptions in Console\Kernel

vdlp-mw:
We encountered an issue with de new deprecation logging feature. Through configuration this feature is switched off by default but we noticed that deprecations were still logged as warnings when running console command.

Turns out that the new `HandleExceptions` class was not registered in the Console\Kernel. 

This change is already implemented in the Http\Kernel which was added in https://github.com/octobercms/library/commit/ed6aa00a590f232fac38fd5bb8406f1ca9149afe


daftspunk:
Thanks!

PR #642, opened 2025-08-20
Add morphToMany to be able to use with relation sort

Samuell1:
This pull request makes a small update to the `defineSortableRelations` method in `src/Database/Traits/SortableRelation.php` to improve support for relation types. The change adds `morphToMany` to the list of relation types that interact with pivot tables.

* Added `morphToMany` to the `$interactsWithPivot` array to ensure sortable relations work correctly for both `belongsToMany` and `morphToMany` relationships.


daftspunk:
Thanks!

PR #643, opened 2025-09-05
Update ComposerManager.php

Dheia:
Problem Description and Solution for "Class core.composer does not exist" in October CMS v4

Problem

In October CMS version v4 (built on Laravel 12), an issue occurred when loading certain plugins that contained their own vendor folder and composer.json file. The system was throwing the following error:

```
ReflectionException: Class "core.composer" does not exist
```

This error occurred because the system was trying to access the core.composer object from the Laravel Container at a time when it wasn't yet available, particularly during the initial system initialization process.

Root Cause

The error was occurring in the instance() method within the October\Rain\Composer\ComposerManager class, which contained the following code:

```php
public static function instance(): static
{
    return App::make('core.composer');
}
```

In previous versions of October CMS, the core.composer object was registered in the container synchronously with class loading. However, in v4 with Laravel 12, the service provider loading mechanism changed, causing timing conflicts.

Solution

The instance() method was modified to first check if the object exists in the container, and if not, create a new instance instead:

```php
public static function instance(): static
{
    try {
        return App::make('core.composer');
    } catch (\Illuminate\Contracts\Container\BindingResolutionException $e) {
        // If not available in the container, create a new instance
        return new static();
    }
}
```

Required Changes

1. Modify the file: plugins/october/rain/src/Composer/ComposerManager.php
2. Update the instance() method as shown above

Additional Notes

· This solution maintains compatibility with previous versions of October CMS · It doesn't affect the current functionality of the system · It solves the problem without requiring major changes to the system architecture · It works as a preventive solution even when the core.composer object isn't registered in the container

This solution has been tested on October CMS v4 and has proven effective in resolving the issue without causing any side effects.

### DIFF
--- a/src/Composer/ComposerManager.php
+++ b/src/Composer/ComposerManager.php
@@ -41,7 +41,11 @@ class ComposerManager
      */
     public static function instance(): static
     {
-        return App::make('core.composer');
+        try {
+            return App::make('core.composer');
+        } catch (\Illuminate\Contracts\Container\BindingResolutionException $e) {
+            return new static();
+        }
     }
 
     /**


daftspunk:
Thanks!

PR #644, opened 2025-09-08
fix(ErrorHandler): replace array_first with foreach in isNotFoundExce…

m49n:
…ption

On newer stacks (Laravel 12.23+/PHP 8.5 polyfill), global array_first no longer accepts a callback, making isNotFoundException() always true and forcing 404 for all exceptions. Replace with a plain foreach to restore correct detection.


daftspunk:
Perfect! Thanks

PR #648, opened 2026-01-04
Update PHP requirement to 8.1 and remove strftime() deprecation

Samuell1:
- Bumps minimum PHP version from 8.0.2 to 8.1
- Replaces deprecated strftime() in FormBuilder::selectMonth() with DateTime::format()
- Changes default format parameter from strftime '%B' to DateTime 'F' (both output full month name)

### DIFF
--- a/src/Html/FormBuilder.php
+++ b/src/Html/FormBuilder.php
@@ -558,15 +558,16 @@ class FormBuilder
      * @param  string  $name
      * @param  string  $selected
      * @param  array   $options
-     * @param  string  $format
+     * @param  string  $format DateTime format string (default: 'F' for full month name)
      * @return string
      */
-    public function selectMonth($name, $selected = null, $options = [], $format = '%B')
+    public function selectMonth($name, $selected = null, $options = [], $format = 'F')
     {
         $months = [];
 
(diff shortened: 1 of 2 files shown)


daftspunk:
Thanks!

src/Html/FormBuilder.php
```diff
@@ -558,15 +558,16 @@ public function selectYear()
...
      * @return string
      */
-    public function selectMonth($name, $selected = null, $options = [], $format = '%B')
+    public function selectMonth($name, $selected = null, $options = [], $format = 'F')
     {
         $months = [];
 
         foreach (range(1, 12) as $month) {
-            $months[$month] = strftime($format, mktime(0, 0, 0, $month, 1));
+            $date = new \DateTime("2024-{$month}-01");
```
samuelpatro, line 569: @daftspunk now looking at this there is year 2024 :D can this break something?

daftspunk: It's fine. It needs a year, even if unused...

```
DEPRECATED  DateTime::setDate(): Passing null to parameter #1 ($year) of type int is deprecated
```

PR #650, opened 2026-01-13
Add MariaDB driver and error detector singletons from Laravel 12

Samuell1:
- Add MariaDbConnection class as separate driver (not MySQL variant)
- Update ConnectionFactory to support 'mariadb' driver
- Register ConcurrencyErrorDetector singleton for deadlock detection
- Register LostConnectionDetector singleton for reconnection logic
- Remove unnecessary Model::flushEventListeners() call (reduces overhead)

The error detectors are used internally by Laravel for:
- ConcurrencyErrorDetector: Detecting deadlocks and serialization failures to enable automatic transaction retries
- LostConnectionDetector: Detecting lost database connections to trigger reconnection


daftspunk:
Thanks @Samuell1 !

PR #652, opened 2026-01-15
Remove attribute access events for performance

Samuell1:
Remove model.beforeGetAttribute, model.getAttribute, model.beforeSetAttribute, and model.setAttribute events from Database and Halcyon models.

These events fired on every attribute access/mutation, causing significant overhead (e.g., 50 users × 15 attributes × 2 events = 1,500 event fires for a simple list). Laravel does not fire events for attribute access.

Breaking change: Code relying on these events will need refactoring.

### DIFF
--- a/src/Database/Concerns/HasAttributes.php
+++ b/src/Database/Concerns/HasAttributes.php
@@ -19,13 +19,6 @@ trait HasAttributes
     {
         $attributes = $this->getArrayableAttributes();
 
-        // Before Event
-        foreach ($attributes as $key => $value) {
-            if (($eventValue = $this->fireEvent('model.beforeGetAttribute', [$key], true)) !== null) {
-                $attributes[$key] = $eventValue;
-            }
-        }
-
         // Dates
         $attributes = $this->addDateAttributesToArray($attributes);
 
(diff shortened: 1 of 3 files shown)


daftspunk:
This breaks backwards compatibility so we'll need to introduce a `LeanModel` to pilot these changes. 

Later, come up with a strategy that doesn't impact the thousands of legacy sites running October CMS.

Samuell1:
If we can do like something that doesnt require changing classes in whole codebase i am for it.

Does anyone uses this event? because i think its pretty unique to hook something here now, when we have traits per model.

daftspunk:
The event is used by Translate plugin mostly. It was used by `SettingsModel` behavior, but since we moved to `SettingModel` base class we can use method overrides instead. 

The event can be supplemented by behaviors defining `getFoobarAttribute()` and October model will look for it. However, it doesn't work well for dynamic attribute getters/setters, so the event is there for that. Perhaps we can include some dedicated method for behaviors instead to avoid events (Keep in mind, unbound events are just a couple of method calls as well).

Here is the biggest reason to keep it around:
https://github.com/rainlab/translate-plugin/blob/master/classes/TranslatableBehavior.php#L63-L83

Samuell1:
Okay, that translatable behavior is my next stop to check it and correctly migrate to model trait, its pretty heavy from what i seen in my tests and performance logs.

Samuell1:
@daftspunk this maybe can be moved to core because of changes for other stuff you did, not to leanmodel?

daftspunk:
It breaks RainLab.Translate though

Samuell1:
then maybe backport this specific change to translate and core should be without this events?

daftspunk:
They all extend this model so its not easy to do without a breaking change

PR #653, opened 2026-01-15
Add empty key check to getAttribute() to match Laravel

Samuell1:
Sync getAttribute() with Laravel's implementation by adding an early return for null/empty keys. This prevents unnecessary processing and matches Laravel 12's behavior.


daftspunk:
Merged to feature proposal branch. Thanks!

PR #655, opened 2026-01-16
Removes query-level caching from QueryBuilder

samuelpatro:


daftspunk:
No real benefit in removing this. It is used in Twig,

```twig
{% set records = model.where('foo', 'bar').remember(20).get() %}
```

Samuell1:
From my point it too much depends on query what needs to cache because just this line 
`md5($name.$this->toSql().serialize($this->getBindings()))` makes its heavier then it looks and then key cant be controlled externally, its just makes more limitations then possibilities then i needed to use Cache::remember closure to make it work better.

Maybe move it to trait, then it can be used on model where people need it. Thats why i added is as draft, to maybe add more to it or rethink this :)

daftspunk:
Twig doesn't have closures so this is why we kept it.

Laravel removed it for the same reasons you mention, it can't serve every possible combination of query. It serves simple use cases very well, though.

I'm not sure it is possible to make a trait for the query builder.

Samuell1:
I will try to look into it for best possible solution to dont limit performance improvements and code quality.

daftspunk:
Let's keep this for now, it is fine for smaller/simple use cases. For advancaed stuff, it is recommended to use the standard Cache approach given by Laravel.

This also isn't site / context aware:

- [ ] add the `site_id` context to the auto generated key

PR #656, opened 2026-01-17
Remove PHP 8+ polyfills and use native string functions

Samuell1:
- Remove starts_with() and ends_with() helper functions
- Replace with native str_starts_with() and str_ends_with()
- Use Str::startsWith() for array-needle cases (HtmlBuilder, SyntaxModelTrait)


daftspunk:
Thanks!

PR #657, opened 2026-01-18
Refactor FormBuilder to use Request facade instead of global helpers

Samuell1:
Replace deprecated input() and post() global function calls with Request facade methods in FormBuilder for better maintainability.


daftspunk:
Thanks!

PR #659, opened 2026-01-18
Replace legacy array helper functions with Arr class methods

Samuell1:
Replaces deprecated global array helper functions with their Illuminate\Support\Arr equivalents across 21 files:
- array_get() → Arr::get()
- array_first() → Arr::first()
- array_except() → Arr::except()

They still exists in codebase just moving that we can deprecate them later.

### DIFF
--- a/src/Database/Attach/File.php
+++ b/src/Database/Attach/File.php
@@ -6,6 +6,7 @@ use Cache;
 use Storage;
 use Response;
 use File as FileHelper;
+use Illuminate\Support\Arr;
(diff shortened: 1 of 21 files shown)


daftspunk:
Laravel seems to be a bit confused about using these helpers or using Facades... is this the "one true way"?

Samuell1:
@daftspunk i think this is issue that php started doing in core this methods and they dont match laravel param order or they dont work correctly, thats why they started doing this

daftspunk:
Thanks!

PR #665, opened 2026-06-16
Prevent filesystem fallback for soft-deleted DB templates

scottbedard:
https://github.com/octobercms/october/issues/6056

## Problem

When database templates are enabled (`CMS_DB_TEMPLATES`), soft-deleting a template sets `deleted_at` on the DB row but `AutoDatasource` still falls through to the filesystem layer. The template incorrectly appears in the CMS and on the frontend.

## Solution

**`DbDatasource`**
- `isTemplateTrashed` - detect soft-deleted records at a path
- `selectTrashedFileNames` - list tombstoned file names in a directory
- `$trashedPathCache` - load tombstoned paths once per source (same pattern as `$mtimeCache`)
- Shared helpers for `select` / `selectTrashedFileNames`: `buildDirectoryQuery`, `applyExtensionsFilter`, `pathToFileName`, `matchesFileMatch`

**`AutoDatasource`**
- Guard `hasTemplate`, `selectOne`, `select`, and `lastModified` so a DB tombstone blocks all lower layers
- On-disk files are left untouched

## Tests

New tests under `tests/Halcyon/Datasource` with `HalcyonDbTestHarness` (in-memory SQLite + facade isolation). The harness saves and restores the Laravel facade app after each test to avoid breaking other suites.

```bash
./vendor/bin/phpunit tests/Halcyon/Datasource/
```

## QA

**Setup**

1. Create a codespace from [`octobercms/october`](https://github.com/octobercms/october)
2. Add `CMS_DB_TEMPLATES=true` to `.env`
3. Modify `composer.json`
```diff
+ "repositories": [
+     {
+         "type": "git",
+         "url": "https://github.com/scottbedard/library.git"
+     }
+ ],
  "require": {
-     "october/rain": "^4.3",
+     "october/rain": "4.x-dev",
  },
```
4. Run `composer install` and `php artisan config:clear`

**Confirmation 1: Deleting un-modified file**

1. Delete a page _without modifying it_
3. Refresh backend, confirm page is gone
3. Navigate to page, confirm 404
4. Refresh codespace, confirm file exists on disk

**Confirmation 2: Deleting a modified file**

1. Modify a page, then click save
2. Delete the page
3. Refresh backend, confirm page is gone
4. Navigate to page, confirm 404
5. Refresh codespace, confirm file exists on disk

**Confirmation 3: New file behavior does not change**

1. Create a page
2. Navigate to page, confirm 200
3. Refresh codespace, confirm no file
4. Delete file
5. Navigate to page, confirm 404

### DIFF
--- a/src/Halcyon/Datasource/AutoDatasource.php
+++ b/src/Halcyon/Datasource/AutoDatasource.php
@@ -39,6 +39,10 @@ class AutoDatasource extends Datasource implements DatasourceInterface
      */
     public function hasTemplate(string $dirName, string $fileName, string $extension): bool
     {
+        if ($this->isTemplateTrashed($dirName, $fileName, $extension)) {
+            return false;
+        }
+
         foreach ($this->datasources as $datasource) {
             if ($datasource->hasTemplate($dirName, $fileName, $extension)) {
                 return true;
@@ -53,6 +57,10 @@ class AutoDatasource extends Datasource implements DatasourceInterface
      */
     public function selectOne(string $dirName, string $fileName, string $extension)
     {
+        if ($this->isTemplateTrashed($dirName, $fileName, $extension)) {
+            return null;
+        }
+
         foreach ($this->datasources as $source) {
             if (!$source->hasTemplate($dirName, $fileName, $extension)) {
                 continue;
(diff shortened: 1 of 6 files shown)


daftspunk:
Thanks!

PR #658, opened 2026-01-18
Improve routing caching

samuelpatro:

PR #669, opened 2026-08-21
Compile routes for matching and cache the compiled state

samuelpatro:
Reworked follow-up to #658. Routes compile into optimized lookup structures on first match: static routes in a hash map, dynamic routes in combined `(*MARK)` regexes bucketed by first segment, wildcard routes sequential. The compiled state travels through `toArray()`/`fromArray()`, so the CMS route cache carries it with no changes on the CMS side. Rules restored from cache hydrate lazily, only the matched rule becomes an object.

Matching behavior is identical to the sequential matcher, verified by a parity test and by 500 sampled URLs from a production site's traffic shape (52.8k visited URLs over ~25 pages, query strings included). Custom segment expressions are validated post-match with the original code path (case sensitivity, substring search and error tolerance preserved), condition rejections continue to later rules, wildcards keep their sorted priority.

Benchmarks at 120 routes (`RouterBench`, Apple Silicon, PHP 8.4):

| | before | after |
|---|---|---|
| static URL match | 10.7 µs | 0.4 µs |
| dynamic URL match | 13-21 µs | 1.5-1.7 µs |
| no match | 14.9 µs | 1.1 µs |
| cached boot + match | 21.1 µs | 4.4 µs |

Matching stays ~1 µs up to 1000 routes (bucketed regexes), where sequential grows to ~35 µs.

`fromArray()` accepts both the new format and the legacy rules list, existing caches survive. `toArray()` now returns `['rules' => ..., 'compiled' => ...]`, the serialized payload grows ~45% (25 routes: 10.6 KB to 15.3 KB). New API: `hasRoute()`, `getRoute()`, `isCompiled()`, `compileRules()`, `getCompiledRoutes()`/`setCompiledRoutes()`, `saveCompiledRoutes()`/`loadCompiledRoutes()`.

With this in place the CMS side no longer needs the per-URL `cms-url-list` cache (unbounded blob, deserialized fully per request), removed separately in octobercms/october-private#791.

### DIFF
--- a/src/Router/README.md
+++ b/src/Router/README.md
@@ -39,3 +39,20 @@ You can also reverse match a route by it's identifier:
 // Returns: /post/2
 $url = $router->url('myRouteId', ['id' => 2]);
 ```
+
+## Compiled route caching
+
+Routes compile automatically on the first match into optimized lookup structures (a hash map for static routes and a combined regex for dynamic routes). The compiled state can be stored and restored to skip compilation entirely, similar to Laravel's route caching:
+
+```php
+// Store the routes with their compiled state
+$cached = $router->toArray();
+
+// Restore later, no recompilation needed
+$router = new Router;
+$router->fromArray($cached);
+
+// Alternatively, save to and load from a PHP file
+$router->saveCompiledRoutes('/path/to/routes.php');
+$router = Router::loadCompiledRoutes('/path/to/routes.php');
+```
--- a/src/Router/RouteCompiler.php
+++ b/src/Router/RouteCompiler.php
@@ -95,7 +95,11 @@ class RouteCompiler
 
         $dynamicRegexes = [];
         foreach ($patterns as $bucket => $branches) {
-            $dynamicRegexes[$bucket] = '#^(?|' . implode('|', $branches) . ')$#i';
+            // The u modifier keeps case-insensitive matching correct for
+            // multibyte (translated) static segments. URLs with invalid
+            // UTF-8 make preg_match fail, the router falls back to
+            // sequential matching for those.
+            $dynamicRegexes[$bucket] = '#^(?|' . implode('|', $branches) . ')$#iu';
         }
 
         return [
--- a/src/Router/RouteCompiler.php
+++ b/src/Router/RouteCompiler.php
@@ -1,0 +1,343 @@
+<?php namespace October\Rain\Router;
+
+/**
+ * RouteCompiler transforms route rules into optimized lookup structures
+ * for faster route matching using static hash maps and combined regexes.
+ *
+ * Uses the same approach as Laravel:
+ * - Static routes: Hash map for O(1) lookup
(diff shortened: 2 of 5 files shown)


samuelpatro:
Scaling benchmark, µs per `match()`, worst-position routes (Apple Silicon, PHP 8.4). Sequential is the current rule scan, compiled is this PR.

Compiled matching is flat no matter how many routes exist: a static hit is one hash lookup, a dynamic hit is one regex against the URL's first-segment bucket.

| lookup | routes | sequential | compiled | speedup | saved per request |
|---|---:|---:|---:|---:|---:|
| static hit | 120 | 5.3 µs | 0.45 µs | 12× | 4.9 µs |
| | 1,000 | 43.9 µs | 0.43 µs | 101× | 43.5 µs |
| | 5,000 | 217.6 µs | 0.45 µs | 483× | 217.2 µs |
| | 10,000 | 460.4 µs | 0.44 µs | 1,051× | 0.46 ms |
| dynamic hit | 120 | 13.2 µs | 0.94 µs | 14× | 12.3 µs |
| | 1,000 | 104.4 µs | 0.94 µs | 111× | 103.5 µs |
| | 5,000 | 527.6 µs | 1.00 µs | 527× | 0.53 ms |
| | 10,000 | 1,118.7 µs | 0.96 µs | 1,162× | 1.12 ms |
| no match | 120 | 4.5 µs | 0.47 µs | 10× | 4.0 µs |
| | 1,000 | 34.8 µs | 0.46 µs | 75× | 34.3 µs |
| | 5,000 | 181.2 µs | 0.49 µs | 370× | 180.7 µs |
| | 10,000 | 352.3 µs | 0.46 µs | 759× | 0.35 ms |

The table above is matcher vs matcher. Full CMS routing today has a second layer on top, the `cms-url-list` cache, which decodes its whole blob on every request and rewrites it on every new URL. End-to-end per request, measured on a decoded production blob (52.8k cached URLs / 47 MB, one locale of a live shop) together with octobercms/october-private#791:

| request | old routing (blob + matcher) | new (this PR + #791) | saved |
|---|---:|---:|---:|
| URL already cached | ~23 ms | ~4 µs | ~23 ms |
| new URL (crawler, tracking params) | ~34 ms | ~4 µs | ~34 ms |

The new 4 µs includes restoring the compiled route map from cache and matching. Removing the blob is only safe because compiled matching is cheap enough to run on every request.

Compilation runs once per cache rebuild, not per request:

| routes | compile | cache size |
|---:|---:|---:|
| 120 | 0.5 ms | 87 KB |
| 1,000 | 1.2 ms | 740 KB |
| 5,000 | 7.2 ms | 3.7 MB |
| 10,000 | 18.9 ms | 7.4 MB |

Route mix: 30% static, 60% dynamic, 10% dynamic with a regex constraint.

daftspunk:
Thanks!

PR #671, opened 2026-08-30
Read a template once when resolving through AutoDatasource

samuelpatro:
**Problem.** `FileDatasource::hasTemplate()` was implemented as `(bool) $this->selectOne()`, so checking whether a template exists read the whole file. `AutoDatasource::selectOne()` and `lastModified()` both called `hasTemplate()` on each datasource before doing their own work. Net effect per template: two full file reads on a cold load, one full file read on every warm load (via `Builder::isCacheBusted()` → `lastModified()`).

**Changes**
- `src/Halcyon/Datasource/FileDatasource.php`: `hasTemplate()` now returns `Filesystem::isFile()` on the resolved path.
- `src/Halcyon/Datasource/AutoDatasource.php`: `selectOne()` returns the first non-null result from the chain; `lastModified()` returns the first non-null mtime. The `hasTemplate()` pre-check in both is gone. Trashed-template check before the chain is unchanged.

**Tests** (`tests/Halcyon/Datasource/AutoDatasourceTest.php`), using a `Filesystem` subclass that records `get()` calls:
- `testSelectOneReadsTemplateContentOnce`: 1 read (develop: 2)
- `testSelectOneSkipsDatasourcesWithoutTheTemplate`: db datasource first, file second, still 1 read
- `testLastModifiedDoesNotReadTemplateContent`: 0 reads (develop: 1)
- `testHasTemplateDoesNotReadTemplateContent`: 0 reads

### DIFF
--- a/src/Halcyon/Datasource/AutoDatasource.php
+++ b/src/Halcyon/Datasource/AutoDatasource.php
@@ -61,12 +61,13 @@ class AutoDatasource extends Datasource implements DatasourceInterface
             return null;
         }
 
+        // Each datasource answers with the template or null, so the
+        // existence check and the read are a single operation
         foreach ($this->datasources as $source) {
-            if (!$source->hasTemplate($dirName, $fileName, $extension)) {
-                continue;
+            $result = $source->selectOne($dirName, $fileName, $extension);
+            if ($result !== null) {
+                return $result;
             }
-
-            return $source->selectOne($dirName, $fileName, $extension);
         }
 
         return null;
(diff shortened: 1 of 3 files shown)


samuelpatro:
Recheck of head `b844e1be` found a datasource selection regression.

**[P2] Reject directory paths before accepting a template mtime.** `AutoDatasource::lastModified()` now accepts the first non-null result, but `FileDatasource::lastModified()` calls filesystem `lastModified()` without checking `isFile()`. A directory has a valid timestamp too.

Reproduced using two real filesystem datasources: source one contains a directory named `pages/foo.htm`; source two contains the actual `pages/foo.htm` file. `selectOne()` correctly skips source one and reads source two. On the merge base, `lastModified()` also returns source two's timestamp (1500000000). On this PR it returns source one's directory timestamp (1000000000). Cache validation can therefore watch the wrong source and miss changes to the actual template.

Add a cheap `isFile()` guard in FileDatasource::lastModified() (return null for non-files), preserving zero content reads, plus a regression covering the directory/file collision and subsequent template modification. The existing suite passes (234 tests, 1460 assertions, one pre-existing risky test); this edge is not covered.

daftspunk:
Good refactor! Thanks

PR #673, opened 2026-08-30
Remove Laravel 10 compatibility shims

samuelpatro:
`composer.json` requires `laravel/framework ^12`. These three guards were written for 9/10 and are always true on 12; verified against `vendor/laravel/framework`.

- `src/Events/EventServiceProvider.php`: removed `method_exists($dispatcher, 'setTransactionManagerResolver')` (added in Laravel 10.30, present in `Illuminate\Events\Dispatcher` on 12). The singleton now uses the chained call that was sitting above it as a comment.
- `src/Translation/FileLoader.php`: removed the redeclared `$path` and `$paths` properties and the `(array) $this->path ?: $this->paths` expression. `Illuminate\Translation\FileLoader` on 12 has only `$paths`, so `$this->path` was always null and the right side always won. `loadNamespaceOverrides()` now iterates `$this->paths` directly.
- `src/Database/Dongle.php`: removed `interface_exists(\Illuminate\Contracts\Database\Query\Expression::class)` from `rawValue()`. The contract ships with 12; the `(string) $this->db->raw($sql)` fallback was unreachable.

No behavior change. Library suite and CMS suite pass unchanged.

### DIFF
--- a/src/Database/Dongle.php
+++ b/src/Database/Dongle.php
@@ -39,16 +39,10 @@ class Dongle
 
     /**
      * rawValue converts a raw expression to a string
-     *
-     * @todo Can be refactored if Laravel >= 10
      */
     public function rawValue($sql): string
     {
-        if (interface_exists(\Illuminate\Contracts\Database\Query\Expression::class)) {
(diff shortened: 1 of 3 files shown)


daftspunk:
Thanks!

PR #674, opened 2026-08-30
Drop the doctrine/dbal requirement

samuelpatro:
Removes `"doctrine/dbal": "^2.13.3|^3.1.4"` from `composer.json`.

It was added in 461271a4 for the Dongle. Laravel 11 dropped its own DBAL dependency and does column changes natively. Today nothing references it: `grep -ri "doctrine\|dbal"` over `src/`, `init/`, and every CMS module, config and plugin returns nothing, and `composer why doctrine/dbal` lists only two `conflicts` entries (carbon-doctrine-types, symfony/http-foundation), no dependents. The `^2|^3` constraint also blocks DBAL 4 for projects that need it themselves.

Verified by running `composer update doctrine/dbal` (removes dbal, deprecations and event-manager from vendor) and then the full library suite (230 tests) and the CMS suite against this library (618 tests). Both pass.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
         "composer/composer": "^2.0.0",
         "composer/installers": "^1 || ^2",
         "larajax/larajax": "^2.0",
-        "doctrine/dbal": "^2.13.3|^3.1.4",
         "intervention/image": "^3.10",
         "jaybizzle/crawler-detect": "^1.3",
         "linkorb/jsmin-php": "~1.0",


daftspunk:
It's used by Builder I think

samuelpatro:
Recheck of head `70b57959` against current Builder source found a compatibility blocker.

**[P1] Preserve Builder's DBAL dependency before removing Rain's requirement.** [Builder Plugin.php](https://github.com/rainlab/builder-plugin/blob/72e3589e0c46d04b91d22e007e2449913d7ff292/Plugin.php#L195-L198) calls `Doctrine\DBAL\Types\Type::hasType()` during plugin registration, and its database tools also use DBAL. [Builder composer.json](https://github.com/rainlab/builder-plugin/blob/72e3589e0c46d04b91d22e007e2449913d7ff292/composer.json) does not declare DBAL; it relies on Rain supplying it. The maintainer's concern above is confirmed.

On an installation with Builder and no other DBAL dependency, updating dependencies after this removal can remove DBAL and make registration fail with a missing Doctrine class. Library/CMS suites without Builder do not cover this. Retain the requirement until Builder declares a compatible direct dependency and the release transition is accounted for, or coordinate that migration before merging. Narrow the description's claim that no plugin references DBAL.

The branch also conflicts with current develop in composer.json; retain develop's PHP/larajax requirements when updating.

PR #677, opened 2026-09-05
Fix memoized attachment existence after generation and deletion

samuelpatro:
When remote thumbnail generation follows a previously cached missing-file result, attachment existence checks keep returning the request's memoized `false` even after the thumbnail is published. Repeated `getThumb()` calls then regenerate the same image. Deleting a file or thumbnail can likewise leave a memoized `true`.

Write successful publication and both deletion paths through `Cache::memo()` so the backing cache and request cache stay consistent. Cache keys, storage behavior, and thumbnail URLs are unchanged.

Four regression tests run the real thumbnail flow with the real resizer on a generated 4x4 image, replacing only the remote disk with an in-memory fake. The persisted-miss case now publishes once across two calls (previously twice); deletion tests verify the next read observes removal. Full suite on PHP 8.4: 234 tests, 1468 assertions.

Rechecked on 2026-09-18: the real-resizer fixture publishes a thumbnail twice on develop and once on this PR across two calls following a cached miss. The full suite still passes 234 tests / 1468 assertions. The fixture restores the resolved facade registry as well as its application.


daftspunk:
Thanks!

PR #678, opened 2026-09-05
Limit combined attachment eager loading to requested fields

samuelpatro:
Eager-loading one attachment field currently fetches every attachment for the selected owners, then filters the hydrated models by field. An entry with one cover and 100 gallery files therefore hydrates 101 file records for `with('cover')`.

Restrict the combined attachment query to the requested fields that share the same related model, while keeping one query for multiple fields. The combined result cache is reset at the start of each eager-load pass so a reused builder loads the correct owners and fields. Scoped/conditional relations, `combineEager => false`, and first-constraint behavior are preserved.

10 regression tests cover cover-only hydration (101 → 1 file, one attachment query), multiple owners and fields, distinct related classes, opt-outs, nested loads, custom constraints, relation hooks, and reused builders. Full suite on PHP 8.4: 240 tests, 1493 assertions.

Rechecked on 2026-09-18 with PHP 8.4 / SQLite. Seven samples of ten warm cover-only loads (fixture setup excluded): with 100 unrelated gallery files, hydrated files drop from 101 to 1 and median time from 0.618 ms to 0.100 ms; with 1000 gallery files, from 1001 to 1 and 5.740 ms to 0.130 ms. Both versions issue two total queries (owner plus attachments). These are local illustrative timings; work-count tests are the regression gate. Full suite remains 240 tests / 1493 assertions.

### DIFF
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -17,6 +17,19 @@ class Builder extends BuilderModel
     use \October\Rain\Database\Concerns\HasEagerLoadAttachRelation;
 
     /**
+     * eagerLoadRelations eagerly load the relationships for the models, combining
+     * attachments freshly for each set of models.
+     * @param  array  $models
+     * @return array
+     */
+    public function eagerLoadRelations(array $models)
+    {
+        $this->eagerLoadAttachResultCache = [];
+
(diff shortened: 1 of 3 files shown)


daftspunk:
Wow! 🤯  Thanks

PR #679, opened 2026-09-05
Batch translations during model hydration

samuelpatro:
Fetching translated models in a nondefault locale currently performs one translation query per row during `afterFetch`, before normal eager loading can help. Fetching 25 records takes 26 queries, or 27 with `with('translations')`.

The builder creates its hydration prototype before preloading the active locale's translations in chunks of 500 IDs. The snapshot belongs only to that prototype; there is no shared static cache. Matching rows seed each model's existing locale cache before fetched callbacks, with the snapshot suspended during that model's lifecycle and restored in `finally`. Independent reads inside prototype, instance-creation, and fetched callbacks therefore use fresh data. Translation-storage connection, table, morph type, locale, and key must match. Cursor and other non-batch hydration use a fast path. The normal model override, collection, event, and lazy-loading behavior is retained.

The lookup is built directly per chunk without intermediate collection grouping. The scoped hydration lifecycle is retained because regression tests demonstrate stale reads when it is removed.

Validation: 19 regression tests; full PHP 8.4 suite passes 249 tests / 1519 assertions, with one pre-existing risky test lacking assertions. Coverage includes callbacks, nested reads, exceptions, locale switches, casts, complete eager relations, connection changes, missing keys, query chunk bounds, and lazy-loading flags. The prototype callback regression fails on the previous PR head and passes on this version.

Local PHP 8.4 / SQLite benchmark (seven samples, ten warm fetches per sample, median per fetch; fixture setup excluded):

| Records | Develop queries | PR queries | Develop time | PR time |
|---|---:|---:|---:|---:|
| 25 | 26 | 2 | 0.798 ms | 0.398 ms |
| 100 | 101 | 2 | 3.068 ms | 1.434 ms |
| 1001 | 1002 | 4 | 32.227 ms | 14.489 ms |

These are local illustrative timings, not production guarantees. Default-locale hydration still uses one query; cursor translation remains one lookup per record. Count-based assertions are the regression gates.

Alternating control runs (five rounds, median of the per-round medians) measured 100-record default-locale hydration at 0.561 ms on develop versus 0.584 ms here (~4% overhead), and cursor translation at 4.144 ms versus 4.373 ms (~6% overhead). These paths do not gain query batching; this is a measured tradeoff of the hydration hooks.

### DIFF
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -17,6 +17,32 @@ class Builder extends BuilderModel
     use \October\Rain\Database\Concerns\HasEagerLoadAttachRelation;
 
     /**
+     * hydrate creates models with batch-local translations available to fetched events.
+     */
+    public function hydrate(array $items)
+    {
+        if (!method_exists($this->model, 'withTranslatableBatch')) {
+            return parent::hydrate($items);
+        }
+
+        $instance = $this->newModelInstance();
+
+        $hydrate = function () use ($items, $instance) {
+            return $instance->newCollection(array_map(function ($item) use ($items, $instance) {
+                $model = $instance->newFromBuilder($item);
+
+                if (count($items) > 1) {
+                    $model->preventsLazyLoading = Model::preventsLazyLoading();
+                }
+
+                return $model;
+            }, $items));
+        };
+
+        return $instance->withTranslatableBatch($items, $hydrate);
+    }
+
+    /**
      * eagerLoadRelation eagerly load the relationship on a set of models, with support
      * for attach relations.
      * @param  array  $models
--- a/src/Database/Builder.php
+++ b/src/Database/Builder.php
@@ -17,29 +17,18 @@ class Builder extends BuilderModel
     use \October\Rain\Database\Concerns\HasEagerLoadAttachRelation;
 
     /**
-     * hydrate creates models with batch-local translations available to fetched events.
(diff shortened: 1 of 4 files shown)


samuelpatro:
Recheck of head `3c60da88` found an uncovered model override regression.

**[P2] Allow replacement translation loaders to bypass the default store.** `hydrateWithTranslatableBatch()` queries `getTranslateAttributeTable()` before a custom `loadTranslatableData($locale)` can run. A model whose loader supplies translations entirely from JSON/an external service, without calling parent, need not have the standard translation table.

Reproduced on PHP 8.4 with such a loader and no `translate_attributes` table: merge-base `cursor()` and `get()` both return the translated value; this PR's `cursor()` works but `get()` throws `SQLSTATE[HY000]: General error: 1 no such table: translate_attributes` from the preload query. The new override test decorates the parent loader and misses replacement storage.

Preserve this extension behavior with an override-aware fallback or overridable batch eligibility/storage, and add a regression without the default translation table. This contradicts the promised preservation of normal model overrides. The existing full suite passes (251 tests, 1529 assertions, one pre-existing risky test), so this needs additional coverage.

Separately, current develop has an add/add conflict in tests/Database/Traits/TranslatableTest.php; retain both sets of tests when updating.

daftspunk (commented):
Need to resolve conflicts

PR #681, opened 2026-09-18
Cache Halcyon DbDatasource indexes between requests

scottbedard:
`DbDatasource` stored template mtimes and tombstoned paths only in process-local statics, so every request reran both index queries. This persists those indexes in the application cache (`halcyon.db.{table}.{source}`) so a later request can refill the statics without hitting the database.

Writes still `flushCache()`, which now also forgets the cache entry. `DbDatasource::clearCache($source, $table)` covers deletes that never go through this class. Empty datasources are cached; failed queries are not.


daftspunk (approved):
Waiting on https://github.com/octobercms/october/pull/6126

daftspunk:
Thanks!

